PR #5, opened 2019-09-21
Improving build speed and fixed 2 lint issues

marandaneto:


sentry-android/build.gradle.kts
```diff
@@ -7,19 +9,45 @@ plugins {
...
                 includeCompileClasspath = true
             }
         }
+
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+    }
+
+    buildTypes {
+        getByName("release") {
+            isMinifyEnabled = false
```
bruno-garcia: is release going to be `isMinifyEnabled = false`?

marandaneto: good catch, just enabled it for release build

PR #9, opened 2019-10-03
fixed Scope and SentryEvent attributes

marandaneto:
removed not used attributes from Scope and removed the inheritance from SentryEvent and duplicated needed events

### DIFF
--- a/sentry-core/src/main/java/io/sentry/Scope.java
+++ b/sentry-core/src/main/java/io/sentry/Scope.java
@@ -1,24 +1,20 @@
 package io.sentry;
 
-import io.sentry.protocol.Request;
-import io.sentry.protocol.SdkVersion;
 import io.sentry.protocol.User;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 public class Scope {
   private SentryLevel level;
   private String transaction;
-  private String environment;
   private User user;
-  private Request request;
-  private SdkVersion sdkVersion;
   private List<String> fingerprint = new ArrayList<>();
-  private CopyOnWriteArrayList<Breadcrumb> breadcrumbs = new CopyOnWriteArrayList<>();
-  private ConcurrentHashMap<String, String> tags = new ConcurrentHashMap<>();
-  private ConcurrentHashMap<String, Object> extra = new ConcurrentHashMap<>();
+  private List<Breadcrumb> breadcrumbs = new CopyOnWriteArrayList<>();
+  private Map<String, String> tags = new ConcurrentHashMap<>();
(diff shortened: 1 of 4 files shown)


sentry-core/src/main/java/io/sentry/Scope.java
```diff
@@ -1,24 +1,20 @@
...
   private String transaction;
-  private String environment;
   private User user;
-  private Request request;
-  private SdkVersion sdkVersion;
   private List<String> fingerprint = new ArrayList<>();
-  private CopyOnWriteArrayList<Breadcrumb> breadcrumbs = new CopyOnWriteArrayList<>();
-  private ConcurrentHashMap<String, String> tags = new ConcurrentHashMap<>();
-  private ConcurrentHashMap<String, Object> extra = new ConcurrentHashMap<>();
+  private List<Breadcrumb> breadcrumbs = new CopyOnWriteArrayList<>();
```
bruno-garcia: For the private field I'd leave the actual type explicit here.

sentry-core/src/main/java/io/sentry/Scope.java
```diff
@@ -52,22 +40,6 @@ public void setUser(User user) {
     this.user = user;
   }
 
-  public Request getRequest() {
```
bruno-garcia: Why did Scope lose request?

sentry-core/src/main/java/io/sentry/SentryEvent.java
```diff
@@ -19,6 +26,16 @@
...
   private SentryValues<SentryThread> threads;
   private SentryValues<SentryException> exceptions;
+  private SentryLevel level;
+  private String transaction;
+  private String environment;
+  private User user;
+  private Request request;
+  private SdkVersion sdkVersion;
+  private List<String> fingerprint = new ArrayList<>();
+  private List<Breadcrumb> breadcrumbs = new CopyOnWriteArrayList<>();
```
bruno-garcia: Event doesn't need `CopyOnWrite`. It is not thread-safe so we can use simple(cheaper) types here.

sentry-core/src/main/java/io/sentry/protocol/SdkVersion.java
```diff
@@ -1,13 +1,12 @@
...
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 public class SdkVersion {
   private String name;
   private String version;
-  private CopyOnWriteArrayList<Package> packages = new CopyOnWriteArrayList<>();
+  private List<Package> packages = new CopyOnWriteArrayList<>();
```
bruno-garcia: What's the motivation of using the abstract type here?

sentry-core/src/main/java/io/sentry/Scope.java
marandaneto: @HazAT said 

> and remove environment, request, sdkVersion from the scope

PR #10, opened 2019-10-03
android serializer

bruno-garcia:


sentry-android/src/main/java/io/sentry/internal/AndroidSerializer.java
```diff
@@ -0,0 +1,16 @@
+package io.sentry.internal;
+
+import io.sentry.ISerializer;
+import io.sentry.SentryEvent;
+
+public class AndroidSerializer implements ISerializer {
+  @Override
+  public SentryEnvelope DeserializeEnvelope(String envelope) {
```
marandaneto: name it as camelCase

sentry-android/src/main/java/io/sentry/internal/AndroidSerializer.java
```diff
@@ -0,0 +1,16 @@
...
+    return null;
+  }
+
+  @Override
+  public String Serialize(SentryEvent event) {
```
marandaneto: name it as camelCase

sentry-core/src/main/java/io/sentry/ISerializer.java
```diff
@@ -0,0 +1,8 @@
+package io.sentry;
+
+import io.sentry.internal.SentryEnvelope;
+
+public interface ISerializer {
+  SentryEnvelope DeserializeEnvelope(String envelope);
```
marandaneto: name it as camelCase

sentry-android/src/main/java/io/sentry/internal/AndroidSerializer.java
bruno-garcia: I'll never learn

PR #13, opened 2019-10-07
feat: missing proto types

bruno-garcia:


sentry-core/src/main/java/io/sentry/protocol/Device.java
```diff
@@ -1,5 +1,239 @@
...
+
+  private String name;
+  private String manufacturer;
+  private String brand;
+  private String family;
+  private String model;
+  private String modelId;
+  private String architecture;
+  private Short batteryLevel;
+  private Boolean isCharging;
```
marandaneto: is there a reason why always using Objects instead of primitive types?

sentry-core/src/main/java/io/sentry/protocol/Device.java
```diff
@@ -1,5 +1,239 @@
...
+
+  public Short getBatteryLevel() {
+    return batteryLevel;
+  }
+
+  public void setBatteryLevel(Short batteryLevel) {
+    this.batteryLevel = batteryLevel;
+  }
+
+  public Boolean getCharging() {
```
marandaneto: what about `isCharging()` ?

sentry-core/src/main/java/io/sentry/protocol/Device.java
bruno-garcia: A primitive type can't be null so is charging by default would be `false` and if that was explicitly set or not would be unknown.
The protocol is composed of _all optional_ fields so it's important that by default everything is null. Serializing an instance of `SentryEvent` where no setter was called should result in a JSON with `event_id` and `timestamp` only even though the server would accept it even if it was simply `{}`.

marandaneto: got it, makes sense, thanks

sentry-core/src/main/java/io/sentry/protocol/Device.java
```diff
@@ -1,5 +1,239 @@
...
+  public TimeZone getTimezone() {
+    return timezone;
+  }
+
+  public void setTimezone(TimeZone timezone) {
+    this.timezone = timezone;
+  }
+
+  public enum DeviceOrientation {
+    Portrait,
```
marandaneto: enums in java are common to be UPPERCASE like PORTRAIT, LANDSCAPE

PR #14, opened 2019-10-07
🚧 Experimental rewrite of event transport

metlos:
After having looked at `sentry-dotnet` and how the connection handling and event sending is done in `sentry-java` I came up with this new implementation.

Many of the features from `sentry-java` are supported (tests peding ;) ) but obviously we are going to lack the stability of the tried implementation. On the other hand I think some of the approaches are for the better, e.g. the replacement of the `LockDownManager` with `ITransportGate` that is supposed to just listen for network availability.

The choice of a sub-class of ScheduledThreadPoolExecutor for implementing the retry logic is actually a little bit questionable and maybe could be replaced with some completable future dance. But I found the logic still simple enough to warrant this lower-level implementation.


sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
```diff
@@ -0,0 +1,186 @@
...
+    connection.setConnectTimeout(connectionTimeout);
+    connection.setReadTimeout(readTimeout);
+
+    if (bypassSecurity && connection instanceof HttpsURLConnection) {
+      ((HttpsURLConnection) connection).setHostnameVerifier((__, ___) -> true);
+    }
+
+    connection.connect();
+
+    try (OutputStream outputStream = connection.getOutputStream()) {
```
ninniuz: *try-with-resources* is supported on all Android API levels ONLY when using AGP >= 3.0.0

sentry-core/src/main/java/io/sentry/transport/RetryingThreadPoolExecutor.java
```diff
@@ -0,0 +1,215 @@
...
+    public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+      AttemptedRunnable<?> that = (AttemptedRunnable<?>) o;
+      return task.equals(that.task);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(task);
```
ninniuz: This is not supported on Android API < 19

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
marandaneto: we are using AGP 3.5.0

ninniuz: This is true. But as far as I know the Java library (or an aar) will include Java bytecode classes which will be dexed when the consumer APK is built. What if the consumer APK is built with a version of AGP that does not use D8?
I might be completely wrong here, but it should be pretty easy to check.

ninniuz: Small test from a Java class in an android library aar built with AGP 3.5.0.
Added a try-with-resources statement and checked the java bytecode in the relative class file:
```
     144: invokevirtual #124                // Method it/immobiliare/android/Test.close:()V
     147: goto          211
     150: astore        4
     152: aload_3
     153: aload         4
     155: invokevirtual #126                // Method java/lang/Throwable.addSuppressed:(Ljava/lang/Throwable;)V
     158: goto          211
     161: aload_2
     162: invokevirtual #124                // Method it/immobiliare/android/Test.close:()V
```

As you can see javac adds this line:
`     155: invokevirtual #126                // Method java/lang/Throwable.addSuppressed:(Ljava/lang/Throwable;)V`
which needs D8 desugaring to work on API < 19.

marandaneto: it's worth to be checked, but I believe as soon as the lib is built, it doesn't matter who consumes it, our AGP/D8/R8 already desugared it, but if somebody else adds it as a dependency and don't use D8/R8/AGP >= 3, might be a problem, otherwise all Java 8 features will have the same problem.

sentry-android/src/main/java/io/sentry/android/AndroidSerializer.java
```diff
@@ -11,7 +13,5 @@ public SentryEnvelope deserializeEnvelope(String envelope) {
   }
 
   @Override
-  public String serialize(SentryEvent event) {
-    return null;
-  }
+  public void serialize(SentryEvent event, Writer writer) throws IOException {}
```
marandaneto: why do we need a `Writer` here?

sentry-core/src/main/java/io/sentry/transport/AlwaysAllowedTransportGate.java
```diff
@@ -0,0 +1,12 @@
+package io.sentry.transport;
+
+/**
+ * A dummy implementation of the {@link ITransportGate} interface that always allows sending the
+ * events.
+ */
+public class AlwaysAllowedTransportGate implements ITransportGate {
```
marandaneto: would it be possible to not be public, but only visible within the package? not only for this class, but for most of them

sentry-core/src/main/java/io/sentry/transport/AsyncConnection.java
```diff
@@ -0,0 +1,108 @@
...
+      ITransportGate transportGate,
+      int maxRetries,
+      IBackOffIntervalStrategy backOffIntervalStrategy,
+      IEventCache eventCache,
+      SentryOptions options) {
+    this.transport = transport;
+    this.transportGate = transportGate;
+    this.eventCache = eventCache;
+    this.options = options;
+    this.executor =
```
marandaneto: I'd say to create a method to set it and keep the constructor a bit cleaner

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
```diff
@@ -0,0 +1,181 @@
...
+import java.io.OutputStreamWriter;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.Proxy;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.net.URLConnection;
+import java.nio.charset.StandardCharsets;
+import java.util.function.Consumer;
```
marandaneto: Added in API level 24 Consumer

sentry-core/src/test/java/io/sentry/transport/HttpTransportTest.kt
```diff
@@ -0,0 +1,130 @@
...
+        val serializer = mock<ISerializer>()
+        var proxy: Proxy? = null
+        var requestUpdater: Consumer<URLConnection> = Consumer { ; }
+        var connectionTimeout = 1000
+        var readTimeout = 500
+        var bypassSecurity = false
+        val connection = mock<HttpURLConnection>()
+
+        init {
+            whenever(connection.outputStream).thenReturn(mock())
```
marandaneto: would it not be better `@BeforeTest`?

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
marandaneto: @ninniuz thanks for checking it out, I also checked at https://android.googlesource.com/platform/external/desugar/+/master/java/com/google/devtools/build/android/desugar/runtime/ThrowableExtension.java
After `aar` is assembled using AGP 3 >= 0 with D8/R8, but correct me if I am wrong, does the client still needs to have desugaring enabled? don't think so, right?

sentry-android/src/main/java/io/sentry/android/AndroidSerializer.java
metlos: Because serializing to a string only to write that string into a writer is non-optimal when you can serialize directly to the writer.

metlos: Because serializing to a string only to write that string to a writer is kinda suboptimal when you can serialize directly into a writer. E.g. https://github.com/getsentry/sentry-android/pull/14/files#diff-eec907f5d2232bf99fecad94aaf3e4beR100

sentry-core/src/main/java/io/sentry/transport/AlwaysAllowedTransportGate.java
metlos: Well, this does not need to be here at all actually. It's just a dummy impl to illustrate the usage. I will remove it.

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
metlos: How does one find this out without checking every bloody method/class in all of `java` package? :)

sentry-core/src/test/java/io/sentry/transport/HttpTransportTest.kt
metlos: I thought the whole `Fixture` thing is a kind of before-test. But I will adapt to whatever style you prefer.

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
marandaneto: static code analysis is the way to go (eg lint), but this is a java library and it doesn't seem to check `NewApi` as @ninniuz pointed out. might be needed to create a custom lint rule, I can look into it

sentry-core/src/test/java/io/sentry/transport/HttpTransportTest.kt
marandaneto: Fixture tests are a fixed state, but if we need to extend those tests and somehow need the state of the mocks to be cleaned, the testing lifecycle @Before/After would be better

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
marandaneto: Screenshot 2019-10-09 at 13 52 55
you may use IntelliJ for now till we find a way to break the build based on those warnings.

bruno-garcia: We need to solve this. Worst case we just make `sentry-core` an android library for now. Or at least one build configuration for CI that builds it as an android library

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
ninniuz: The bytecode I was showing you was extracted from a `.class` file inside an `aar`, so I can state that desugaring is not applied.

I think your statement `otherwise all Java 8 features will have the same problem.` is correct.
It might be worth to check what happens on a test apk which uses AGP < 3.0.0

metlos: Huh, for reasons unbeknownst to me, I had Android support disabled in Intellij. I now see the squiggly lines! :)

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
metlos: Huh, for reasons unbeknownst to me, I had Android support disabled in Intellij. I now see the squiggly lines! :)

sentry-core/src/test/java/io/sentry/transport/HttpTransportTest.kt
metlos: JUnit4 does create a new instance for each test run I think.

sentry-core/src/main/java/io/sentry/transport/TransportResult.java
```diff
@@ -0,0 +1,85 @@
...
+
+import io.sentry.ISerializer;
+import io.sentry.SentryEvent;
+
+/**
+ * A result of {@link ITransport#send(SentryEvent, ISerializer)}. Note that this class is
+ * intentionally not subclassable and has only two factory methods to capture the 2 possible states
+ * - success or error.
+ */
+public abstract class TransportResult {
```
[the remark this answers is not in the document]
metlos: Well, this class is not inheritable, because it has a private constructor (intentionally). It just a little bit of mostly unnecessary gymnastics, but with this approach, the "success" instance is a singleton and doesn't need to hold any data. The "error" results don't need to hold a boolean saying that they're not a success.

I can simplify this by doing what you suggest at the cost of ever so slightly higher memory consumed by the instances.

sentry-core/src/test/java/io/sentry/transport/HttpTransportTest.kt
marandaneto: it does, if we don't wanna play with the values of the `Fixture`, all good.
if somehow we would like to change some values of the `Fixture` and run tests in parallel/out of order, might find flaky tests, so in that case, `@Before` will make the setup and clean the states each test, but again, if it's not the case in this class, I'm happy to keep it like this

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
marandaneto: @metlos just submitted a PR https://github.com/getsentry/sentry-android/pull/23 so after merging, build will fail when using APIs which is not retro compatible

sentry-core/src/test/java/io/sentry/transport/HttpTransportTest.kt
metlos: So let's leave this in and fix it if we ever find a problem you describe (which I think would only ever manifest if we changed the test instantiation to be per-class, not per test-method, and we had parallel test runs, which I know is possible with JUnit5 but I am not sure about JUnit4). But IMHO in the case of per-class instantiation and parallel test invocations, a @Before method doesn't make a difference - you'd still be mutating shared data.

I will try to read up on the fixture approach because frankly it's new to me.

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
metlos: Actually not, because the lint doesn't support the Android API checks for java libraries.

metlos: Ah, I see what you did there :) the `checkDependencies` will cause the android part to fail if core contains API errors?

marandaneto: yep, so we are safe to write code which is compatible with android API 14 and if not, build will fail

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
```diff
@@ -0,0 +1,192 @@
...
+    this.connectionConfigurator = connectionConfigurator;
+    this.serializer = options.getSerializer();
+    this.connectionTimeout = connectionTimeout;
+    this.readTimeout = readTimeout;
+    this.options = options;
+    this.sentryUrl = new URI(options.getDsn()).toURL();
+    this.bypassSecurity = bypassSecurity;
+  }
+
+  // visible for testing
```
marandaneto: whats about creating an annotation for it? we might need in many other places, like
https://android.googlesource.com/platform/frameworks/support/+/ee8deb2/annotations/src/android/support/annotation/VisibleForTesting.java

marandaneto: we already have 2 `io.sentry.util.NonNull`

bruno-garcia: Does it work for `java-library` or requires `android-library`? I say that because this is on `sentry-core`

sentry-core/src/main/java/io/sentry/transport/TransportResult.java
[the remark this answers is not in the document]
ninniuz: The approach used by @metlos is the *"default"* way to get a sealed class in Java. I think it is fine.
It would have been nicer to read in Kotlin 😛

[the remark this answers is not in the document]
marandaneto: yeah, I'd agree with @ninniuz that's the java verbose way of doing it.

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
metlos: I think the main reason for having them (without an annotation processor) is that the IDE (or at least Intellij) can be set up to take advantage of the annos when offering intentions (at least for `@Nullable` and `@NonNull`). The `@VisibleForTesting` (originally from Guava) is/would be a documentation-only annotation, although there are plugins for e.g. Findbugs that make use of it. As far as I looked Intellij doesn't support it (the "make access weaker" intention has no possibility to be affected by an annotation).

sentry-core/src/main/java/io/sentry/transport/TransportResult.java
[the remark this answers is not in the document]
bruno-garcia: That's solves it then. Thanks everyone for chipping in

sentry-core/src/main/java/io/sentry/transport/HttpTransport.java
metlos: Note that I added `@VisibleForTesting` in this PR.

sentry-core/src/main/java/io/sentry/core/transport/AsyncConnection.java
```diff
@@ -0,0 +1,160 @@
...
+      if (!executor.awaitTermination(1, TimeUnit.MINUTES)) {
+        log(
+            options.getLogger(),
+            SentryLevel.WARNING,
+            "Failed to shutdown the async connection async sender within 1 minute. Trying to force it now.");
+        executor.shutdownNow();
+      }
+      transport.close();
+    } catch (InterruptedException e) {
+      // ok, just give up then...
```
marandaneto: should we log it?

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -0,0 +1,194 @@
...
+            log(
+                DEBUG,
+                "Event '"
+                    + event.getEventId()
+                    + "' was rejected by the Sentry server due to a filter.");
+          }
+        }
+        return TransportResult.error(retryAfterMs, responseCode);
+      } catch (IOException responseCodeException) {
+        // pass
```
marandaneto: should we log it?

sentry-core/src/main/java/io/sentry/core/util/VisibleForTesting.java
```diff
@@ -0,0 +1,20 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Annotates a declaration as visible solely for testing purposes. Such declaration should not be
+ * used in production code.
+ */
+@Documented
+@Target({TYPE, METHOD, FIELD, CONSTRUCTOR})
+@Retention(CLASS)
```
marandaneto: should it not be `SOURCE` ?

sentry-core/src/main/java/io/sentry/core/ILogger.java
```diff
@@ -1,7 +1,42 @@
...
+   * message only if logger is not null.
+   *
+   * @param logger the logger or null if none found
+   * @param level the log level
+   * @param message the message
+   * @param args the formatting arguments
+   * @see #log(SentryLevel, String, Object...)
+   */
+  static void log(@Nullable ILogger logger, SentryLevel level, String message, Object... args) {
+    if (logger != null) {
```
bruno-garcia: Although not documented, it's nice to be able to swap loggers at runtime.
Would that work with?:

```java
ILogger currentLogger = logger;
if (currentLogger != null) {
  currentLogger.log(level, message, throwable);
}

File: sentry-core/src/main/java/io/sentry/core/SentryClient.java [hunk 14]
```diff
@@ -1,12 +1,26 @@
...
+import io.sentry.core.transport.IBackOffIntervalStrategy;
+import io.sentry.core.transport.IConnectionConfigurator;
+import io.sentry.core.transport.IEventCache;
+import io.sentry.core.transport.ITransportGate;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
 
 public class SentryClient implements ISentryClient {
+  static final String SENTRY_PROTOCOL_VERSION = "6";
```
```suggestion
  static final String SENTRY_PROTOCOL_VERSION = "7";
```

bruno-garcia:
I've left a couple of small things but I'm all for merging this in and opening new PRs for any change.

sentry-core/src/main/java/io/sentry/core/util/VisibleForTesting.java
metlos: Depends :) If we want some static analysis that is able to work on the bytecode level then we need `CLASS` so that the annotation is retained in the class file. `SOURCE` would make it only available during compilation and not put into the compiled jarfile. E.g. if we were linting `sentry-core` only as a dependency, as we are atm, then we have to rely on the bytecode analysis.

sentry-core/src/main/java/io/sentry/core/ILogger.java
metlos: Yes, I understood that intention. That's why this is a static method that you pass a logger to. So that the next time, you're free to pass a different logger. That is the reason why the classes in transport don't hold on to the logger directly but rather hold the options instance and call this method as:

```java
log(options.getLogger(), ...);
```

so as long as there is some code holding the options instance able to set different loggers, this method support swapping the loggers.

bruno-garcia: much nicer, thanks!

sentry-core/src/main/java/io/sentry/core/util/VisibleForTesting.java
marandaneto: mmm, that's interesting, but I'm not an expert about this matter, just did a quick research and it seems that `android lint` does source code analyses and `findbugs` does bytecode analyses.
could you just double-check it? I'm not sure how it works when a java lib. is a dependency for an Android lib.
if it should be CLASS, all good :D

ninniuz: AFAIK Lint is not applied on bytecode but on an AST derived from source code files.
Guava as well has got a `VisibleForTesting` annotation which has `CLASS` retention.

However, even if our annotation is set to have retention `CLASS` I don't think the current available Lint check would take it into account (unless there is a way to configure it) nor any static analysis tool will be able to process it by default.

So now I am wondering: what is the advantage of adding such annotations in our package scope?

metlos: I think you're right - lint seems to only be using the sources of java libs available during the build. I tried depending in `android-core` on a test lib I created which uses `Objects.hash()` in a function called from `android-core`. The lib is built separately and so lint doesn't have access to its sources. And our build passes even though we're clearly using a lib that can break at runtime due to API incompat. . 

So in another words lint cannot check that e.g. some random java lib from maven central doesn't use an API that is not present in the required API level. That is kinda disappointing and limits what `checkDependencies` can be used for. I learned something today :)

metlos: > So now I am wondering: what is the advantage of adding such annotations in our package scope?

Given the findings above, I don't think it makes sense to have that annotation or indeed just have it at `SOURCE` level so that it doesn't "pollute" the compiled artifacts.

ninniuz: Yeah I would keep for internal reference and set it to `SOURCE` retention.

ninniuz: > So in another words lint cannot check that e.g. some random java lib from maven central doesn't use an API that is not present in the required API level. That is kinda disappointing and limits what `checkDependencies` can be used for. I learned something today :)

Absolutely, `checkDependencies` is basically for dependent library projects (in Gradle terminology)

PR #15, opened 2019-10-08
Added gson and adapters for SentryId and Date

marandaneto:


build.gradle.kts
```diff
@@ -64,3 +64,7 @@ spotless {
         ktlint()
     }
 }
+
+tasks.named("build") {
+    dependsOn(":spotlessApply")
```
bruno-garcia: ❤️

sentry-android/src/main/java/io/sentry/android/SentryIdDeserializerAdapter.java
```diff
@@ -0,0 +1,42 @@
...
+      throws JsonParseException {
+    return json == null ? null : new SentryId(fromStringSentryId(json.getAsString()));
+  }
+
+  private UUID fromStringSentryId(String sentryIdString) {
+    if (sentryIdString == null) {
+      return null;
+    }
+
+    if (sentryIdString.length() == 32) {
```
bruno-garcia: Maybe call a function on `SentryId` that holds this logic

sentry-android/src/main/java/io/sentry/android/AndroidSerializer.java
```diff
@@ -1,17 +1,37 @@
...
+
   @Override
   public SentryEnvelope deserializeEnvelope(String envelope) {
-    return null;
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public SentryEvent deserializeEvent(String envelope) {
+    return gson.fromJson(envelope, SentryEvent.class);
```
metlos: Do we ever deserialize in the client?

bruno-garcia: We will for communication between sentry-android and sentry-native.

marandaneto: because of NDK, yes

bruno-garcia:
> Since we're actually serializing a single type - SentryEvent - do we really have to use the whole complex reflection based GSON machinery? What about just manually using the JsonWriter from GSON and use its primitives to manually serialize the SentryEvent?

We have a few requirements coming from the integration with sentry-native. One of them is to keep around unknown fields in some sort of map. That's so that when we get an event from disk written by sentry-native, the data shoudn't be lost due to deserializing/serializing again to send to Sentry.

We might optimize in the future for usages of the package without NDK in which case we could not bring the whole `gson` package but only vendor in the few classes needed to serialize only as you suggested.

build.gradle.kts
marandaneto: this peace of love is already merged into master

marandaneto:
> How come you are not exposing proguard config file via `consumerProguardFiles`?

for some reason, travis ci is hanging when I set `consumerProguardFiles`
https://github.com/getsentry/sentry-android/pull/15/commits/a0171047c825089d5aac120f50736589d7333fa7 so I rollback it and have as a task on our backlog to not forget it.

PR #18, opened 2019-10-08
ndk gradle project

bruno-garcia:
Attempt to supersede this branch with #52


sentry-android-ndk/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest package="io.sentry.android.ndk">
```
marandaneto: maybe `io.sentry.android-ndk`?

sentry-android-ndk/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,3 @@
...
+</manifest>
```
marandaneto: <manifest package="io.sentry.android.ndk" `/>`

sentry-android/build.gradle.kts
```diff
@@ -68,6 +68,7 @@ android {
 
 dependencies {
     api(project(":sentry-core"))
+    api(project(":sentry-android-ndk"))
```
marandaneto: should not it be `implementation` and expose it through java/JNI classes?

marandaneto: also, I might do the opposite, `sentry-android-ndk` have `sentry-android` so who don't use NDK doesn't need to carry it's dependencies

tonyo:
Current limitation: `sentry-android-ndk/sentry-native` should be a symlink to `sentry-native`

sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java
```diff
@@ -5,5 +5,14 @@
     System.loadLibrary("sentry");
   }
 
-  public static void init() {}
+  static {
+    System.loadLibrary("sentry-android");
```
ninniuz: You may want to consider usage of [ReLinker](https://github.com/KeepSafe/ReLinker) (suggestion [here](https://developer.android.com/training/articles/perf-jni#native-libraries))

sentry-android/build.gradle.kts
bruno-garcia: Discussed offline:

sentry-android depends on:
1. sentry-android-core
2. sentry-android-ndk

sentry-android-core depends on:
1. sentry-core

sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java
bruno-garcia: That does sound like a good suggestion. Although I'm not sure it's worth the extra dependency. /cc @mitsuhiko

sentry-android/build.gradle.kts
ninniuz: Would it be possible to have `sentry-android-ndk` depend on e.g. `sentry-core` and have the user add either `io.sentry.android:sentry-android` and/or `io.sentry.android:sentry-android-ndk` (whatever the group and artifacts name will be) depending on what she needs?

bruno-garcia: Yes but we need to prioritize easy docs. A single install instruction for basic usage that exposes all features. Then further down the docs (for those 10% that read that far) will be details on the project structure and how they can drop NDK support, some pros and cons etc. Often we print install instructions on flyers and take to conferences, it must be small and include all features when possible.

sentry-android-ndk/build.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+                includeCompileClasspath = true
+            }
+        }
+
+        minSdkVersion(21)
+        externalNativeBuild {
+            cmake {
+                arguments.add(0, "-DANDROID_STL=c++_static")
+                arguments.add(0, "-DCMAKE_VERBOSE_MAKEFILE:BOOL=ON")
+
```
bruno-garcia:
```suggestion
```

sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java
marandaneto: that makes sense, even the official documentation assumes bugs on older versions and tag Relinker as a solver, I'd go for it, ReLinker also has Asynchronous loading which could faster App. init.

buildSrc/src/main/java/Config.kt
```diff
@@ -11,7 +11,7 @@ object Config {
         private val sdkVersion = 29
 
         val buildToolsVersion = "29.0.2"
-        val minSdkVersion = 14
+        val minSdkVersion = 21
```
marandaneto: I'd not do it, set version 21 directly then instead

sentry-android-ndk/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+cmake_minimum_required(VERSION 3.4.1)
+project("sentry-android")
+
+# Adding NDCrash submodule subdirectory
```
marandaneto:
```suggestion
# Adding sentry-native submodule subdirectory
```

sentry-android-ndk/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+add_subdirectory("sentry-native")
+
+# NDCrash submodule include path.
```
marandaneto:
```suggestion
# sentry-native submodule include path.
```

sentry-android-ndk/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+include_directories("sentry-native/include")
+
+# JNDCrash source code.
```
marandaneto:
```suggestion
# sentry-native source code.
```

sentry-android-ndk/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+file(GLOB SENTRY_ANDROID_SOURCES ${CMAKE_SOURCE_DIR}/src/main/jni/*.c)
+
+# Adding dynamic library and linking it with log and NDCrash.
```
marandaneto:
```suggestion
# Adding dynamic library and linking it with log and sentry-native.
```

buildSrc/src/main/java/Config.kt
bruno-garcia: I was hoping we'd figure out a way to keep this on 14 when possible. Perhaps only NDK support means requiring 21?
We're not requiring 21+ across the board, are we?

sentry-sample/src/main/AndroidManifest.xml
```diff
@@ -23,5 +23,7 @@
     <meta-data android:name="io.sentry.dsn" android:value="https://key@sentry.io/123" />
     <meta-data android:name="io.sentry.debug" android:value="true" />
   </application>
+  <uses-permission android:name="android.permission.INTERNET" />
+  <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
marandaneto: why do we need that? `INTERNET` should be enough

ninniuz: Why don't you move required permissions to the library project so they are inherited automatically by consuming apps via manifest merging?

buildSrc/src/main/java/Config.kt
marandaneto: actually the client App. overrides it, but this is important to our static code analysis

marandaneto: I didn't investigate it deeper yet, but I'd say that if we need to set API up to 21, we might be using ndk features that were not retro compatible and we would need to do some workarounds if we wanna fully support older versions.
ps: https://android.googlesource.com/platform/ndk/+/master/docs/user/common_problems.md#Target-API-Set-Higher-Than-Device-API

sentry-sample/src/main/AndroidManifest.xml
marandaneto: or that, but not sure if this is a good idea, if I am not wrong, the current SDK crashes if there's no `INTERNET` permission, so we would need to write down the behavior change.

ninniuz: Is that because you do not want to add permissions on behalf of the app?

marandaneto: @ninniuz yes, only `INTERNET` I don't see as a problem at all, but I'm pretty sure we are gonna require more, then it could be a problem.

bruno-garcia: Once we merge #32 we can delete these two lines.

marandaneto: merged

bruno-garcia: tks

sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java
bruno-garcia: Let's do this on a separate PR to avoid keeping this open for longer

bruno-garcia: I've opened an issue internally to track this

buildSrc/src/main/java/Config.kt
marandaneto: also found this one https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md
which says:

> For example, applications with a minSdkVersion of 20 should use API 19 for their NDK target.
it means that if we wanna support v14, we should set v13

sentry-android-ndk/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+        minSdkVersion(21)
+        externalNativeBuild {
+            cmake {
+                arguments.add(0, "-DANDROID_STL=c++_static")
+                arguments.add(0, "-DCMAKE_VERBOSE_MAKEFILE:BOOL=ON")
+            }
+        }
+        ndk {
+            abiFilters("x86", "armeabi-v7a", "x86_64", "arm64-v8a")
```
marandaneto: if we wanna support non-NEON devices, we might need to look into it.
https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md#NEON

marandaneto: https://android-developers.googleblog.com/2019/10/introducing-ndk-r21-our-first-long-term.html NDK 21 has NEON enabled by default, so I'd say it's gonna be fixed by itself when we bump up.

PR #19, opened 2019-10-09
implemented custom gson adapters for unknown fields

marandaneto:


sentry-android/src/main/java/io/sentry/android/adapters/UnknownAdapter.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    return object;
+  }
+
+  private List<String> getKeys(T clazz) {
+    List<String> keys = null;
+
+    if (clazz instanceof SentryEvent) {
+      keys = SentryEvent.KEYS;
```
bruno-garcia: Since we need to be aware of the types here anyway I'd suggest we don't go the inheritance path (i.e forcing the protocol to inherit form `UnknownField` but instead just have a static map:
`Map<Class, Map<String>>`. Here we can get get the right map by the `clazz`.
We can also create the map lazily via reflection on the first call to this adapter to avoid having to keep a manual list of the properties. What do you think?

sentry-android/src/main/java/io/sentry/android/AndroidSerializer.java
```diff
@@ -18,6 +19,11 @@
           .registerTypeAdapter(SentryId.class, new SentryIdDeserializerAdapter())
           .registerTypeAdapter(Date.class, new DateSerializerAdapter())
           .registerTypeAdapter(Date.class, new DateDeserializerAdapter())
+          .registerTypeAdapter(
+              SentryEvent.class,
+              new UnknownAdapter<
```
ninniuz: My eyes are crossing on those two lines 🤣

sentry-android/src/main/java/io/sentry/android/adapters/UnknownAdapter.java
```diff
@@ -0,0 +1,60 @@
...
+import io.sentry.protocol.SentryId;
+import java.lang.reflect.Type;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class UnknownAdapter<T extends UnknownField> implements JsonDeserializer<T> {
+
+  private Gson gson =
```
ninniuz: This partially mirrors the `Gson` instance you register in `AndroidSerializer` and seems to me like a possible cause of troubles (those two instances must be kept in sync)

sentry-android/src/main/java/io/sentry/android/AndroidSerializer.java
```diff
@@ -18,6 +19,11 @@
           .registerTypeAdapter(SentryId.class, new SentryIdDeserializerAdapter())
           .registerTypeAdapter(Date.class, new DateSerializerAdapter())
           .registerTypeAdapter(Date.class, new DateDeserializerAdapter())
+          .registerTypeAdapter(
```
ninniuz: You should evaluate usage of `TypeAdapterFactory` here instead of manually adding type adapters for each class that must support unknown properties.

sentry-android/src/main/java/io/sentry/android/adapters/UnknownAdapter.java
marandaneto: jep, but they cannot have the same adapters nor JsonDeserializationContext or it will be an infinite loop

sentry-android/src/main/java/io/sentry/android/adapters/UnknownAdapter.java
marandaneto: it seems a good idea, but I'd rather have a look at the new PR, this is a crap hack

marandaneto:
in favor of https://github.com/getsentry/sentry-android/pull/22

PR #20, opened 2019-10-09
feat: main package includes all

bruno-garcia:
Docs will include installation steps that are simple and provide all features.
That will be through the package `sentry-android`.
It does mean that users will be bringing NDK support and in many cases are not interested and would prefer dropping the extra package.

For that reason, _advanced_ docs will exemplify how to install individual packages like `sentry-android-core` or `sentry-android-ndk` only.


bruno-garcia:
I'm planning not to squash these two commits so we can see the rename commit separately and keep history

marandaneto:
> Can you please detail the purpose of `SentryInitProvider`?
> Is it supposed to automatically start and configure `Sentry`? If yes, how can a developer prevent this to happen?

the idea is that the SDK auto init. and if you have an empty DSN or not set at all, SDK will be "disabled"

bruno-garcia:
Makes sense for us to allow disabling this, so users can do `init` themselves and configure things like `BeforeSend`. We could also allow some parts of SentryOptions to be modified after `init` although I can see this might not fly as it'll diverge a bit from the rest of the SDKs

ninniuz:
Easier way is to check the app's manifest for a `meta-data` node that dictates if the init provider should be disabled.

bruno-garcia:
We have a similar thing in aspnet core. The framework will auto init but if u use the framework co fig to set auto init to false, it's up to use code to do it.
We can sort that in a separate pr

PR #22, opened 2019-10-09
added gson adapter factory for unknown fields

marandaneto:


sentry-android/src/test/java/io/sentry/android/AndroidSerializerTest.kt
```diff
@@ -61,13 +63,77 @@ class AndroidSerializerTest {
...
+        sentryEvent.unknown["object"] = objects
+
+        val actual = serializer.serialize(sentryEvent)
+
+        val expected = "{\"unknown\":{\"object\":{\"boolean\":true,\"int\":1}}}"
+
+        assertEquals(expected, actual)
+    }
+
     private fun generateEmptySentryEvent(): SentryEvent {
```
ninniuz: What happens in case you have to serialise an instance of `SentryEvent()` with default values? What is the expected result?

sentry-core/src/main/java/io/sentry/SentryEvent.java
```diff
@@ -30,6 +30,7 @@
   private List<Breadcrumb> breadcrumbs = new ArrayList<>();
   private Map<String, String> tags = new HashMap<>();
   private Map<String, Object> extra = new HashMap<>();
+  private Map<String, Object> unknown = new HashMap<>();
```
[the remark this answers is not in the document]
marandaneto: yes, it could, but as the other Map were initialized, and this is greenfield, not sure if backend wants it empty or null at all, will check that.

sentry-android/src/test/java/io/sentry/android/AndroidSerializerTest.kt
marandaneto: those tests were focused on the gson adapters, like ISO Date format, SentryId and unknown objects, the full parsing and its tests will come as soon as we define our `SentryEnvelope` class, I could isolate the tests, but I wanted to test on top of `AndroidSerializer`

sentry-core/src/main/java/io/sentry/SentryEvent.java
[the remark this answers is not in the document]
bruno-garcia: If we can get away initializing these only when needed, that'd be ideal.
The event is not thread safe so makes it easier

sentry-android/src/test/java/io/sentry/android/AndroidSerializerTest.kt
ninniuz: Ok!

PR #24, opened 2019-10-10
feat: Sentry and Hub APIs

bruno-garcia:
Add the public API surface to `Sentry`. That will force down the implementation throughout the SDK like `Hub`, `SentryClient` scope management etc.


sentry-core/src/main/java/io/sentry/Hub.java
```diff
@@ -36,6 +36,32 @@ public void close() {
...
+  public void withScope(ScopeCallback callback) {}
+
+  @Override
+  public void configureScope(ScopeCallback callback) {}
+
+  @Override
+  public void bindClient(SentryClient client) {}
+
+  @Override
+  public void flush(int timeoutMills) {}
```
marandaneto:
```suggestion
  public void flush(long timeoutMills) {}
```

sentry-core/src/main/java/io/sentry/IHub.java
```diff
@@ -1,5 +1,23 @@
...
+
+  void popScope();
+
+  void withScope(ScopeCallback callback);
+
+  void configureScope(ScopeCallback callback);
+
+  void bindClient(SentryClient client);
+
+  void flush(int timeoutMills);
```
marandaneto:
```suggestion
  void flush(long timeoutMills);
```

HazAT:
@bruno-garcia Good job so far 👍

PR #25, opened 2019-10-10
set proguard rules to be merged

marandaneto:


sentry-sample/build.gradle.kts
```diff
@@ -27,7 +27,7 @@ android {
             isMinifyEnabled = false
         }
         getByName("release") {
-            proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
+            consumerProguardFiles("proguard-rules.pro")
```
ninniuz: Do we need `consumerProguardFiles` on the application module? No one is going to consume its Proguard rules

marandaneto: opss, it was supposed to be sentry-android, will fix it, good catch, thanks

PR #30, opened 2019-10-14
opt in and out to auto-init SDK

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
Added the option via Manifest conf. to disable the auto-init of the Sentry SDK, so before fully initing the SDK we check this flag and don't do anything, the user may init. using a custom Application class.

the user may want to init. the SDK by code and pass custom parameters based on buildType/flavors and so on.

Mocking the manifest value and unit testing.

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
This should be written in the setup documentation

### DIFF
--- a/sentry-android-core/build.gradle.kts
+++ b/sentry-android-core/build.gradle.kts
@@ -24,10 +24,9 @@ android {
     }
 
     buildTypes {
-        getByName("debug") {
-            isMinifyEnabled = false
-        }
+        getByName("debug")
         getByName("release") {
+//            isMinifyEnabled = true // Do we want to enable it for our lib?
             consumerProguardFiles("proguard-rules.pro")
         }
     }
--- a/sentry-android-core/build.gradle.kts
+++ b/sentry-android-core/build.gradle.kts
@@ -26,7 +26,6 @@ android {
     buildTypes {
         getByName("debug")
         getByName("release") {
-//            isMinifyEnabled = true // Do we want to enable it for our lib?
             consumerProguardFiles("proguard-rules.pro")
         }
     }
--- a/sentry-android-core/src/main/AndroidManifest.xml
+++ b/sentry-android-core/src/main/AndroidManifest.xml
@@ -6,7 +6,6 @@
     <provider
       android:name=".SentryInitProvider"
       android:authorities="${applicationId}.SentryInitProvider"
-      android:exported="false"
-      android:initOrder="2147483647" />
+      android:exported="false" />
   </application>
 </manifest>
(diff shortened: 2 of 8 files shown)


sentry-android-core/build.gradle.kts
```diff
@@ -24,10 +24,9 @@ android {
     }
 
     buildTypes {
-        getByName("debug") {
-            isMinifyEnabled = false
-        }
+        getByName("debug")
         getByName("release") {
+//            isMinifyEnabled = true // Do we want to enable it for our lib?
```
ninniuz: Does it really make sense? I mean, it is an open source project and the source code is available online. Also you would have to carefully craft your proguard file (in terms of code roots) to make sure code that is used only when integrated into a consumer app does not get deleted by minification.
I don't think it is worth the pain.

marandaneto: @bruno-garcia can you give input about that?

marandaneto: @ninniuz I agree

bruno-garcia: IMHO it's best we don't minify be default. We're not trying to obfuscate anything, code is OSS and if they proguard their app it will do so fo the lib.

bruno-garcia: Also agreed. But don't we need to proguard file anyway? To be brought into the app and included when the app is minified?

sentry-android-core/src/main/AndroidManifest.xml
```diff
@@ -6,7 +6,6 @@
     <provider
       android:name=".SentryInitProvider"
       android:authorities="${applicationId}.SentryInitProvider"
-      android:exported="false"
-      android:initOrder="2147483647" />
+      android:exported="false" />
```
bruno-garcia: Curious what happens if we don't define an order here?
We'd like our code to init early so we can capture error during initialization

marandaneto: the system is gonna define the `android:initOrder` and it's gonna be before the `Application` anyway.
this seems to be more like a magic number

sentry-android-core/build.gradle.kts
marandaneto: this is already being done by https://github.com/getsentry/sentry-android/blob/587ed63fb465e11b9ee6271eaeccf0422f21f7be/sentry-android-core/build.gradle.kts#L31

PR #31, opened 2019-10-14
No need to generate BuildConfig for libs

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
libs don't need to generate BuildConfig and lints check don't need to run over Release type as well every time.

Speed up the build.
It doesn't generate the useless BuildConfig for libs.

Debugging and changing the unit tests to not use BuildConfig.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/build.gradle.kts
+++ b/sentry-android-core/build.gradle.kts
@@ -69,6 +69,16 @@ android {
     lintOptions {
         isWarningsAsErrors = true
         isCheckDependencies = true
+
+        // We run a full lint analysis as build part in CI, so skip vital checks for assemble tasks.
+        isCheckReleaseBuilds = false
+    }
+
+    // replace with https://issuetracker.google.com/issues/72050365 once released.
+    libraryVariants.all {
+        generateBuildConfigProvider?.configure {
+            enabled = false
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


sentry-android-core/build.gradle.kts
```diff
@@ -69,6 +69,16 @@ android {
     lintOptions {
         isWarningsAsErrors = true
         isCheckDependencies = true
+
+        // We run a full lint analysis as build part in CI, so skip vital checks for assemble tasks.
+        isCheckReleaseBuilds = false
```
bruno-garcia: I'm not sure I follow: What do you mean by 'vital checks'?
Right now in CI we run `gradle build`, so this task will run all that's needed?

marandaneto: text comes from https://android.googlesource.com/platform/tools/base/+/studio-master-dev/build-system/gradle-core/src/main/java/com/android/build/gradle/internal/dsl/LintOptions.java#407
linting should be for debugging purpose and we are gonna fix issues before running a release build.

bruno-garcia: that makes sense.
Just the name of the property is confusing.

Does it mean then: Lint in Release but not in Debug mode?

marandaneto: @bruno-garcia it's because the lint task is called Vital
if you run `./gradlew tasks`
> lintVitalRelease - Runs lint on just the fatal issues in the release build.

PR #32, opened 2019-10-14
feat: Require internet permission

bruno-garcia:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Won't require user configuration in their app to add the permissions required by Sentry's SDK.

## :pencil: Checklist

- [ ] All tests passing

### DIFF
--- a/sentry-android-core/src/main/AndroidManifest.xml
+++ b/sentry-android-core/src/main/AndroidManifest.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest package="io.sentry.android.core"
           xmlns:android="http://schemas.android.com/apk/res/android">
+  <uses-permission android:name="android.permission.INTERNET" />
   <application>
     <!-- 'android:authorities' must be unique in the device, across all apps -->
     <provider


sentry-android-core/src/main/AndroidManifest.xml
```diff
@@ -9,4 +9,5 @@
       android:exported="false"
       android:initOrder="2147483647" />
   </application>
+  <uses-permission android:name="android.permission.INTERNET" />
```
ninniuz: I don't know why but you are asking apps to also include `android.permission.ACCESS_NETWORK_STATE` in the [docs](http://tools.android.com/tech-docs/new-build-system/user-guide/manifest-merger)

bruno-garcia: I also don't know why, that's why I leaving it out :)
Don't recall code in our 1.x SDK that checks for network status.

For 2.0 we'll need to update the docs and write a migration plan.

bruno-garcia:
@marandaneto I saw the build is broken now.
I should run a local build more often 👿

sentry-android-core/src/main/AndroidManifest.xml
marandaneto: if we would like to have network status, we would need it.
current SDK also does that https://github.com/getsentry/sentry-java/blob/a5b4fd12b65f46a593f0f8317dbfad4aac95097b/sentry-android/src/main/java/io/sentry/android/event/helper/AndroidEventBuilderHelper.java#L546

bruno-garcia: Fair. If we add then this check we add also the permission then.
Let's discuss whether to do that on a separate PR

PR #34, opened 2019-10-15
feat: hub holds stack of scope/client

bruno-garcia:
This PR is more of the _placeholder_ ones.
It shows the structure we expect in the SDKs where Hub holds a stack of client/scope.
Push/Pop scope will add/remove.
There's always one scope (aka root scope).
Client takes an event and an optional scope, which gets merged with the event.

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Hub.java
+++ b/sentry-core/src/main/java/io/sentry/core/Hub.java
@@ -1,8 +1,8 @@
 package io.sentry.core;
 
 import io.sentry.core.protocol.SentryId;
-import java.util.ArrayDeque;
 import java.util.Deque;
+import java.util.concurrent.ConcurrentLinkedDeque;
 
 public class Hub implements IHub {
 
@@ -19,8 +19,7 @@ public class Hub implements IHub {
   private SentryId lastEventId;
   private SentryOptions options;
   private volatile boolean isEnabled;
-  private Deque<StackItem> stack = new ArrayDeque<>();
-  private final Object lock = new Object();
+  private final Deque<StackItem> stack = new ConcurrentLinkedDeque<>();
 
   public Hub(SentryOptions options) {
     this.options = options;
@@ -41,52 +40,42 @@ public class Hub implements IHub {
   @Override
   public SentryId captureEvent(SentryEvent event) {
     SentryId sentryId;
-    synchronized (lock) {
-      StackItem item = stack.peek();
-      sentryId = item.client.captureEvent(event, item.scope);
-      this.lastEventId = sentryId;
-    }
+    StackItem item = stack.peek();
+    sentryId = item.client.captureEvent(event, item.scope);
+    this.lastEventId = sentryId;
     return sentryId;
   }
 
   @Override
   public SentryId captureMessage(String message) {
     SentryId sentryId;
-    synchronized (lock) {
-      StackItem item = stack.peek();
-      sentryId = item.client.captureMessage(message, item.scope);
-      this.lastEventId = sentryId;
-    }
+    StackItem item = stack.peek();
+    sentryId = item.client.captureMessage(message, item.scope);
+    this.lastEventId = sentryId;
     return sentryId;
   }
 
   @Override
   public SentryId captureException(Throwable throwable) {
     SentryId sentryId;
-    synchronized (lock) {
-      StackItem item = stack.peek();
-      sentryId = item.client.captureException(throwable, item.scope);
-      this.lastEventId = sentryId;
-    }
+    StackItem item = stack.peek();
+    sentryId = item.client.captureException(throwable, item.scope);
+    this.lastEventId = sentryId;
     return sentryId;
   }
 
   @Override
   public void close() {
-    synchronized (lock) {
-      // Close the top-most client
-      StackItem item = stack.peek();
-      item.client.close(options.getShutdownTimeout());
-    }
+    // Close the top-most client
+    StackItem item = stack.peek();
+    item.client.close(options.getShutdownTimeout());
     isEnabled = false;
   }
 
   @Override
   public void addBreadcrumb(Breadcrumb breadcrumb) {
-    synchronized (lock) {
-      StackItem item = stack.peek();
-      item.scope.addBreadcrumb(breadcrumb);
-    }
+    StackItem item = stack.peek();
+    item.scope.addBreadcrumb(breadcrumb);
   }
 
   @Override
@@ -96,18 +85,16 @@ public class Hub implements IHub {
 
   @Override
   public void pushScope() {
-    synchronized (lock) {
-      StackItem item = stack.peek();
-      Scope clone = item.scope.clone();
-      StackItem newItem = new StackItem(item.client, clone);
-      stack.push(newItem);
-    }
+    StackItem item = stack.peek();
+    Scope clone = item.scope.clone();
+    StackItem newItem = new StackItem(item.client, clone);
+    stack.push(newItem);
   }
 
   @Override
   public void popScope() {
-    synchronized (lock) {
-      // Don't drop the root scope
+    // Don't drop the root scope
+    synchronized (stack) {
       if (stack.size() != 1) {
         stack.pop();
       }
@@ -118,10 +105,8 @@ public class Hub implements IHub {
   public void withScope(ScopeCallback callback) {
     pushScope();
     try {
-      synchronized (lock) {
-        StackItem item = stack.peek();
-        callback.run(item.scope);
-      }
+      StackItem item = stack.peek();
+      callback.run(item.scope);
     } finally {
       popScope();
     }
@@ -129,26 +114,20 @@ public class Hub implements IHub {
 
   @Override
   public void configureScope(ScopeCallback callback) {
-    synchronized (lock) {
-      StackItem item = stack.peek();
-      callback.run(item.scope);
-    }
+    StackItem item = stack.peek();
+    callback.run(item.scope);
   }
 
   @Override
   public void bindClient(SentryClient client) {
-    synchronized (lock) {
-      StackItem item = stack.peek();
-      item.client = client;
-    }
+    StackItem item = stack.peek();
+    item.client = client;
   }
 
   @Override
   public void flush(long timeoutMills) {
-    synchronized (lock) {
-      StackItem item = stack.peek();
-      item.client.flush(options.getShutdownTimeout());
-    }
+    StackItem item = stack.peek();
+    item.client.flush(options.getShutdownTimeout());
   }
 
   @Override
--- a/sentry-core/src/main/java/io/sentry/core/Hub.java
+++ b/sentry-core/src/main/java/io/sentry/core/Hub.java
@@ -6,9 +6,9 @@ import java.util.Deque;
 
 public class Hub implements IHub {
 
-  private class StackItem {
-    public ISentryClient client;
-    public Scope scope;
+  private static final class StackItem {
+    ISentryClient client;
+    Scope scope;
 
     public StackItem(ISentryClient client, Scope scope) {
       this.client = client;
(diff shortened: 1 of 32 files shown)


sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -1,13 +1,34 @@
...
+
+    public StackItem(ISentryClient client, Scope scope) {
+      this.client = client;
+      this.scope = scope;
+    }
+  }
+
   private SentryOptions options;
   private volatile boolean isEnabled;
+  private Deque<StackItem> stack = new ArrayDeque<>();
```
marandaneto: why don't we use an eg. `ConcurrentLinkedDeque` instead of creating a `locker`?

bruno-garcia: I'm holding a lock while I do more than one operation with it like:

```java
    synchronized (lock) {
      StackItem item = stack.peek();
      item.client.close(options.getShutdownTimeout());
    }
```

So that it synchronizes with `bindClient`.

Scope and Client must be thread-safe so the question is: Do we care changes to client/scope are atomic?

bruno-garcia:
> I'd prefer to use a thread-safe version of Deque
It only makes sense if we drop the `lock`. To drop the lock we have to decide whether it's OK or not to deal with scope and client when either could change at anytime. What's the side effect?

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -18,52 +39,122 @@ public boolean isEnabled() {
 
   @Override
   public SentryId captureEvent(SentryEvent event) {
-    return null;
+    SentryId sentryId;
+    synchronized (lock) {
+      StackItem item = stack.peek();
+      sentryId = item.client.captureEvent(event, item.scope);
+      item.scope.setLastEventId(sentryId);
```
bruno-garcia: If access through `Sentry.pushScope` for example, a new Hub instance will exist per thread. So it wouldn't be possible for a new scope to be pushed while another thread is executing line 45, before setting the last event Id here.
It could be a problem is a user creates a instance of a Hub and passes it around. In that case `lastEventId` could potentially be set on the wrong scope.

@HazAT, @mitsuhiko  is this just something we accept the risk and mitigate with docs? Or do we need to account for and lock `client.captureEvent + scope.setLastEventId`?

sentry-core/src/main/java/io/sentry/core/Hub.java
marandaneto: got it, I'd not know how to answer it.

In case we need to lock the operations itself, instead of creating our own lock Object, depending on the use-case, these ones could be a bit more performatic `ReentrantLock/ReadWriteLock`

sentry-core/src/main/java/io/sentry/core/Hub.java
HazAT: For other SDKs `lastEventId` lives directly on the Hub,
So even there we can have race conditions which is fine.

bruno-garcia: Things that usually don't go together are race condition and 'this is fine'.
But I won't fight anyone.
For as long as it doesn't blow up I don't care right here. 
Concurrent collection is ok. RWlock, full lock would give contention under high load though

bruno-garcia: Given that a single `Hub` is _not supposed_ to be used by more than 1 thread I would not optimize for correctness of calls to `Hub.something` to be atomic with regards to client and scope calls but rather just make sure it is thread-safe. Meaning calls to Hub done concurrently will not throw.

That said, IMO a immutable or thread-safe collection is good enough here.

sentry-core/src/main/java/io/sentry/core/Hub.java
bruno-garcia: Given that a single Hub is not supposed to be used by more than 1 thread I would not optimize for correctness of calls to Hub.something to be atomic with regards to client and scope pair but rather just make sure it is thread-safe. Meaning calls to Hub done concurrently will not throw.

That said, IMO a immutable or thread-safe collection is good enough here.

sentry-core/src/main/java/io/sentry/core/Hub.java
bruno-garcia: Agreed offline to move lastEventId to the Hub instead of the scope.

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -1,13 +1,34 @@
...
+    public Scope scope;
+
+    public StackItem(ISentryClient client, Scope scope) {
+      this.client = client;
+      this.scope = scope;
+    }
+  }
+
+  private SentryId lastEventId;
   private SentryOptions options;
```
metlos: Should be `final`.

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -1,13 +1,34 @@
 package io.sentry.core;
 
 import io.sentry.core.protocol.SentryId;
+import java.util.Deque;
+import java.util.concurrent.ConcurrentLinkedDeque;
 
 public class Hub implements IHub {
+
+  private class StackItem {
+    public ISentryClient client;
```
metlos: If this is supposed to be used from multiple threads and is mutable, then it should be `volatile`. Since this class is internal, `public` doesn't make sense and it can be left either package private or even `private`.

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -1,13 +1,34 @@
 package io.sentry.core;
 
 import io.sentry.core.protocol.SentryId;
+import java.util.Deque;
+import java.util.concurrent.ConcurrentLinkedDeque;
 
 public class Hub implements IHub {
+
+  private class StackItem {
```
metlos: Should be `static final` IMHO to not needlessly store the reference to the hub internally and as a slight hint to the jdk that it can potentially do more magic with it, respectively.

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -18,52 +39,100 @@ public boolean isEnabled() {
...
-    return null;
+    SentryId sentryId;
+    StackItem item = stack.peek();
+    sentryId = item.client.captureException(throwable, item.scope);
+    this.lastEventId = sentryId;
+    return sentryId;
   }
 
   @Override
   public void close() {
```
metlos: This impl makes sense in the light of `IHub` extending `ISentryClient` but is highly confusing otherwise :confused:

sentry-core/src/main/java/io/sentry/core/ISentryClient.java
```diff
@@ -1,15 +1,24 @@
 package io.sentry.core;
 
 import io.sentry.core.protocol.SentryId;
+import io.sentry.core.util.Nullable;
 
 public interface ISentryClient {
   boolean isEnabled();
 
   SentryId captureEvent(SentryEvent event);
```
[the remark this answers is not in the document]
bruno-garcia: Sadly we can't: `Default method required API level 24 (current min is 15)`

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -1,13 +1,34 @@
...
+    public ISentryClient client;
+    public Scope scope;
+
+    public StackItem(ISentryClient client, Scope scope) {
+      this.client = client;
+      this.scope = scope;
+    }
+  }
+
+  private SentryId lastEventId;
```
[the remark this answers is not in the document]
bruno-garcia: This value is only really relevant to the single thread using the current hub. Would adding volatile only help with making a value change visible to other threads (no CPU cache involved)? In that case it wouldn't matter.

Also: This value will not be copied over to a _clone_ of the Hub.

sentry-core/src/main/java/io/sentry/core/Hub.java
bruno-garcia: Hub is still the object one uses to dispatch calls to the client that involve Scope.
Closing the hub just means closing the client it holds. 

`Sentry.close` holds `Hub`s. The alternative would be to expose the `SentryClient` a `Hub` holds so one can call `close`.

Why would you say it's confusing?

sentry-core/src/main/java/io/sentry/core/Hub.java
bruno-garcia: Unfortunately ConcurrentLinkedDeque requires API 21

sentry-core/src/main/java/io/sentry/core/Hub.java
metlos: Because hub holds a stack of scopes yet close is only closing the client of the top of the stack. If all scopes are supposed to reference the same client, then I think the StackItem should NOT be static as I suggested above, but should also NOT have the client as an explicit field.

bruno-garcia:
> Frankly, I don't completely understand why the scopes form a stack yet using a concurrent queue (or locking in the previous commits) suggests that this is supposed to be used from multiple threads and the scope is supposed to "logically flow" with the business logic.
> To my mind if they're supposed to be used concurrently and hub is essentially a singleton, then the scopes should form a tree. That is because, especially in case of Rx and other more functional approaches, it is common to start multiple independent stages of business logic concurrently and only continue with the "parent" once they're all finished. If each of the stages should start its new scope, I am not sure how that could be modelled with a singleton Hub with just a stack of scopes.

`Hub` is expected to be used by a single thread but we can't guarantee users will do that. Nothing stops someone from passing the reference to another thread and share it although documented not to do that. Also, if no Hub exists in thread local storage (TLS) we'll have to clone it, and for that alone we need to have it thread-safe so we can read its state and copy over.

We have a scope stack per Hub, and a Hub per thread. The hub per thread happens because we clone the hub when you try to access it, if it doesn't yet exist in TLS. A scope is not needed to be pushed when a new thread is spawned since that thread will have its own Hub.

> Or is the fact that the `Hub` is cloneable hinting at the fact that the "branches" of the above mentioned tree are composed of hubs instead of just scopes? If that is the case though then I'd argue there is no need to use concurrent datastructures or locking on `Hub`.

Yes I guess I confirmed this in the comment above. Also that we need hub to be thread-safe so we can clone it (which means reading its fields while another thread might be writing to it). We need it not to blow up.

> Even if I'm completely off with the above, I think the impl requires some more attention still. Please see my comments in the code.

Thanks for your review

sentry-core/src/main/java/io/sentry/core/ISentryClient.java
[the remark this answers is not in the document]
bruno-garcia: @ninniuz pointed out it's actually possible. So I'm very happy I'll be able to drop a bunch of stuff from interfaces now :)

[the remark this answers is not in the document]
bruno-garcia: Thanks @ninniuz and @metlos !

bruno-garcia:
This PR is now missing the synchronization on the Hub.

We need to make sure it can be cloned while it's being mutated.
Ideally a immutable collection although seems there's nothing OOTB
ReaderWriter lock now seems like an option.

sentry-core/src/main/java/io/sentry/core/Hub.java
marandaneto: @bruno-garcia LinkedBlockingDeque requires API 9, it's also thread-safe afaik

sentry-core/src/main/java/io/sentry/core/ISentryClient.java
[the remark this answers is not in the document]
marandaneto: default methods have desugaring already

sentry-core/src/main/java/io/sentry/core/Hub.java
bruno-garcia: Thanks. I'll look at it

sentry-core/src/main/java/io/sentry/core/ISentryClient.java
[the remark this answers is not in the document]
bruno-garcia: That's great. Let's give it a shot

PR #35, opened 2019-10-16
Device context

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/build.gradle.kts
+++ b/sentry-android-core/build.gradle.kts
@@ -73,11 +73,11 @@ android {
     }
 
     // replace with https://issuetracker.google.com/issues/72050365 once released.
-    libraryVariants.all {
-        generateBuildConfigProvider?.configure {
-            enabled = false
-        }
-    }
+//    libraryVariants.all {
+//        generateBuildConfigProvider?.configure {
+//            enabled = false
+//        }
+//    }
 }
 
 dependencies {
--- a/sentry-android-core/build.gradle.kts
+++ b/sentry-android-core/build.gradle.kts
@@ -21,6 +21,8 @@ android {
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
 
         defaultPublishConfig = "${Config.Flavors.production}Release"
+
+        versionName = "$version"
     }
 
     buildTypes {
--- a/sentry-android-core/build.gradle.kts
+++ b/sentry-android-core/build.gradle.kts
@@ -73,13 +73,6 @@ android {
         // We run a full lint analysis as build part in CI, so skip vital checks for assemble tasks.
         isCheckReleaseBuilds = false
     }
-
-    // replace with https://issuetracker.google.com/issues/72050365 once released.
-//    libraryVariants.all {
-//        generateBuildConfigProvider?.configure {
-//            enabled = false
-//        }
-//    }
 }
 
 dependencies {
--- a/sentry-android-core/src/main/AndroidManifest.xml
+++ b/sentry-android-core/src/main/AndroidManifest.xml
@@ -1,14 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest package="io.sentry.android.core"
           xmlns:android="http://schemas.android.com/apk/res/android">
-    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
-    <uses-permission android:name="android.permission.INTERNET" />
-<!--    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />-->
+  <uses-permission android:name="android.permission.INTERNET"/>
   <application>
     <!-- 'android:authorities' must be unique in the device, across all apps -->
     <provider
       android:name=".SentryInitProvider"
       android:authorities="${applicationId}.SentryInitProvider"
-      android:exported="false" />
+      android:exported="false"/>
   </application>
 </manifest>
--- a/sentry-android-core/src/main/AndroidManifest.xml
+++ b/sentry-android-core/src/main/AndroidManifest.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest package="io.sentry.android.core"
           xmlns:android="http://schemas.android.com/apk/res/android">
-  <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
+    <uses-permission android:name="android.permission.INTERNET" />
   <application>
     <!-- 'android:authorities' must be unique in the device, across all apps -->
     <provider
--- a/sentry-android-core/src/main/AndroidManifest.xml
+++ b/sentry-android-core/src/main/AndroidManifest.xml
@@ -3,6 +3,7 @@
           xmlns:android="http://schemas.android.com/apk/res/android">
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
     <uses-permission android:name="android.permission.INTERNET" />
+<!--    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />-->
   <application>
     <!-- 'android:authorities' must be unique in the device, across all apps -->
     <provider
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -1,20 +1,520 @@
 package io.sentry.android.core;
 
+import static android.content.Context.ACTIVITY_SERVICE;
+
+import android.app.ActivityManager;
 import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.content.pm.PackageInfo;
+import android.content.res.Configuration;
+import android.net.ConnectivityManager;
+import android.os.*;
+import android.util.DisplayMetrics;
 import io.sentry.core.EventProcessor;
 import io.sentry.core.SentryEvent;
+import io.sentry.core.SentryLevel;
+import io.sentry.core.SentryOptions;
+import io.sentry.core.protocol.Device;
+import io.sentry.core.protocol.SdkVersion;
+import io.sentry.core.util.Objects;
+import java.io.File;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.Locale;
+import java.util.TimeZone;
 
 public class DefaultAndroidEventProcessor implements EventProcessor {
   Context context;
+  SentryOptions options;
 
-  public DefaultAndroidEventProcessor(Context context) {
-    if (context == null) throw new IllegalArgumentException("The application context is required.");
+  public DefaultAndroidEventProcessor(Context context, SentryOptions options) {
+    Objects.requireNonNull(context, "The application context is required.");
+    Objects.requireNonNull(options, "The SentryOptions is required.");
+
     this.context = context.getApplicationContext();
+    this.options = options;
   }
 
   @Override
   public SentryEvent process(SentryEvent event) {
-    // TODO: Do cool android things.
+    if (event.getSdkVersion() == null) {
+      SdkVersion sdkVersion = new SdkVersion();
+      sdkVersion.setName("android");
+      // version, don't we have getRelease() for that?
+      // packages, which packages?
+
+      event.setSdkVersion(sdkVersion);
+    }
+    PackageInfo packageInfo = getPackageInfo();
+    if (packageInfo != null) {
+      if (event.getRelease() == null) {
+        event.setRelease(packageInfo.packageName + "-" + packageInfo.versionName);
+      }
+      if (event.getDist() == null) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+          event.setDist(Long.toString(packageInfo.getLongVersionCode()));
+        } else {
+          event.setDist(getVersionCode(packageInfo));
+        }
+      }
+    }
+    // sentry interface? we dont need anymore, only user
+
+    // proguard uuids? check if we need sentry-debug-meta.properties
+    //
+
+    if (event.getContexts().getDevice() == null) {
+      event.getContexts().setDevice(getDevice());
+    }
+
     return event;
+  }
+
+  @SuppressWarnings("deprecation")
+  private String getVersionCode(PackageInfo packageInfo) {
+    return Integer.toString(packageInfo.versionCode);
+  }
+
+  @SuppressWarnings("deprecation")
+  private String getAbi() {
+    return Build.CPU_ABI;
+  }
+
+  /**
+   * Return the Application's PackageInfo if possible, or null.
+   *
+   * @return the Application's PackageInfo if possible, or null
+   */
+  private PackageInfo getPackageInfo() {
+    try {
+      return context.getPackageManager().getPackageInfo(context.getPackageName(), 0);
+    } catch (Exception e) {
+      options.getLogger().log(SentryLevel.ERROR, "Error getting package info.", e);
+      return null;
+    }
+  }
+
+  @SuppressWarnings("ObsoleteSdkInt")
+  private void setArchitectures(Device device) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+      device.setArchitectures(Build.SUPPORTED_ABIS);
+    } else {
+      device.setArchitecture(getAbi());
+    }
+  }
+
+  @SuppressWarnings("ObsoleteSdkInt")
+  private Long getMemorySize(ActivityManager.MemoryInfo memInfo) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
+      return memInfo.totalMem;
+    }
+    // TODO: find a way?
+    return null;
+  }
+
+  // we can get some inspiration here
+  // https://github.com/flutter/plugins/blob/master/packages/device_info/android/src/main/java/io/flutter/plugins/deviceinfo/DeviceInfoPlugin.java
+  private Device getDevice() {
+    Device device = new Device();
+    // name of what? maybe from the BluetoothAdapter
+    device.setManufacturer(Build.MANUFACTURER);
+    device.setBrand(Build.BRAND);
+    device.setFamily(getFamily());
+    device.setModel(Build.MODEL);
+    device.setModelId(Build.ID); // or DISPLAY which is user friendly id
+    setArchitectures(device);
+
+    Intent batteryIntent = getBatteryIntent();
+    if (batteryIntent != null) {
+      device.setBatteryLevel(getBatteryLevel(batteryIntent));
+      device.setCharging(isCharging(batteryIntent));
+    }
+    device.setOnline(isConnected());
+    device.setOrientation(getOrientation());
+    device.setSimulator(isEmulator());
+
+    ActivityManager.MemoryInfo memInfo = getMemInfo();
+    if (memInfo != null) {
+      device.setMemorySize(getMemorySize(memInfo));
+      device.setFreeMemory(memInfo.availMem);
+      device.setLowMemory(memInfo.lowMemory);
+      // device.setUsableMemory(); // TODO: check that
+      // do we need threshold?
+    }
+
+    StatFs internalStorageStat = getInternalStorageStat();
+    device.setStorageSize(getTotalInternalStorage(internalStorageStat));
+    device.setFreeStorage(getUnusedInternalStorage(internalStorageStat));
+
+    StatFs externalStorageStat = getExternalStorageStat();
+    if (externalStorageStat != null) {
+      device.setExternalStorageSize(getTotalExternalStorage(externalStorageStat));
+      device.setExternalFreeStorage(getUnusedExternalStorage(externalStorageStat));
+    }
+
+    DisplayMetrics displayMetrics = getDisplayMetrics();
+    if (displayMetrics != null) {
+      device.setScreenResolution(getResolution(displayMetrics));
+      device.setScreenDensity(displayMetrics.density);
+      device.setScreenDpi(displayMetrics.densityDpi);
+    }
+
+    device.setBootTime(getBootTime());
+    device.setTimezone(getTimeZone());
+
+    return device;
+  }
+
+  private TimeZone getTimeZone() {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+      LocaleList locales = context.getResources().getConfiguration().getLocales();
+      if (!locales.isEmpty()) {
+        Locale locale = locales.get(0);
+        return Calendar.getInstance(locale).getTimeZone();
+      }
+    }
+    return Calendar.getInstance().getTimeZone();
+  }
+
+  private Date getBootTime() {
+    // should it be a Date?
+    return new Date(
+        System.currentTimeMillis()
+            - SystemClock.elapsedRealtime()
+            + new Date().getTime()); // check Date().getTime() and if its UTC
+  }
+
+  private String getResolution(DisplayMetrics displayMetrics) {
+    // do we need to calculate the density in here?
+    int largestSide = Math.max(displayMetrics.widthPixels, displayMetrics.heightPixels);
+    int smallestSide = Math.min(displayMetrics.widthPixels, displayMetrics.heightPixels);
+    return largestSide + "x" + smallestSide;
+  }
+
+  /**
+   * Get MemoryInfo object representing the memory state of the application.
+   *
+   * @return MemoryInfo object representing the memory state of the application
+   */
+  private ActivityManager.MemoryInfo getMemInfo() {
+    try {
+      ActivityManager actManager = (ActivityManager) context.getSystemService(ACTIVITY_SERVICE);
+      ActivityManager.MemoryInfo memInfo = new ActivityManager.MemoryInfo();
+      if (actManager != null) {
+        actManager.getMemoryInfo(memInfo);
+        return memInfo;
+      }
+      options.getLogger().log(SentryLevel.INFO, "Error getting MemoryInfo.");
+      return null;
+    } catch (Exception e) {
+      options.getLogger().log(SentryLevel.ERROR, "Error getting MemoryInfo.", e);
+      return null;
+    }
+  }
+
+  private Intent getBatteryIntent() {
+    return context.registerReceiver(null, new IntentFilter(Intent.ACTION_BATTERY_CHANGED));
+  }
+
+  /**
+   * Fake the device family by using the first word in the Build.MODEL. Works well in most cases...
+   * "Nexus 6P" -> "Nexus", "Galaxy S7" -> "Galaxy".
+   *
+   * @return family name of the device, as best we can tell
+   */
+  private String getFamily() {
+    try {
+      return Build.MODEL
+          .split(" ")[
+          0]; // might be possible to do it better, should we change that or keep compatibility?
+    } catch (Exception e) {
+      options.getLogger().log(SentryLevel.ERROR, "Error getting device family.", e);
+      return null;
+    }
+  }
+
+  /**
+   * Get the device's current battery level (as a percentage of total).
+   *
+   * @return the device's current battery level (as a percentage of total), or null if unknown
+   */
+  private Float getBatteryLevel(Intent batteryIntent) {
+    try {
+      int level = batteryIntent.getIntExtra(BatteryManager.EXTRA_LEVEL, -1);
+      int scale = batteryIntent.getIntExtra(BatteryManager.EXTRA_SCALE, -1);
+
+      if (level == -1 || scale == -1) {
+        return null;
+      }
+
+      // CHECKSTYLE.OFF: MagicNumber
+      float percentMultiplier = 100.0f;
+      // CHECKSTYLE.ON: MagicNumber
+
+      return ((float) level / (float) scale) * percentMultiplier;
+    } catch (Exception e) {
+      options.getLogger().log(SentryLevel.ERROR, "Error getting device battery level.", e);
+      return null;
+    }
+  }
+
+  /**
+   * Checks whether or not the device is currently plugged in and charging, or null if unknown.
+   *
+   * @return whether or not the device is currently plugged in and charging, or null if unknown
+   */
+  private Boolean isCharging(Intent batteryIntent) {
+    try {
+      int plugged = batteryIntent.getIntExtra(BatteryManager.EXTRA_PLUGGED, -1);
+      return plugged == BatteryManager.BATTERY_PLUGGED_AC
+          || plugged == BatteryManager.BATTERY_PLUGGED_USB;
+    } catch (Exception e) {
+      options.getLogger().log(SentryLevel.ERROR, "Error getting device charging state.", e);
+      return null;
+    }
+  }
+
+  /**
+   * Check whether the application has internet access at a point in time.
+   *
+   * @return true if the application has internet access
+   */
+  private Boolean isConnected() {
+    ConnectivityManager connectivityManager =
+        (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+    if (connectivityManager == null) {
+      options
+          .getLogger()
+          .log(SentryLevel.INFO, "ConnectivityManager is null and cannot check network status");
+      return null;
+    }
+    return getActiveNetworkInfo(connectivityManager);
+    // does the connection type matters? wifi, lte...
+    // do we care about VPNs? getActiveNetworkInfo might return null if VPN doesn't specify its
+    // underlying network
+
+    // when min. API 24, use:
+    // connectivityManager.registerDefaultNetworkCallback(...)
+  }
+
+  @SuppressWarnings("deprecation")
+  private Boolean getActiveNetworkInfo(ConnectivityManager connectivityManager) {
+    // do not import class or deprecation lint will throw
+    android.net.NetworkInfo activeNetwork =
+        connectivityManager.getActiveNetworkInfo(); // it requires ACCESS_NETWORK_STATE
+
+    if (activeNetwork != null) {
+      return activeNetwork.isConnected();
+    }
+    options
+        .getLogger()
+        .log(SentryLevel.INFO, "NetworkInfo is null and cannot check network status");
+    return null;
+  }
+
+  /**
+   * Get the device's current screen orientation.
+   *
+   * @return the device's current screen orientation, or null if unknown
+   */
+  private Device.DeviceOrientation getOrientation() {
+    try {
+      switch (context.getResources().getConfiguration().orientation) {
+        case Configuration.ORIENTATION_LANDSCAPE:
+          return Device.DeviceOrientation.LANDSCAPE;
+        case Configuration.ORIENTATION_PORTRAIT:
+          return Device.DeviceOrientation.PORTRAIT;
+        default:
+          options
+              .getLogger()
+              .log(SentryLevel.INFO, "No device orientation available (ORIENTATION_UNDEFINED)");
+          return null;
+      }
+    } catch (Exception e) {
+      options.getLogger().log(SentryLevel.ERROR, "Error getting device orientation.", e);
+      return null;
+    }
+  }
+
+  /**
+   * Check whether the application is running in an emulator.
+   * https://github.com/flutter/plugins/blob/master/packages/device_info/android/src/main/java/io/flutter/plugins/deviceinfo/DeviceInfoPlugin.java#L105
+   *
+   * @return true if the application is running in an emulator, false otherwise
+   */
+  private Boolean isEmulator() {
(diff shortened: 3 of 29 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+    }
+  }
+
+  private StatFs getInternalStorageStat() {
+    File path = Environment.getDataDirectory(); // /data
+    return new StatFs(path.getPath());
+  }
+
+  private StatFs getExternalStorageStat() {
+    // might need READ_EXTERNAL_STORAGE permission
```
bruno-garcia: Can we test for the permission and only call it if the app has it?
IMHO not worth requesting permission for the library so that we can report this.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+  }
+
+  private OperatingSystem getOperatingSystem() {
+    OperatingSystem os = new OperatingSystem();
+    os.setName("Android");
+    os.setVersion(Build.VERSION.RELEASE);
+    os.setBuild(Build.DISPLAY);
+    os.setKernelVersion(getKernelVersion());
+    os.setRooted(isRooted());
+    //    os.setRawDescription("");  whats that? maybe the const of VERSION_CODES?
```
bruno-garcia: We don't need raw description here since the API gives us all the individual fields.
That was introduced when a framework only gives us a string like `Android 1.3.2 Kernel 4.4 Bla`

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+
+    return os;
+  }
+
+  private App getApp(PackageInfo packageInfo) {
+    App app = new App();
+    app.setIdentifier(packageInfo.packageName);
+    app.setStartTime(
```
bruno-garcia: This is supposed to be the timestamp of when the app started.
If Android doesn't give us that, we can create a static field somewhere in the SDK and set the current time when `Sentry.init` is called.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+        DateUtils
+            .getCurrentDateTime()); // is it App. start? this was not supposed to be a lazy class?
+    //    app.setHash(); whats that?
```
bruno-garcia: @HazAT is this a iOS thing only?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+    //    app.setBuildType(); whats that? might be possible with BuildConfig.BUILD_VARIANT but Apps
+    // side, also for flavor
```
bruno-garcia: Should we expand the protocol to be more Android friendly then?
Add a Flavor and variant field or something?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+    app.setName(getApplicationName());
+    app.setVersion(packageInfo.versionName);
+    app.setBuild(getVersionCode(packageInfo));
+
+    // App name and starttime don't depend on packageInfo, should we move outside of the packageInfo
+    // NPE check like the old the SDK?
```
bruno-garcia: Yeah we could guard each `getProtocolThing` method so we don't drop all context data in case of exceptions.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+  }
+
+  /**
+   * Get the device's current kernel version, as a string. Attempts to read /proc/version, and falls
+   * back to the 'os.version' System Property.
+   *
+   * @return the device's current kernel version, as a string
+   */
+  private String getKernelVersion() {
+    // its possible to try to execute 'uname' and parse it or also another unix commands or even
```
bruno-garcia: expensive stuff like this should be cached on first use. Since we have multiple hubs. Maybe this could be done in a class with a static field that holds the value and a double check lock for reading/initializing?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+      "/system/bin/failsafe/su",
+      "/system/bin/su",
+      "/system/sd/xbin/su",
+      "/system/xbin/daemonsu",
+      "/system/xbin/su"
+    };
+
+    for (String probableRootPath : probableRootPaths) {
+      try {
+        if (new File(probableRootPath).exists()) {
```
bruno-garcia: How much value `isRooted` gives us? This looks really expensive. If we are keeping it we should cache the result on first use.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+    User user = new User();
+
+    user.setId(getAndroidId());
+
+    return user;
+  }
+
+  @SuppressWarnings("HardwareIds")
+  private String getAndroidId() {
+    // Android 29 has changed and -> Avoid using hardware identifiers, find another way in the
```
bruno-garcia: What did Android 29 change? Don't follow why we need to find another way

sentry-android-core/src/main/java/io/sentry/android/core/adapters/DateSerializerAdapter.java
```diff
@@ -8,7 +8,7 @@
 import java.lang.reflect.Type;
 import java.util.Date;
 
-class DateSerializerAdapter implements JsonSerializer<Date> {
+public class DateSerializerAdapter implements JsonSerializer<Date> {
```
bruno-garcia: Adapters need to be public? Is it so that we can move to a different package for organization?

sentry-android-core/src/main/java/io/sentry/android/core/adapters/OrientationDeserializerAdapter.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.Locale;
+
+public class OrientationDeserializerAdapter implements JsonDeserializer<Device.DeviceOrientation> {
+  @Override
+  public Device.DeviceOrientation deserialize(
+      JsonElement json, Type typeOfT, JsonDeserializationContext context)
+      throws JsonParseException {
+    return json == null
+        ? null
+        : Device.DeviceOrientation.valueOf(json.getAsString().toUpperCase(Locale.ROOT));
```
bruno-garcia: if the value in the JSON doesn't match `DeviceOrientation` is this going to throw?
Should we be supressing exception in adapters like this?

sentry-core/src/main/java/io/sentry/core/Dsn.java
```diff
@@ -48,20 +48,20 @@ public Dsn(String dsn) throws InvalidDsnException {
...
+      if (publicKey == null || publicKey.isEmpty()) {
         throw new IllegalArgumentException("Invalid DSN: No public key provided.");
       }
       secretKey = keys.length > 1 ? keys[1] : null;
       String uriPath = uri.getPath();
       int projectIdStart = uriPath.lastIndexOf("/") + 1;
       path = uriPath.substring(0, projectIdStart);
       projectId = uriPath.substring(projectIdStart);
-      if (projectId == null || projectId.length() == 0) {
+      if (projectId.isEmpty()) {
```
bruno-garcia: It can never be null?

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
```diff
@@ -12,10 +12,13 @@
   private String family;
   private String model;
   private String modelId;
-  private String architecture;
-  private Short batteryLevel;
-  private Boolean isCharging;
-  private Boolean isOnline;
+
+  @Deprecated private String architecture; // it was only arch, did we change it?
```
bruno-garcia: It must serialize to `arch` yes.
The serialized names are documented here: https://docs.sentry.io/development/sdk-dev/event-payloads/

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
```diff
@@ -12,10 +12,13 @@
...
+
+  private String[] architectures;
+  private Float batteryLevel;
+  private Boolean isCharging; // charging
```
bruno-garcia: New fields have to be added to the protocol (change semaphore, docs, other SDKs eventually)

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+    this.options = Objects.requireNonNull(options, "The SentryOptions is required.");
   }
 
   @Override
   public SentryEvent process(SentryEvent event) {
-    // TODO: Do cool android things.
+    if (event.getSdkVersion() == null) {
+      SdkVersion sdkVersion = new SdkVersion();
+      sdkVersion.setName("android");
+      // version, don't we have getRelease() for that?
```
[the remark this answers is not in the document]
marandaneto: got it, mabe we would need to generate the `BuildConfig` again, otherwise, we would need to do manually every time

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+      // packages, which packages?
```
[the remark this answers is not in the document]
marandaneto: I see, will check that

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+      event.setSdkVersion(sdkVersion);
+    }
+    if (event.getUser() == null) {
+      event.setUser(getUser());
+    }
+
+    PackageInfo packageInfo = getPackageInfo();
+    if (packageInfo != null) {
+      if (event.getRelease() == null) {
+        event.setRelease(packageInfo.packageName + "-" + packageInfo.versionName);
```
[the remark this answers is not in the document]
marandaneto: I kept the same as before, so it looks like `io.sentry.sample-1.0`, but we also have the same info with `contexts.app.identifier` and `contexts.app.identifier.version` and more

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+    //
+
+    if (event.getContexts().getDevice() == null) {
+      event.getContexts().setDevice(getDevice());
+    }
+    if (event.getContexts().getOperatingSystem() == null) {
+      event.getContexts().setOperatingSystem(getOperatingSystem());
+    }
+
+    // whats about Runtime, Browser, GPU object, do they make sense to Android?
```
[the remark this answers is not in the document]
marandaneto: well `Build.VERSION.RELEASE` it's part of the `OperatingSystem`, does it make sense then?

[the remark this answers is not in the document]
marandaneto: will check about GPU, maybe interesting for games?

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
```diff
@@ -232,6 +260,30 @@ public void setTimezone(TimeZone timezone) {
     this.timezone = timezone;
   }
 
+  public String[] getArchitectures() {
```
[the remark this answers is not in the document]
marandaneto: I'm sending both now, but on our side, it's called `architecture`, should I rename to `arch`?

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
```diff
@@ -26,7 +29,12 @@
   private Long freeStorage;
   private Long externalStorageSize;
   private Long externalFreeStorage;
-  private String screenResolution;
+
+  @Deprecated private String screenResolution;
+
+  private Integer widthPixels;
```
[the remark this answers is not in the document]
marandaneto: I'm sending both, I just think it's more specific

sentry-android-core/src/main/AndroidManifest.xml
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest package="io.sentry.android.core"
           xmlns:android="http://schemas.android.com/apk/res/android">
-  <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
+    <uses-permission android:name="android.permission.INTERNET" />
+<!--    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />-->
```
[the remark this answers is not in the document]
marandaneto: it seems that only checking the `StatFs` (external storage) which is disk size and so on, it's not needed, but I'd like to test more before releasing it, so it's a reminder for me.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+    return getVersionCodeDep(packageInfo);
+  }
+
+  @SuppressWarnings("deprecation")
+  private String getVersionCodeDep(PackageInfo packageInfo) {
+    return Integer.toString(packageInfo.versionCode);
+  }
+
+  @SuppressWarnings("deprecation")
+  private String getAbi() {
```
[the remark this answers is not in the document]
marandaneto: the thing is, this line is `deprecated` and the new value is an `Array`, and I'd not like to annotate the whole method if I inline it, when new deprecation comes out, we will not notice them, there are more cases like this, another option is to create a lint `baseline` file which is ignoring these errors, but we need to recreate this file every time a new suppress should be added, what do you think?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+      device.setArchitecture(supportedAbis[0]);
+      device.setArchitectures(supportedAbis);
+      // we were not checking CPU_ABI2, but I've added to the list now
+    }
+  }
+
+  @SuppressWarnings("ObsoleteSdkInt")
+  private Long getMemorySize(ActivityManager.MemoryInfo memInfo) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
+      return memInfo.totalMem;
```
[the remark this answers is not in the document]
marandaneto: it's the same as before, the documentation doesn't specify it, will look for it.

> The total memory accessible by the kernel. This is basically the RAM size of the device, not including below-kernel fixed allocations like DMA buffers, RAM for the baseband CPU, etc.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+   * Fake the device family by using the first word in the Build.MODEL. Works well in most cases...
+   * "Nexus 6P" -> "Nexus", "Galaxy S7" -> "Galaxy".
+   *
+   * @return family name of the device, as best we can tell
+   */
+  private String getFamily() {
+    try {
+      return Build.MODEL
+          .split(" ")[
+          0]; // might be possible to do it better, should we change that or keep compatibility?
```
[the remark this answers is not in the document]
marandaneto: so the aim here is to find the "Family", which is Nexus/Galaxy and so on, as they have many versions.
eg "Galaxy S7" with the current approach becomes "Galaxy" which is right, but there are cases like this:
eg "Samsung Galaxy S7", so it becomes "Samsung" which is wrong, it should be "Galaxy" again.
some workarounds are to look for the manufacturer name, brand name, model name and ignore them and get the next String, so "Samsung Galaxy S7" will become "Galaxy" because we ignore the 1st String as it's the Build.MANUFACTURER and won't be the family name anyway, got it?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+   */
+  private Boolean isConnected() {
+    ConnectivityManager connectivityManager =
+        (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+    if (connectivityManager == null) {
+      log(SentryLevel.INFO, "ConnectivityManager is null and cannot check network status");
+      return null;
+    }
+    return getActiveNetworkInfo(connectivityManager);
+    // does the connection type matters? wifi, lte...
```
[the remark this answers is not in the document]
marandaneto: I know, but as a developer, I'd like to know if the user was using 2g/3g/4g/lte/wifi...

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+    // do we care about VPNs? getActiveNetworkInfo might return null if VPN doesn't specify its
+    // underlying network
+
+    // when min. API 24, use:
+    // connectivityManager.registerDefaultNetworkCallback(...)
```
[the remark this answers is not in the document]
marandaneto: it's a callback and basically it's listening to a network broadcast when network is up or down, so way more reliable, but we cannot use it, it's a quite new API and won't work for us right now.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: yeah, I'll do it together with https://github.com/getsentry/sentry-android/pull/35#discussion_r335923486 if necessary.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: let me check if it's even possible on the lib. side, these properties with BuildConfig might be available only for the App. so the user will need to set them on BeforeSend

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: `isRooted` is only useful when you rely on features for rooted devices, not for normal applications usually, and yes, a bit expensive, definitely those things should be cached

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: for now, it's too early to deal with this problem, but important to point out.
https://developer.android.com/training/articles/user-data-ids

one important thing is to test permission `READ_PRIVILEGED_PHONE_STATE`

sentry-android-core/src/main/java/io/sentry/android/core/adapters/DateSerializerAdapter.java
marandaneto: yes, just moved for organization, the number of adapters is growing, let's deal with the public access later on

sentry-core/src/main/java/io/sentry/core/Dsn.java
marandaneto: no, otherwise will throw NPE or something else on the line above

sentry-core/src/main/java/io/sentry/core/protocol/Contexts.java
```diff
@@ -27,6 +27,8 @@ public void setBrowser(Browser browser) {
   }
 
   public Device getDevice() {
+    // getter is not working -> Cannot cast com.google.gson.internal.LinkedTreeMap to
```
marandaneto: @bruno-garcia should I change this class with fields? it's gonna solve it

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
marandaneto: yes, so far only architectures and (screenWidthPixels, screenHeightPixels)

sentry-android-core/src/main/AndroidManifest.xml
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest package="io.sentry.android.core"
           xmlns:android="http://schemas.android.com/apk/res/android">
-  <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
```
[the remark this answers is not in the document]
marandaneto: removed `ACCESS_NETWORK_STATE` and will only get `isConnected` if permission is available

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
bruno-garcia: I see so it would be a non breaking change to keep it like this.

Btw `release` will be automatically index in Sentry. The data in `App` is not so it's OK to have it "duplicated".

Often in other SDKs (other than Java) what ppl use for Release is the git commit SHA. Or the app numeric version like _1.0.0_. I haven't seen it yet as a combination of appname-version.
@HazAT Do you have an opinion here?

sentry-android-core/src/main/java/io/sentry/android/core/adapters/OrientationDeserializerAdapter.java
marandaneto: guess so, for all of them

sentry-core/src/main/java/io/sentry/core/protocol/Contexts.java
marandaneto: it seems that we need to have as a `map` because we need to support any type of object that comes from our users.

> Unknown data for the contexts is rendered as a key/value list.
@bruno-garcia is it because of `BeforeSend` as well?

sentry-android-core/src/main/java/io/sentry/android/core/adapters/OrientationDeserializerAdapter.java
bruno-garcia: Yeah I think that'll help debugging the library a lot when weird stuff happens and we get support tickets.

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
bruno-garcia: Cool. Would be nice to get feedback from others on this. @HazAT hasn't got added his review yet

sentry-core/src/main/java/io/sentry/core/protocol/Contexts.java
bruno-garcia: It's a way for users to add their own serializable objects to be displayed in Sentry.
The Contexts interface (like `Extra`) will display your serialized object (JSON) nicely and it's great for debugging

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
marandaneto: ok, I'm not sure how `craft` works, so hard to say the best approach to set our package versions, for now using android-core version and when publishing, we figure it out

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,742 @@
...
+      if (event.getDist() == null) {
+        event.setDist(getVersionCode(packageInfo));
+      }
+      if (event.getContexts().getApp() == null) {
+        event.getContexts().setApp(getApp(packageInfo));
+      }
+    }
+    // sentry interface? we dont need anymore, only user
+
+    // proguard uuids? check if we need sentry-debug-meta.properties
```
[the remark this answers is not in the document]
marandaneto: ok, we have its own task for that

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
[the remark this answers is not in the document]
marandaneto: I renamed to `arch` and `archs`

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
marandaneto: one way requires system admin permission and the other returns null for everything, root might be needed, will create a task to investigate later on.
Runtime doesn't make sense as we have our OS values

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
marandaneto: @bruno-garcia

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
marandaneto: its also bytes like the protocol

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
marandaneto: not doing for now

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: users should set that on BeforeSend if they want to.
expanding the protocol might be the case.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: I think not only what is heavy, but also everything that would not change, let's do it in a different PR

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: same as https://github.com/getsentry/sentry-android/pull/35#discussion_r336448620

sentry-android-core/src/main/java/io/sentry/android/core/adapters/OrientationDeserializerAdapter.java
marandaneto: created an internal task for that

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
marandaneto: will keep compatibility, for now, it will require more tests with different brands/models

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -35,6 +35,11 @@ public SentryClient(SentryOptions options, @Nullable AsyncConnection connection)
   public SentryId captureEvent(SentryEvent event, @Nullable Scope scope) {
     log(options.getLogger(), SentryLevel.DEBUG, "Capturing event: %s", event.getEventId());
 
+    // process events, should it be done before or after BeforeSecondCallback?
```
marandaneto: @bruno-garcia

sentry-core/src/main/java/io/sentry/core/protocol/Runtime.java
```diff
@@ -6,7 +6,8 @@
   private String name;
   private String version;
   private String rawDescription;
-  private String build;
+  private String build; // this is not in the documentation
```
marandaneto: @bruno-garcia

sentry-core/src/main/java/io/sentry/core/protocol/Contexts.java
marandaneto: let's do in a different PR as well, it's not related to this one

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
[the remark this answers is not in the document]
bruno-garcia: The field was renamed? Or it serializes as `arch`? The latter would be my preferred option to keep the class with a more descriptive name

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
HazAT: For now I would leave it like this. Once we revamp the release system we will change this for sure.

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
HazAT: Just fyi, adding new fields here shouldn't be a blocker it will just show up in the UI unformatted.

sentry-android-core/build.gradle.kts
```diff
@@ -73,11 +75,11 @@ android {
...
-    libraryVariants.all {
-        generateBuildConfigProvider?.configure {
-            enabled = false
-        }
-    }
+//    libraryVariants.all {
+//        generateBuildConfigProvider?.configure {
+//            enabled = false
+//        }
+//    }
```
HazAT: Can be removed.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -1,20 +1,769 @@
...
+  /**
+   * Return the Application's PackageInfo if possible, or null.
+   *
+   * @return the Application's PackageInfo if possible, or null
+   */
+  private PackageInfo getPackageInfo() {
+    try {
+      return context.getPackageManager().getPackageInfo(context.getPackageName(), 0);
+    } catch (Exception e) {
+      log(SentryLevel.ERROR, "Error getting package info.", e);
```
HazAT: That counts for the majority of occurrences of `SentryLevel.ERROR`, I would just log a warning instead of an error. Worst case a value in the context is missing but ERROR means really something critical.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
HazAT: Yes probably, also no idea what this is.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
bruno-garcia: I'd say it's good as error. 
For _critical_ things we can use:

https://github.com/getsentry/sentry-android/blob/1bf3fed0ed92394ae8dd05db769e95674c6d3ac5/sentry-core/src/main/java/io/sentry/core/SentryLevel.java#L8
By default, we can have the SDK `debug=true` log out anything level `DEBUG` or higher. But for a dogfooding scenario or users interested in following internal errors of the SDK, they case raise the level to `ERROR` and get these logged out. This is how the .NET SDK works and it's helpful because users do report internal errors of the SDK which otherwise would go unnoticed.

We're stilll missing taking a SentryLevel for the internal logging in the Options object.

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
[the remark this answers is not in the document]
marandaneto: for now, I renamed all the fields to meet the protocol, if we decide to vendor fully on Gson, then we could use `@SerializedName`.
as some fields were different than the protocol and I'd not like to write a custom adapter to each of them.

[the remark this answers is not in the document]
bruno-garcia: sounnds good

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
marandaneto: GPU has it's own task as it requires way more code, Android SDK doesn't have a simple way of how to get it.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: I'd agree with @bruno-garcia if the other unified SDK already works like this.
@bruno-garcia should the log level be a task on asana?

sentry-core/src/main/java/io/sentry/core/protocol/Runtime.java
marandaneto: should I remove it?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
bruno-garcia: done

sentry-core/src/main/java/io/sentry/core/SentryClient.java
bruno-garcia: Let's run event processors first. `BeforeSend` is the last thing we do before sending an event out to Sentry.

sentry-core/src/main/java/io/sentry/core/protocol/Runtime.java
bruno-garcia: I've added this because we report it in .NET:
https://github.com/getsentry/sentry-dotnet-protocol/blob/afd65b92ed7122a628dc3631711b407f56e9a8a8/src/Sentry.Protocol/Context/Runtime.cs#L40-L45

Another one we should "add to the protocol".

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
marandaneto: I got the version dynamically, `sentry-core` and `sentry-android-core` will be always there, for the NDK and integrations we would depend on reflection or SentryOptions.
Could not find how to load maven dependencies dynamically.

sentry-core/src/main/java/io/sentry/core/protocol/Runtime.java
marandaneto: I see, but we already have it on Contexts.app.appBuild

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
bruno-garcia: Got it. Reflection is OK I guess.
This can become a new task so we can close this one.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
bruno-garcia: got it

sentry-core/src/main/java/io/sentry/core/protocol/Contexts.java
bruno-garcia: agreed

PR #36, opened 2019-10-16
feat: before send

bruno-garcia:


sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -26,58 +20,27 @@ public boolean isEnabled() {
     return isEnabled;
   }
 
-  public SentryClient(SentryOptions options) {
+  public SentryClient(SentryOptions options, @Nullable AsyncConnection connection) {
```
marandaneto: would not it be better just to have an overload instead of forcing to pass null if we want to?

bruno-garcia: Great! yes now that I know I can have default implementation :)
Thanks

bruno-garcia: Right! That's a good point.
Thanks

PR #38, opened 2019-10-17
Reading auto-init even before calling Sentry.init(...)

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
Sentry.init(...) will be called only if auto-init is enabled

We check if auto-init is enabled inside of the Sentry.init(...) method, this should be done even before calling it.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
@@ -4,6 +4,7 @@ import android.content.Context;
 import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageManager;
 import android.os.Bundle;
+import io.sentry.core.ILogger;
 import io.sentry.core.SentryLevel;
 import io.sentry.core.SentryOptions;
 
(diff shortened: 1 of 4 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -36,18 +42,19 @@ public static void applyMetadata(Context context, SentryOptions options) {
...
+  public static boolean isAutoInit(Context context, ILogger logger) {
+    if (context == null) throw new IllegalArgumentException("The application context is required.");
+
     boolean autoInit = true;
     try {
       Bundle metadata = getMetadata(context);
       if (metadata != null) {
         autoInit = metadata.getBoolean(AUTO_INIT, true);
-        options.getLogger().log(SentryLevel.DEBUG, "Auto-init: %s", autoInit);
+        logger.log(SentryLevel.DEBUG, "Auto-init: %s", autoInit);
```
bruno-garcia: Do we nave to null check for `logger`? Since this method is public

marandaneto: ideally should not be public, but for now yeah, will do NPE check, good catch.

PR #41, opened 2019-10-21
added the unknown field to all the objects

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
added the unknown field to all the objects of the protocol

if the protocol is extended and all the SDKs are not synced, this will still send the properties over.

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
in the future get rid of IUnknownPropertiesConsumer if we decide to do our own json parser or stick to Gson


marandaneto:
> `Time` is mispelled here (capital `i`)

good catch, thanks

marandaneto:
> Would be good to add some tests though

tests are written on top of AndroidSerializer while parsing SentryEvent.
https://github.com/getsentry/sentry-android/blob/master/sentry-android-core/src/test/java/io/sentry/android/core/AndroidSerializerTest.kt

PR #45, opened 2019-10-21
fix: Fixes for the HTTP transport

bruno-garcia:
* DSN has to be parsed and Sentry URL built
* Timeouts are in millseconds

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/EnvelopeReader.java
+++ b/sentry-core/src/main/java/io/sentry/core/EnvelopeReader.java
@@ -21,9 +21,9 @@ public class EnvelopeReader {
   private final Gson gson =
       new GsonBuilder()
           .registerTypeAdapter(
-              SentryEnvelopeHeader.class, new EnvelopeReader.SentryEnvelopeHeaderAdapter())
+              SentryEnvelopeHeader.class, new SentryEnvelopeHeaderAdapter())
           .registerTypeAdapter(
-              SentryEnvelopeItemHeader.class, new EnvelopeReader.SentryEnvelopeItemHeaderAdapter())
+              SentryEnvelopeItemHeader.class, new SentryEnvelopeItemHeaderAdapter())
           .create();
 
   public @Nullable SentryEnvelope read(InputStream stream) throws IOException {
@@ -36,6 +36,7 @@ public class EnvelopeReader {
     while ((currentLength = stream.read(buffer)) > 0) {
       for (int i = 0; envelopeEndHeaderOffset == -1 && i < currentLength; i++) {
         if (buffer[i] == '\n') {
+          // TODO: IDE warns this is always true, is it?
           if (envelopeEndHeaderOffset == -1) {
             envelopeEndHeaderOffset = streamOffset + i;
             break;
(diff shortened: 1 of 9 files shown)


sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -167,6 +150,21 @@ public TransportResult send(SentryEvent event) throws IOException {
     }
   }
 
+  private void LogErrorInPayload(HttpURLConnection connection) {
```
marandaneto: camel case

sentry-core/src/main/java/io/sentry/core/EnvelopeReader.java
```diff
@@ -36,6 +36,7 @@
     while ((currentLength = stream.read(buffer)) > 0) {
       for (int i = 0; envelopeEndHeaderOffset == -1 && i < currentLength; i++) {
         if (buffer[i] == '\n') {
+          // TODO: IDE warns this is always true, is it?
```
marandaneto: @bruno-garcia can you have a look at it?

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -49,34 +40,33 @@
...
-    this.sentryUrl = new URI(options.getDsn()).toURL();
     this.bypassSecurity = bypassSecurity;
+    this.sentryUrl = sentryUrl;
   }
 
   // giving up on testing this method is probably the simplest way of having the rest of the class
   // testable...
   @VisibleForTesting
   protected HttpURLConnection open(URL url, Proxy proxy) throws IOException {
+    // why do we need url here? its not used
```
marandaneto: @metlos should we remove `url`  as we use only `sentryUrl`?

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -89,6 +79,12 @@ public TransportResult send(SentryEvent event) throws IOException {
     connection.setRequestMethod("POST");
     connection.setDoOutput(true);
     connection.setRequestProperty("Content-Encoding", "UTF-8");
+    connection.setRequestProperty("Content-Type", "application/json");
+    connection.setRequestProperty("Accept", "application/json");
```
marandaneto: @bruno-garcia it works without that, but it's very common to set anyway

sentry-core/src/main/java/io/sentry/core/EnvelopeReader.java
bruno-garcia: would be a different PR eh?

bruno-garcia: by eye I can't see how. buffer gets modified on each call to `read`

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -1,25 +1,16 @@
 package io.sentry.core.transport;
 
 import static io.sentry.core.ILogger.log;
-import static io.sentry.core.SentryLevel.DEBUG;
```
bruno-garcia: I think we need to bring these imports back so spotless let's us thru

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
bruno-garcia: I don't set in .NET, this prob came from `sentry-java`.
No opinion here, could probably get away without it as u suggest.
@HazAT thoughts?

sentry-core/src/main/java/io/sentry/core/EnvelopeReader.java
marandaneto: I really didn't check it, just got the warning from the IDE

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: spotless is good here, is CI failing?

PR #47, opened 2019-10-22
fix: SDK Name

bruno-garcia:


sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -14,7 +14,7 @@ static void init(SentryOptions options, Context context, ILogger logger) {
     options.setLogger(logger);
 
     // TODO this needs to fetch the data from somewhere - defined at build time?
-    options.setSentryClientName("sentry-android/0.0.1");
+    options.setSentryClientName("sentry.java.android/0.0.1");
```
marandaneto: why is it `sentry.java.android`?

bruno-garcia: We name our sdks in the format `sentry.tech` like:
`sentry.dotnet`
`sentry.php`

When we publish packages that are specialization of it, it takes the format:
`sentry.dotnet.aspnetcore`
`sentry.javascript.browser`

Since users of the pure `sentry.java` SDK will have that as an SDK name, to help us easily differentiate what was installed in the app, for android @HazAT proposed `sentry.java.android`.

If it was rewritten in kotlin one day, it'd change to `sentry.kotlin.android` :)

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -85,12 +85,12 @@ private void setAppExtras(SentryEvent event) {
   private SdkVersion getSdkVersion() {
     SdkVersion sdkVersion = new SdkVersion();
 
-    sdkVersion.setName("sentry-android");
+    sdkVersion.setName("sentry.java.android");
```
marandaneto: so do we change here as well? now it's inconsistent, as it's supposed to be the maven packages if I got it right

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
marandaneto: what's about the version? should we also use `BuildConfig.VERSION_NAME`?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
bruno-garcia: This is the SDK name, that'll be reported at the bottom of the event page.
Maven package goes on `packages`

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
bruno-garcia: Yes, I'll merge this before we discuss that thouhg

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -85,12 +85,12 @@ private void setAppExtras(SentryEvent event) {
...
     String version = BuildConfig.VERSION_NAME;
     sdkVersion.setVersion(version);
     sdkVersion.addPackage("sentry-core", version);
```
bruno-garcia: Here are packages

PR #49, opened 2019-10-22
fix sentry UTC timestamp

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
make timestamp in UTC

timestamp was not in UTC

## :pencil: Checklist

- [ ] All tests passing

### DIFF
--- a/sentry-core/src/test/java/io/sentry/core/SentryEventTest.kt
+++ b/sentry-core/src/test/java/io/sentry/core/SentryEventTest.kt
@@ -7,7 +7,6 @@ import java.time.format.DateTimeFormatter
 import java.time.temporal.ChronoUnit
 import java.util.Date
 import java.util.Locale
-import kotlin.test.Ignore
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertNotEquals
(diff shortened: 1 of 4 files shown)


sentry-core/src/test/java/io/sentry/core/SentryEventTest.kt
```diff
@@ -22,6 +23,7 @@ class SentryEventTest {
             assertTrue(Instant.now().plus(1, ChronoUnit.HOURS).isAfter(SentryEvent().timestamp.toInstant()))
 
     @Test
+    @Ignore("fix me")
```
marandaneto: @bruno-garcia what have you tried to test here? just to fix it.

sentry-sample/src/main/AndroidManifest.xml
```diff
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     package="io.sentry.sample">
-
+  <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
bruno-garcia: Should this be set in `android-core` instead?
Since we'll be checking the network status before sending event?

bruno-garcia: we can also do that once we implement the check btw*

marandaneto: I'd make the user to opt-in for that, if the user adds the permission, all good.
the check is already done https://github.com/getsentry/sentry-android/blob/master/sentry-android-core/src/main/java/io/sentry/android/core/util/Permissions.java

sentry-core/src/test/java/io/sentry/core/SentryEventTest.kt
marandaneto: fixed it, `.toInstant()` was applying UTC again (-2h), but Date was UTC already

PR #51, opened 2019-10-22
some android lint ref.

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
a few warnings have been fixed.

warnings were annoying.

## :pencil: Checklist

- [X] All tests passing


sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -50,7 +49,7 @@ public boolean isEnabled() {
   public SentryId captureEvent(SentryEvent event) {
     SentryId sentryId;
     StackItem item = stack.peek();
-    sentryId = item.client.captureEvent(event, item.scope);
+    sentryId = item != null ? item.client.captureEvent(event, item.scope) : SentryId.EMPTY_ID;
```
bruno-garcia: `assert item != null` would or something that works on android would be better.
in other words, if item is null, we have a bug
We could do a `log(SentryLevel.FATAL)` in this case

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -59,7 +58,7 @@ public SentryId captureEvent(SentryEvent event) {
   public SentryId captureMessage(String message) {
     SentryId sentryId;
     StackItem item = stack.peek();
-    sentryId = item.client.captureMessage(message, item.scope);
+    sentryId = item != null ? item.client.captureMessage(message, item.scope) : SentryId.EMPTY_ID;
```
bruno-garcia: same

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -68,7 +67,8 @@ public SentryId captureMessage(String message) {
   public SentryId captureException(Throwable throwable) {
     SentryId sentryId;
     StackItem item = stack.peek();
-    sentryId = item.client.captureException(throwable, item.scope);
+    sentryId =
+        item != null ? item.client.captureException(throwable, item.scope) : SentryId.EMPTY_ID;
```
bruno-garcia: and here

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -77,14 +77,18 @@ public SentryId captureException(Throwable throwable) {
   public void close() {
     // Close the top-most client
     StackItem item = stack.peek();
-    item.client.close();
+    if (item != null) {
```
bruno-garcia: Makes me wonder are we supposed to do when the caller does `bindClient(null)`.
Should we bind null? I'd say we bind `NoOpClient.Instance` and make sure this is never null

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -77,14 +77,18 @@ public SentryId captureException(Throwable throwable) {
...
+    }
     isEnabled = false;
   }
 
   @Override
   public void addBreadcrumb(Breadcrumb breadcrumb) {
     StackItem item = stack.peek();
-    item.scope.addBreadcrumb(breadcrumb);
+    if (item != null) {
+      item.scope.addBreadcrumb(breadcrumb);
```
bruno-garcia: Same here. Item should never be null, neither scope (which isn't checked right now)

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -77,7 +77,11 @@ public void setExtra(String key, String value) {
 
   @Override
   protected Scope clone() {
-    // TODO: clone me
-    return new Scope();
+    Scope d = null;
+    try {
+      d = (Scope) super.clone();
+    } catch (CloneNotSupportedException ignored) {
+    } // Won't happen
```
bruno-garcia: Can we Log `ERROR` here? I love when I see _can't happen_ in the logs 😄

marandaneto: in this case, I'd throw `CloneNotSupportedException` and let `Hub` log it then.
I just did the same as you did before when calling `timestamp.clone()` it returns null if an error has occurred.

sentry-core/src/main/java/io/sentry/core/Hub.java
marandaneto: the NPE check is for the `stack` and not for the `client` itself, but agreed with `NoOpClient.Instance`

sentry-core/src/main/java/io/sentry/core/Hub.java
marandaneto: `scope` can't be null, it's not a parameter but created by us on the constructor

sentry-core/src/main/java/io/sentry/core/Hub.java
marandaneto: I'd not use `assert` on public methods, I'd log the `SentryEvent` if `StackItem` is null though, does it sound good?
same for the other cases.

bruno-garcia: `assert` doesn't get compiled in `release` builds so wouldn't matter. But Android complains about using `assert` anyway.

If StackItem is null the SDK has a huge bug because this class guards all access to it and it cannot let it remove the root scope. Please review the code and see if you spot a way where that could happen.

Logging `FATAL` StackItem is null would be ideal so we can get it fixed. If the SDK is buggy, logging SentryEvent doesn't add much value. These logs are to diagnose problems with the SDK and not let the user aware of errors we captured.

sentry-core/src/main/java/io/sentry/core/Hub.java
bruno-garcia: Right, I meant we're missing one for `client` here.

sentry-core/src/main/java/io/sentry/core/Scope.java
bruno-garcia: Agreed. We have to capture all exception on the `Hub` and log them out. We don't bubble any runtime errors to the users (our SDK **can't** crash apps). We can crash during initialization but not during execution.

sentry-core/src/main/java/io/sentry/core/Hub.java
marandaneto: `client` can't be null either, its created in the constructor or via `bindClient` but then we have `NoOp`

PR #52, opened 2019-10-22
feat: NDK

bruno-garcia:
We need to solve a few things before merging this:

1. Builds go from 2-5 minutes to 10-20. 
Should we postpone merging while the main SDK is heavy under development and just bring master to this branch until then?
2. CI builds from API > 14. 
Introduce a build with/without NDK to run the lower API levels?


.gitignore
```diff
@@ -6,3 +6,6 @@ out/
 local.properties
 **.iml
 *.hprof
+.cxx
+sentry-native-local/
```
tonyo:
```suggestion
```

README.md
```diff
@@ -11,3 +11,14 @@ Android SDK for Sentry
...
+# Development
+
+This repository includes [`sentry-native`](https://github.com/getsentry/sentry-native/) as a git submodule.
+To build against `sentry-native` checked-out elsewhere in your file system, create a symlink to `sentry-android-ndk/sentry-native-local`.
+For example, if you had `sentry-native` checked-out in a sibling directory to this repo:
+
+`ln -s ../sentry-native sentry-android-ndk/sentry-native-local`
+
+which will be picked up by `gradle` and used instead of the git submodule.
+This directory is algo included in `.gitignore` not to be shown as pending changes.
```
tonyo:
```suggestion
This directory is also included in `.gitignore` not to be shown as pending changes.
```

build.gradle.kts
```diff
@@ -42,9 +42,9 @@ allprojects {
                 TestLogEvent.FAILED)
             dependsOn("cleanTest")
         }
-        withType<JavaCompile> {
-            options.compilerArgs.addAll(arrayOf("-Xlint:all", "-Werror", "-Xlint:-classfile"))
-        }
+//        withType<JavaCompile> {
```
marandaneto: why did we remove that? we need for linting

buildSrc/src/main/java/Config.kt
```diff
@@ -11,7 +11,7 @@ object Config {
         private val sdkVersion = 29
 
         val buildToolsVersion = "29.0.2"
-        val minSdkVersion = 14
+        val minSdkVersion = 21
```
marandaneto: we are losing linter here, this should be done on specific ndk build file

tonyo:
@bruno-garcia Added a check, now we should only building NDK for a specific platform set via Travis environment variable `ABI` (looks like it's currently not used for anything else).
But just to keep in mind, when performing a release build, we should build all of them.

build.gradle.kts
bruno-garcia: we removed to get it to build

bruno-garcia: in other words: We need to fix this before merging

buildSrc/src/main/java/Config.kt
bruno-garcia: > Manifest merger failed : uses-sdk:minSdkVersion 14 cannot be smaller than version 21 declared in library [:sentry-android-ndk] /Users/bruno/git/sentry-android/sentry-android-ndk/build/intermediates/library_manifest/debug/AndroidManifest.xml as the library might be using APIs not available in 14
        Suggestion: use a compatible library with a minSdk of at most 14,
                or increase this project's minSdk version to at least 21,
                or use tools:overrideLibrary="io.sentry.android.ndk" to force usage (may lead to runtime failures)

PR #53, opened 2019-10-23
feat: Uncaught exception is captured

bruno-garcia:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

SentryOptions exposes `Integrations`.
A default integration is added to capture uncaught exceptions.

Basic sentry SDK feature

Unit tests

## :pencil: Checklist

- [ ] All tests passing

## :crystal_ball: Next steps
We could attemp unregisters but from what I recall when _unified API_ started @mitsuhiko was strongly against it.
If another integration also called `Thread.setDefaultUncaughtExceptionHandler` after we did, it wouldn't be possible to unregister, so that was the reason not to _pretend_ that we can unregister, even if best effort, and simply say: Once enabled, it's there for the lifetime of the app.

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandler.java
+++ b/sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandler.java
@@ -12,7 +12,7 @@ interface UncaughtExceptionHandler {
       return Adapter.INSTANCE;
     }
 
-    final static Adapter INSTANCE = new Adapter();
+    static final Adapter INSTANCE = new Adapter();
 
     private Adapter() {}
 
--- a/sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandler.java
+++ b/sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandler.java
@@ -1,0 +1,29 @@
+package io.sentry.core;
+
+/** An adapter to make UncaughtExceptionHandler testable */
+interface UncaughtExceptionHandler {
+  Thread.UncaughtExceptionHandler getDefaultUncaughtExceptionHandler();
+
+  void setDefaultUncaughtExceptionHandler(Thread.UncaughtExceptionHandler handler);
+
+  class Adapter implements UncaughtExceptionHandler {
+
+    UncaughtExceptionHandler getInstance() {
+      return Adapter.INSTANCE;
+    }
+
+    static Adapter INSTANCE = new Adapter();
+
+    private Adapter() {}
+
+    @Override
+    public Thread.UncaughtExceptionHandler getDefaultUncaughtExceptionHandler() {
+      return Thread.getDefaultUncaughtExceptionHandler();
+    }
+
+    @Override
+    public void setDefaultUncaughtExceptionHandler(Thread.UncaughtExceptionHandler handler) {
+      Thread.setDefaultUncaughtExceptionHandler(handler);
+    }
+  }
+}
--- a/sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
+++ b/sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
@@ -1,8 +1,8 @@
 package io.sentry.core;
 
-import io.sentry.core.util.Objects;
-
 import static io.sentry.core.ILogger.log;
+
+import io.sentry.core.util.Objects;
 
 /**
  * Sends any uncaught exception to Sentry, then passes the exception on to the pre-existing uncaught
--- a/sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
+++ b/sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
@@ -1,0 +1,73 @@
+package io.sentry.core;
+
+import static io.sentry.core.ILogger.log;
+
+/**
+ * Sends any uncaught exception to Sentry, then passes the exception on to the pre-existing uncaught
+ * exception handler.
+ */
+public class UncaughtExceptionHandlerIntegration
+    implements Integration, Thread.UncaughtExceptionHandler {
+  /** Reference to the pre-existing uncaught exception handler. */
+  private Thread.UncaughtExceptionHandler defaultExceptionHandler;
+
+  private IHub hub;
+  private SentryOptions options;
+
+  private boolean isRegistered = false;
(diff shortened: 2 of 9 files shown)


sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandler.java
```diff
@@ -0,0 +1,29 @@
...
+
+  void setDefaultUncaughtExceptionHandler(Thread.UncaughtExceptionHandler handler);
+
+  class Adapter implements UncaughtExceptionHandler {
+
+    UncaughtExceptionHandler getInstance() {
+      return Adapter.INSTANCE;
+    }
+
+    static Adapter INSTANCE = new Adapter();
```
marandaneto:
```suggestion
    final static Adapter INSTANCE = new Adapter();
```

sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private boolean isRegistered = false;
+  private UncaughtExceptionHandler threadAdapter;
+
+  public UncaughtExceptionHandlerIntegration() {
+    this(UncaughtExceptionHandler.Adapter.INSTANCE);
+  }
+
+  UncaughtExceptionHandlerIntegration(UncaughtExceptionHandler threadAdapter) {
+    if (threadAdapter == null) {
```
marandaneto: you may use our Util class `Objects.requireNonNull`

sentry-core/src/test/java/io/sentry/core/UncaughtExceptionHandlerIntegrationTest.kt
```diff
@@ -0,0 +1,47 @@
...
+        val sut = UncaughtExceptionHandlerIntegration(handlerMock);
+        sut.register(hubMock, options)
+        sut.uncaughtException(threadMock, throwableMock)
+        verify(hubMock).captureException(throwableMock)
+    }
+
+    @Test
+    fun `when register is called, current handler is not lost`()
+    {
+        val handlerMock = mock<UncaughtExceptionHandler>()
```
ninniuz: Can't you test this by setting a mock `Thread.UncaughtExceptionHandler` with `Thread.setDefaultUncaughtExceptionHandler` before running `sut.register` and verifying that `uncaughtException` is called on the mock?

sentry-core/src/test/java/io/sentry/core/UncaughtExceptionHandlerIntegrationTest.kt
```diff
@@ -0,0 +1,47 @@
...
+import com.nhaarman.mockitokotlin2.whenever
+import kotlin.test.Test
+
+class UncaughtExceptionHandlerIntegrationTest {
+    @Test
+    fun `when UncaughtExceptionHandlerIntegration is initialized, uncaught handler is unchanged`()
+    {
+        val handlerMock = mock<UncaughtExceptionHandler>()
+        UncaughtExceptionHandlerIntegration(handlerMock);
+        verifyZeroInteractions(handlerMock)
```
ninniuz: Can't you test this by setting a mock `Thread.UncaughtExceptionHandler` with `Thread.setDefaultUncaughtExceptionHandler` before creating an instance of `UncaughtExceptionHandlerIntegration ` and verifying that `Thread.getDefaultUncaughtExceptionHandler()` instance equals your mock (actually I would use `assertSame` here)?

sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandler.java
```diff
@@ -0,0 +1,29 @@
+package io.sentry.core;
+
+/** An adapter to make UncaughtExceptionHandler testable */
```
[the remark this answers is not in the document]
bruno-garcia: The issue is that running the tests against `Thread.set...` would modify the handler process-wide`. If tests run in parallel they would race and be brittle. I needed to hide the static stuff behind some interface for mocking.

sentry-core/src/test/java/io/sentry/core/UncaughtExceptionHandlerIntegrationTest.kt
bruno-garcia: Mocking static fields is flaky. it'll work until another test run in parallel also modifying `Thread.setDefaultUncaughtExceptionHandler`

sentry-core/src/test/java/io/sentry/core/UncaughtExceptionHandlerIntegrationTest.kt
bruno-garcia: Won't work when tests run in parallel. We shouldn't rely on any static state for unit tests

sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandler.java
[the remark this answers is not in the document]
ninniuz: As far as I know Gradle `test` task only supports parallelism at the class level.
Moreover, you could add `@Before` and `@After` methods to store/restore any previous default uncaught exception handler to not pollute other test methods.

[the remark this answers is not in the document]
bruno-garcia: That's fair, doesn't change the fact that relying on static state is a problem. We might depend on other code (like a test library) that wraps that later and things start breaking once in 10 builds. Becomes super hard to track down. I leave code that relies on `static` state untested. I wrap (like in this case) when I want to test the code that interacts with it.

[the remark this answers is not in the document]
ninniuz: agreed

sentry-core/src/test/java/io/sentry/core/UncaughtExceptionHandlerIntegrationTest.kt
ninniuz: agreed

sentry-core/src/test/java/io/sentry/core/UncaughtExceptionHandlerIntegrationTest.kt
ninniuz: agreed

PR #59, opened 2019-10-23
ref: build release only

bruno-garcia:


bruno-garcia:
Closing this for now since it doesn't make things much faster and comes with caveats like `lint` and `spotless`

PR #60, opened 2019-10-23
init. ndk only if ndk opt on manifest is enabled

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
Init. NDK only if sdk opt is enabled.

Right now is trying to load every time.

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
Write tests

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -1,6 +1,7 @@
 package io.sentry.android.core;
 
 import android.content.Context;
+import android.os.Build;
 import io.sentry.core.ILogger;
 import io.sentry.core.SentryLevel;
 import io.sentry.core.SentryOptions;
@@ -25,7 +26,7 @@ class AndroidOptionsInitializer {
     options.addEventProcessor(new DefaultAndroidEventProcessor(context, options));
     options.setSerializer(new AndroidSerializer(options.getLogger()));
 
-    if (options.isEnableNdk()) {
+    if (options.isEnableNdk() && isNdkAvailable()) {
       try {
         // TODO: Create Integrations interface and use that to initialize NDK
         Class<?> cls = Class.forName("io.sentry.android.ndk.SentryNdk");
@@ -49,5 +50,9 @@ class AndroidOptionsInitializer {
       envelopesDir.mkdirs();
     }
     options.setCacheDirPath(envelopesDir.getAbsolutePath());
+  }
+
+  private static boolean isNdkAvailable() {
+    return Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP;
   }
 }
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -35,9 +35,11 @@ class AndroidOptionsInitializer {
         Object[] args = new Object[1];
         args[0] = options;
         method.invoke(null, args);
-      } catch (ClassNotFoundException exc) {
-        options.getLogger().log(SentryLevel.ERROR, "Failed to load SentryNdk.");
+      } catch (ClassNotFoundException e) {
+        options.setEnableNdk(false);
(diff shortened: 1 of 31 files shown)


sentry-android-ndk/src/main/AndroidManifest.xml
```diff
@@ -1,2 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest package="io.sentry.android.ndk" />
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          package="io.sentry.android.ndk">
+  <application>
+    <meta-data android:name="io.sentry.ndk" android:value="true"/>
```
bruno-garcia: When the manifest merging happens, I assume what's set in the apps manifest will win? We wouldn't want to force init even when the user is trying to turn it off (even if the ndk package is installed)

marandaneto: @bruno-garcia yes, there are merging rules, App's manifest is a high priority, but needs to set the merging rule, unfortunately, added an example and let's discuss it.
another way would be via `buildConfigField` which is gonna be available on generated BuildConfig class.
https://developer.android.com/studio/build/gradle-tips

sentry-sample/src/main/AndroidManifest.xml
```diff
@@ -22,6 +22,10 @@
 
     <meta-data android:name="io.sentry.dsn" android:value="https://key@sentry.io/123" />
     <meta-data android:name="io.sentry.debug" android:value="true" />
+
 <!--    <meta-data android:name="io.sentry.auto-init" android:value="false" /> // how to disable the auto-init-->
+
+<!--    <meta-data android:name="io.sentry.ndk" android:value="false" /> // how to disable the NDK-->
+<!--               tools:node="remove" />-->
```
marandaneto: @bruno-garcia here is what it needs to be done to disable NDK on the client's side.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -56,4 +51,8 @@ private static void createsEnvelopeDirPath(SentryOptions options, Context contex
     }
     options.setCacheDirPath(envelopesDir.getAbsolutePath());
   }
+
+  private static boolean isNdkAvailable() {
```
ninniuz: Are we settling on the fact that NDK support will not be available on API < 21?

ninniuz: Yeah, tried to build the NDK feature branch on API 14 and got an error on that method :(

sentry-android-ndk/src/main/AndroidManifest.xml
bruno-garcia: Wouldn't it be simpler to just not add this configuration here? Leave it `true` in the code instead?

sentry-sample/src/main/AndroidManifest.xml
bruno-garcia: Does it require the `tools:node="remove" `?
If so, I'm afraid this could become a source of support tickets.

marandaneto: yes, if you try to compile it will tell you to do that as a library already did set it up, it's a common problem so developers might know how to solve by eye, but I'd keep this in the documentation.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
marandaneto: we might have a workaround, parsing `/proc/PID/maps` by hand and getting what we need.
but still thinking if it's a good idea.

marandaneto:
@bruno-garcia the configuration will be to avoid a reflection call if we already know NDK is not there or disabled.
Reflection on Android is expensive.

sentry-android-ndk/src/main/AndroidManifest.xml
ninniuz: I second @bruno-garcia comments. Just leave it enabled in the code and let apps use a `meta-data` to turn it off.

marandaneto: or that, done :D

PR #68, opened 2019-10-25
apply scope to event

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
copying scope data to the event

scope data was not being coped to the event

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Scope.java
+++ b/sentry-core/src/main/java/io/sentry/core/Scope.java
@@ -48,10 +48,6 @@ public class Scope implements Cloneable {
     this.fingerprint = fingerprint;
   }
 
-  public void addFingerprint(String fingerprint) {
-    this.fingerprint.add(fingerprint);
-  }
-
   public List<Breadcrumb> getBreadcrumbs() {
     return breadcrumbs;
   }
--- a/sentry-core/src/main/java/io/sentry/core/Scope.java
+++ b/sentry-core/src/main/java/io/sentry/core/Scope.java
@@ -48,6 +48,10 @@ public class Scope implements Cloneable {
     this.fingerprint = fingerprint;
   }
 
+  public void addFingerprint(String fingerprint) {
+    this.fingerprint.add(fingerprint);
+  }
+
   public List<Breadcrumb> getBreadcrumbs() {
     return breadcrumbs;
   }
--- a/sentry-core/src/main/java/io/sentry/core/SentryClient.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryClient.java
@@ -35,6 +35,11 @@ public class SentryClient implements ISentryClient {
   public SentryId captureEvent(SentryEvent event, @Nullable Scope scope) {
     log(options.getLogger(), SentryLevel.DEBUG, "Capturing event: %s", event.getEventId());
 
+    // TODO: This will go into the MainEventProcessor
+    if (event.getThreads() == null) {
+      SentryThreadFactory sentryThreadFactory = new SentryThreadFactory();
+      event.setThreads(sentryThreadFactory.getCurrentThreads());
+    }
     // TODO: To be done in the main processor
     if (event.getRelease() == null) {
       event.setRelease(options.getRelease());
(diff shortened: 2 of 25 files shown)


sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -48,6 +48,15 @@ public SentryId captureEvent(SentryEvent event, @Nullable Scope scope) {
       event.setEnvironment(options.getEnvironment());
     }
 
+    if (scope != null) {
```
marandaneto: * scope level has priority over event level
* move block to MainProcessor
* List should be null by default and check that before setting values

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -48,6 +48,30 @@ public SentryId captureEvent(SentryEvent event, @Nullable Scope scope) {
...
+      if (event.getTransaction() == null) {
+        event.setTransaction(scope.getTransaction());
+      }
+      if (event.getUser() == null) {
+        event.setUser(scope.getUser());
+      }
+      if (event.getFingerprint() == null) {
+        event.setFingerprint(scope.getFingerprint());
+      }
+      if (event.getBreadcrumbs() == null) {
```
bruno-garcia: The ones above look good but for Tags, Extra and Breadcrumbs I think it should look more like this:

```java
if (event.getBreadcrumbs() == null) {
  // Copy breadcrumbs (which are immutable) into a new list
  event.setBreadcrumbs(new ArrayList<>(scope.getBreadcrumbs()));
} else {
  // append items to event breadcrumb list
  List<Breadcrumb> breadcrumbs = event.getBreadcrumbs();
  for (Breadcrumb scopeBreadcrumbs: scope.getBreadcrumbs()) {
    breadcrumbs.add(scopeBreadcrumbs);
  }
}
```

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -48,6 +48,30 @@ public SentryId captureEvent(SentryEvent event, @Nullable Scope scope) {
...
+        event.setBreadcrumbs(scope.getBreadcrumbs());
+      }
+      if (event.getTags() == null) {
+        event.setTags(scope.getTags());
+      }
+      if (event.getExtra() == null) {
+        event.setExtra(scope.getExtra());
+      }
+      if (scope.getLevel() != null) {
```
```suggestion
      // Level from scope exceptionally take precedence over the event
      if (scope.getLevel() != null) {
```

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
```diff
@@ -203,30 +203,58 @@ public void setFingerprint(List<String> fingerprint) {
...
   public Map<String, String> getTags() {
     return tags;
   }
 
-  public void setTags(HashMap<String, String> tags) {
+  public void setTags(Map<String, String> tags) {
     this.tags = tags;
   }
 
+  public void addTag(String key, String value) {
```
bruno-garcia: Unified API calls it `setTag` since it's a dictionaru and will override an existent tag with the same key

```suggestion
  public void setTag(String key, String value) {
```

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
```diff
@@ -203,30 +203,58 @@ public void setFingerprint(List<String> fingerprint) {
...
   public Map<String, Object> getExtra() {
     return extra;
   }
 
-  public void setExtra(HashMap<String, Object> extra) {
+  public void setExtra(Map<String, Object> extra) {
     this.extra = extra;
   }
 
+  public void addExtra(String key, Object value) {
```
bruno-garcia:
```suggestion
  public void setExtra(String key, Object value) {
```

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -48,6 +48,10 @@ public void setFingerprint(List<String> fingerprint) {
     this.fingerprint = fingerprint;
   }
 
+  public void addFingerprint(String fingerprint) {
```
bruno-garcia: the API is `setFingerprint` and it takes the whole `List<String>`. AFAIK we're not offering an API to only append to it. When one sets the fingerprint, it;s the whole list.

PR #69, opened 2019-10-26
added Android friendly init.

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
added friendly Android init.

if auto-init is disabled, it's way more code to init. the SDK

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
@@ -1,0 +1,20 @@
+package io.sentry.android.core;
+
+import android.content.Context;
+import io.sentry.core.Sentry;
+import io.sentry.core.util.NonNull;
+
+/** Sentry initialization class */
+public final class SentryAndroid {
+
+  private SentryAndroid() {}
+
+  /**
+   * Sentry initialization method if auto-init is disabled
+   *
+   * @param context Application. context
+   */
+  public static void init(@NonNull final Context context) {
+    Sentry.init(options -> AndroidOptionsInitializer.init(options, context));
+  }
+}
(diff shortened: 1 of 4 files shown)


sentry-sample/src/main/java/io/sentry/sample/MyApplication.java
```diff
@@ -0,0 +1,17 @@
...
+
+/** Apps. main Application. */
+public class MyApplication extends Application {
+
+  @Override
+  public void onCreate() {
+    super.onCreate();
+
+    // how to init. Sentry manually
+    // SentryAndroid.init(this);
```
bruno-garcia: The use-case of initializing manually is if users want to set stuff into `SentryOptions` like `BeforeSend` so if we go this route of having `SentryAndroid` static class with `init` at least we need a nice API like:

```java
SentryAndroid.init(o -> {
o.beforeSend = e -> { ... },
o.AndroidSepcificStuf = ...
});

```

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
```diff
@@ -0,0 +1,20 @@
+package io.sentry.android.core;
+
+import android.content.Context;
+import io.sentry.core.Sentry;
+import io.sentry.core.util.NonNull;
+
+/** Sentry initialization class */
+public final class SentryAndroid {
```
bruno-garcia: IIRC @mitsuhiko had some thoughts WRT introducing some second `init` like this, maybe reagarding docs or whatnot.
I'd just like to get his review here before we move forward.
/cc @HazAT

sentry-sample/src/main/java/io/sentry/sample/MyApplication.java
marandaneto: that's the idea, the one I've added is only if the user wants to init manually for some reason but still supporting our current set up with a minimal amount of code, of course this API could be extended.

PR #70, opened 2019-10-28
WIP - publish maven local

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
publish artifacts to maven local

for testing the SDK as a maven dependency

App. sample implements the maven artifact.

How to publish:
./gradlew publishReleasePublicationToMavenLocal

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
add javadocs to maven as well


buildSrc/src/main/java/Config.kt
```diff
@@ -3,7 +3,7 @@ object Config {
     val kotlinStdLib = "stdlib-jdk8"
 
     object BuildPlugins {
-        val androidGradle = "com.android.tools.build:gradle:3.5.1"
+        val androidGradle = "com.android.tools.build:gradle:3.6.0-beta01"
```
marandaneto: @bruno-garcia AGP supports maven publish ootb now on its new beta version, so no plugins dependency as before.

.travis.yml
```diff
@@ -11,6 +11,13 @@ licenses:
   - android-sdk-preview-license-.+
   - android-sdk-license-.+
   - google-gdk-license-.+
+install:
```
marandaneto: this is just a try because 1st build also failed, with missing NDK error

sentry-sample/build.gradle.kts
```diff
@@ -58,7 +58,8 @@ android {
 dependencies {
     implementation(fileTree(mapOf("dir" to "libs", "include" to listOf("*.jar"))))
 
-    implementation(project(":sentry-android"))
+//    implementation(project(":sentry-android"))
+    implementation("io.sentry:sentry-android:2.0.0-SNAPSHOT")
```
bruno-garcia: Does this mean we need to run `release` task before updating the sample?
Or can we just change code and hit 'play' on IntelliJ for example?

buildSrc/src/main/java/Config.kt
bruno-garcia: nice

sentry-sample/build.gradle.kts
marandaneto: @bruno-garcia I've inverted the comment, my bad.

marandaneto:
> LGTM but CI isn't happy yet

NDK not configured. Download it with SDK manager. Preferred NDK version is '20.0.5594570'

maybe because it's still beta?!

gradle.properties
```diff
@@ -1,8 +1,15 @@
...
 
 # Parallel execution
 org.gradle.parallel=true
 
 # Cacheable unit tests
 android.testConfig.useRelativePath = true
+
+# Migrating to AndroidX - AGP 3.6.x
+android.useAndroidX=true
+#android.enableJetifier=true - causing SHA1 digest error
```
marandaneto: similar issue https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/455

buildSrc/src/main/java/Config.kt
```diff
@@ -3,7 +3,7 @@ object Config {
     val kotlinStdLib = "stdlib-jdk8"
 
     object BuildPlugins {
-        val androidGradle = "com.android.tools.build:gradle:3.5.2"
+        val androidGradle = "com.android.tools.build:gradle:3.6.0-beta03"
```
marandaneto: @ninniuz what do you think about that?
I'd like to use `maven-publish` with the native support that AGP 3.6.x provides.

ninniuz: It is a great addition to AGP and it makes publishing a lot easier, but I wouldn't use beta releases as they could affect aar generation.

sentry-android/build.gradle.kts
```diff
@@ -30,3 +32,13 @@ dependencies {
     api(project(":sentry-android-core"))
     api(project(":sentry-android-ndk"))
 }
+
+afterEvaluate {
```
ninniuz: Can you factor out the repeated publishing logic into the main `build.gradle` file?

buildSrc/src/main/java/Config.kt
marandaneto: @ninniuz based on the changelog, the "only" different thing that affects aar generation would be `useNewApkCreator` which I've disabled, that's why I think it's safe to go, but I do understand your concern.

marandaneto:
for now in favor of https://github.com/getsentry/sentry-android/pull/129

marandaneto:
using https://github.com/novoda/bintray-release for now

PR #71, opened 2019-10-28
added missing protocol fields

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
added the missing fields of the protocol.

there were a few missing fields

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Scope.java
+++ b/sentry-core/src/main/java/io/sentry/core/Scope.java
@@ -40,11 +40,11 @@ public class Scope implements Cloneable {
     this.user = user;
   }
 
-  public List<String> getFingerprint() {
+  public List<String> getFingerprints() {
     return fingerprint;
   }
 
-  public void setFingerprint(List<String> fingerprint) {
+  public void setFingerprints(List<String> fingerprint) {
     this.fingerprint = fingerprint;
   }
 
@@ -67,7 +67,7 @@ public class Scope implements Cloneable {
     this.tags.put(key, value);
   }
 
-  public Map<String, Object> getExtra() {
+  public Map<String, Object> getExtras() {
     return extra;
   }
 
--- a/sentry-core/src/main/java/io/sentry/core/Scope.java
+++ b/sentry-core/src/main/java/io/sentry/core/Scope.java
@@ -40,11 +40,11 @@ public class Scope implements Cloneable {
     this.user = user;
   }
 
-  public List<String> getFingerprints() {
+  public List<String> getFingerprint() {
     return fingerprint;
   }
 
-  public void setFingerprints(List<String> fingerprint) {
+  public void setFingerprint(List<String> fingerprint) {
     this.fingerprint = fingerprint;
   }
 
--- a/sentry-core/src/main/java/io/sentry/core/SentryEvent.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryEvent.java
@@ -33,7 +33,6 @@ public class SentryEvent implements IUnknownPropertiesConsumer {
   private Map<String, Object> extra;
   private Map<String, Object> unknown;
   private Map<String, String> modules;
-  // TODO: missing Debug Meta Interface?
 
   SentryEvent(SentryId eventId, Date timestamp) {
     this.eventId = eventId;
--- a/sentry-core/src/main/java/io/sentry/core/SentryEvent.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryEvent.java
@@ -32,6 +32,8 @@ public class SentryEvent implements IUnknownPropertiesConsumer {
   private Map<String, String> tags;
   private Map<String, Object> extra;
   private Map<String, Object> unknown;
+  private Map<String, String> modules;
+  // TODO: missing Debug Meta Interface?
 
   SentryEvent(SentryId eventId, Date timestamp) {
     this.eventId = eventId;
@@ -119,11 +121,11 @@ public class SentryEvent implements IUnknownPropertiesConsumer {
     this.threads = new SentryValues<>(threads);
   }
 
-  public List<SentryException> getException() {
+  public List<SentryException> getExceptions() {
     return exception.getValues();
   }
 
-  public void setException(List<SentryException> exception) {
+  public void setExceptions(List<SentryException> exception) {
     this.exception = new SentryValues<>(exception);
   }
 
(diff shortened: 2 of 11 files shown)


sentry-core/src/main/java/io/sentry/core/SentryEvent.java
```diff
@@ -32,6 +32,8 @@
   private Map<String, String> tags;
   private Map<String, Object> extra;
   private Map<String, Object> unknown;
+  private Map<String, String> modules;
```
marandaneto: @bruno-garcia is it a `Map` or should we reuse the clas `Package`?

sentry-core/src/main/java/io/sentry/core/protocol/SdkVersion.java
```diff
@@ -3,12 +3,14 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 public class SdkVersion implements IUnknownPropertiesConsumer {
   private String name;
   private String version;
   private List<Package> packages = new CopyOnWriteArrayList<>();
+  private Map<String, String> integrations = new ConcurrentHashMap<>();
```
marandaneto: @bruno-garcia is it a `Map` or should we reuse the clas `Package`? does it need to be thread-safe?

sentry-core/src/main/java/io/sentry/core/protocol/SentryStackTrace.java
```diff
@@ -7,7 +7,7 @@
 /** The Sentry stacktrace. */
 public class SentryStackTrace implements IUnknownPropertiesConsumer {
   private List<SentryStackFrame> frames;
-  // TODO registers is missing?
+  private Map<String, String> registers;
```
marandaneto: @bruno-garcia is it a `Map`?

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
```diff
@@ -32,6 +32,8 @@
...
+  // TODO: missing Debug Meta Interface?
```
marandaneto: @bruno-garcia do we need that?

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -40,11 +40,11 @@ public void setUser(User user) {
     this.user = user;
   }
 
-  public List<String> getFingerprint() {
+  public List<String> getFingerprints() {
```
bruno-garcia: Sorry it's confusing but the API is `getFingerprint`:

```
scope.set_fingerprint(fingerprint[]): Sets the fingerprint to group specific events together
```

https://docs.sentry.io/development/sdk-dev/unified-api/#scope

A way to look at it is that "A fingerprint is a collection of strings`.

```suggestion
  public List<String> getFingerprint() {
```

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -40,11 +40,11 @@ public void setUser(User user) {
...
     return fingerprint;
   }
 
-  public void setFingerprint(List<String> fingerprint) {
+  public void setFingerprints(List<String> fingerprint) {
```
bruno-garcia:
```suggestion
  public void setFingerprint(List<String> fingerprint) {
```

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
bruno-garcia: Mainly the proguard id

sentry-core/src/main/java/io/sentry/core/Scope.java
marandaneto: fixed, the whole docs but the method is called `fingerprints` (plural).
https://docs.sentry.io/data-management/event-grouping/

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
marandaneto: ok for that we have its own task

bruno-garcia: right

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
bruno-garcia: Protocol says it's just a map of string name and string version but package at this point coincidentally matches that shape. I'd say let's merge this and discuss that offline

sentry-core/src/main/java/io/sentry/core/protocol/SdkVersion.java
bruno-garcia: Anything in the scope needs to be thread-safe.
integrations is a `List<String>` actually not a map, my bad.

sentry-core/src/main/java/io/sentry/core/protocol/SentryStackTrace.java
bruno-garcia: Looks like it: https://docs.sentry.io/development/sdk-dev/event-payloads/stacktrace/#attributes
```
 registers:

    Optional. A map of register names and their values. The values should contain the actual register values of the thread, thus mapping to the last frame in the list.
```

sentry-core/src/main/java/io/sentry/core/protocol/SdkVersion.java
```diff
@@ -45,4 +47,12 @@ public void addPackage(String name, String version) {
   public void acceptUnknownProperties(Map<String, Object> unknown) {
     this.unknown = unknown;
   }
+
+  public Map<String, String> getIntegrations() {
+    return integrations;
+  }
+
+  public void setIntegrations(Map<String, String> integrations) {
```
bruno-garcia: Since the `Scope` requires thread-safe collections, I'm not sure it's a good idea exposing this setter.
Perhaps we only expose: `setIntegration(Map.Key<String, String> integration);` and add that to the list ourselves.

sentry-core/src/main/java/io/sentry/core/protocol/SdkVersion.java
bruno-garcia: We're holding this off of the event not the scope so neither collection need to be synchronized

PR #74, opened 2019-10-29
fixing a bunch of things to send event correctly

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fixing a bunch of things to send event correctly

Because some data were missing or not possible to send them over.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
+++ b/sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
@@ -20,6 +20,7 @@ public class MainEventProcessor implements EventProcessor {
       event.setThreads(sentryThreadFactory.getCurrentThreads());
     }
 
+    // DefaultAndroidEventProcessor also sets it, which one is correct or which one should be 1st?
     if (event.getRelease() == null) {
       event.setRelease(options.getRelease());
     }
--- a/sentry-core/src/main/java/io/sentry/core/Sentry.java
+++ b/sentry-core/src/main/java/io/sentry/core/Sentry.java
@@ -74,35 +74,35 @@ public final class Sentry {
     return getCurrentHub().captureException(throwable);
   }
 
-  public void addBreadcrumb(Breadcrumb breadcrumb) {
+  public static void addBreadcrumb(Breadcrumb breadcrumb) {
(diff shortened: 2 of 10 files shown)


sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
```diff
@@ -20,6 +20,7 @@ public SentryEvent process(SentryEvent event) {
       event.setThreads(sentryThreadFactory.getCurrentThreads());
     }
 
+    // DefaultAndroidEventProcessor also sets it, which one is correct or which one should be 1st?
```
marandaneto: @bruno-garcia

sentry-core/src/main/java/io/sentry/core/Sentry.java
```diff
@@ -74,35 +74,35 @@ public static SentryId captureException(Throwable throwable) {
     return getCurrentHub().captureException(throwable);
   }
 
-  public void addBreadcrumb(Breadcrumb breadcrumb) {
+  public static void addBreadcrumb(Breadcrumb breadcrumb) {
```
marandaneto: @bruno-garcia please check if all of them need to be `static`, as it is doesn't make sense to me

sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
bruno-garcia: `DefaultAndroidEventProcessor` should be added second and run after this one.
But can't `sentry-android` just set the value to `SentryOptions` before passing it to `Sentry.init`?

sentry-core/src/main/java/io/sentry/core/Sentry.java
bruno-garcia: Yeah everything should be static.

sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
marandaneto: MainEventProcessor always run 1st so that's fine.

PR #75, opened 2019-10-29
implemented inApp prefix for stacktraces

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Implements inApp prefix for stack traces, so if the stack traces contain our packages, the noise is hidden and the user may focus on the real problem.

stack traces contain noise if our packages appear there.

unit testing

## :pencil: Checklist

- [X] All tests passing


sentry-core/src/main/java/io/sentry/core/SentryOptions.java
```diff
@@ -28,6 +28,7 @@
   private String environment;
   private Proxy proxy;
   private Double sampling;
+  private String inAppPrefix = "io.sentry";
```
bruno-garcia: `inApp` is a list so that the user can add  the package of his app and other libraries that he'd like to show up in Sentry. I believe we shouldn't show Sentry code in the stacktrace for users by default. They can always hit `show all` to see that.

sentry-core/src/main/java/io/sentry/core/SentryOptions.java
```diff
@@ -196,7 +205,12 @@ public void setSampling(Double sampling) {
   }
 
   public SentryOptions() {
-    eventProcessors.add(new MainEventProcessor(this));
+    SentryStackTraceFactory sentryStackTraceFactory = new SentryStackTraceFactory(this.inAppPrefix);
```
bruno-garcia: Can we bury this into `MainEventProcessor` on a ctor overload? One takes everything, the second passes new instances like this?
`SentryStackTraceFactory` would take either the `List<String>`s for InAppInclude and InAppExclude or the whole `Option`.

PR #79, opened 2019-10-30
added nopen and errorprone and fixed some issues

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
added nopen and errorprone and fixed some issues

Only lint was not enough not to help us to follow the best practices, so I've added nopen and errorprone as they need to work together.

## :pencil: Checklist

- [X] All tests passing


sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -49,7 +51,7 @@ private static void initializeCacheDirs(Context context, SentryOptions options)
     File cacheDir = new File(context.getCacheDir(), "sentry");
     cacheDir.mkdirs();
     options.setCacheDirPath(cacheDir.getAbsolutePath());
-    (new File(options.getOutboxPath())).mkdirs();
+    new File(options.getOutboxPath()).mkdirs();
```
bruno-garcia: This will do I/O on the calling thread which likely will be the main thread since it's `init`.
Let's try to postpone this directory creation and try to do this in a background thread

bruno-garcia: Yeah I see, we gotta fix this though I'll create an issue.

PR #81, opened 2019-10-30
added sentry client name and version dynamically

marandaneto:


sentry-android-core/build.gradle.kts
```diff
@@ -23,6 +23,8 @@ android {
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
 
         versionName = "$version"
+
+        buildConfigField("String", "SENTRY_CLIENT_NAME", "\"${Config.Sentry.SENTRY_CLIENT_NAME}\"")
```
marandaneto: the advantage of doing that is, we could set up different name/version by build tyke, like debug/release if we have a sentry server for different environments so we don't mix up our development and real tests

PR #82, opened 2019-10-30
feat: Auto detect InAppInclude

bruno-garcia:


sentry-core/src/main/java/io/sentry/core/SentryOptions.java
```diff
@@ -194,16 +194,22 @@ public void setSampling(Double sampling) {
     return inAppExcludes;
   }
 
-  public void setInAppExcludes(List<String> inAppExcludes) {
-    this.inAppExcludes = inAppExcludes;
+  public void addInAppExclude(String exclude) {
+    if (this.inAppExcludes == null) {
```
marandaneto: only use `this.` if they have the same name, otherwise is just noisy

sentry-core/src/main/java/io/sentry/core/SentryOptions.java
```diff
@@ -194,16 +194,22 @@ public void setSampling(Double sampling) {
...
   }
 
   public List<String> getInAppIncludes() {
     return inAppIncludes;
   }
 
-  public void setInAppIncludes(List<String> inAppIncludes) {
-    this.inAppIncludes = inAppIncludes;
+  public void addInAppInclude(String include) {
+    if (this.inAppIncludes == null) {
```
marandaneto: same as above

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -46,6 +47,15 @@ static void init(SentryOptions options, Context context, ILogger logger) {
     }
   }
 
+  private static void setDefaultInApp(Context context, SentryOptions options) {
+    // If context passed in is an application-defined class, we can use to white-list frames
+    String packageName = context.getClass().getPackage().getName();
```
marandaneto: why's that? why do we want to white-list the android-package?

sentry-android-core/src/test/java/io/sentry/android/core/FakeApplication.kt
```diff
@@ -0,0 +1,5 @@
+package io.sentry.android.core
+
+import android.app.Application
+
+class FakeApplication : Application()
```
marandaneto: we already have `ApplicationStub`, you may use the same

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
marandaneto: the user could hide the MainApplication whatever they want to, like `com.my.company.mythings.myusefulthings.idontknowwhat`.
not sure if it's a good idea, I'd prefer to keep opt-in by the user.
Or to use the package name like `packageInfo.packageName` which is supposed to be way more accurate

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -46,6 +47,15 @@ static void init(SentryOptions options, Context context, ILogger logger) {
     }
   }
 
+  private static void setDefaultInApp(Context context, SentryOptions options) {
+    // If context passed in is an application-defined class, we can use to white-list frames
```
bruno-garcia:
```suggestion
    // If context passed in is an application-defined class, we can use to mark frames as `InApp`
```

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
bruno-garcia: OptIn is the worst experience. We want to give sensible defaults.
If picking up from packageInfo.packageName is the best approach then lets do that!

PR #83, opened 2019-10-30
feat: Hub captures exception

bruno-garcia:
* Hub catches and logs exception in callbacks and calls to dependencies
* When the Hub is disabled, method calls are no-op


sentry-core/src/test/java/io/sentry/core/HubTest.kt
```diff
@@ -31,12 +38,42 @@ class HubTest {
...
-        assertEquals("transaction", scope.transaction)
-        assertEquals("fingerprint", scope.fingerprint[0])
+        assertEquals("transaction", clone.transaction)
+        assertEquals("fingerprint", clone.fingerprint.first())
         assertEquals("tags", clone.tags["tags"])
         assertEquals("a@a.com", clone.user.email)
     }
 
+    @Test
+    @Ignore("it's a shallow copy and we need a deep-copy") // TODO: https://www.baeldung.com/java-deep-copy
```
marandaneto: @bruno-garcia all the clones should be shallow or deep-copy? right now it returns a new instance of the object but it's fields are still the same.

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -52,165 +54,310 @@ public boolean isEnabled() {
...
-      if (stack.size() != 1) {
-        stack.pop();
+    if (!isEnabled()) {
+      logIfNotNull(
+          options.getLogger(),
+          SentryLevel.WARNING,
+          "Instance is disabled and this 'popScope' call is a no-op.");
+    } else {
+      // Don't drop the root scope
+      synchronized (stack) { // TODO: is it necessary? we should never sync a concurrent object
```
marandaneto: @bruno-garcia  do we need that? `stack` is already thread-safe

sentry-core/src/test/java/io/sentry/core/HubTest.kt
bruno-garcia: I want to say shallow is OK but for Scope data, if the value is not immutable, copying the reference would be a problem.

sentry-core/src/main/java/io/sentry/core/Hub.java
bruno-garcia: There's a race here:

```
        if (stack.size() != 1) {
          stack.pop();
```

sentry-core/src/test/java/io/sentry/core/HubTest.kt
marandaneto: @bruno-garcia ok so we already have a task on asana for it, default implementation of cloning on Java is shallow and it's also faulty if we make the fields `final`

PR #84, opened 2019-10-31
refactoring of application class and leakcanary

marandaneto:
…un release mode

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
refactored to use application class, added leak canary

now we have application class and calls should be done there.
added leakcanary so if we have memory leaks we may catch them

## :pencil: Checklist

- [X] All tests passing


marandaneto:
@bruno-garcia I need the debug signing to test the uuid thing, so I think I have your amen, merging it.

PR #85, opened 2019-10-31
added proguard metadata

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
adding proguard metadata to the event

we need the proguard file id to deobfuscate stacktraces

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
how do I use sentry-cli?


sentry.properties
```diff
@@ -0,0 +1,3 @@
+defaults.project=android
+defaults.org=man
+auth.token=
```
marandaneto: @bruno-garcia we might need a take token where the server will not refuse it

build.gradle.kts
```diff
@@ -22,6 +22,7 @@ buildscript {
         classpath(Config.BuildPlugins.androidGradle)
         classpath(kotlin(Config.BuildPlugins.kotlinGradlePlugin, version = Config.kotlinVersion))
         classpath(Config.QualityPlugins.errorpronePlugin)
+        classpath("io.sentry:sentry-android-gradle-plugin:1.7.28")
```
bruno-garcia: Can this be only for the sample instead?

sentry.properties
bruno-garcia: We wouldn't want to check-in an auth token though. I'm not sure we could limit it only for uploading proguard files so.
We can set this via environment variable, and `sentry-cli` will pick it up

marandaneto: ok I kept it commented for now and let's discuss it later on

build.gradle.kts
marandaneto: @bruno-garcia this is only for the sample, classpath means available, who wants to add the plugin.

I kept it commented for now and let's discuss it later on

PR #87, opened 2019-10-31
feat: Sample raises sigsegv

bruno-garcia:
Button to raise a sigsegv to help us test the signal handler integration with NDK


sentry-sample/src/main/cpp/native-sample.cpp
```diff
@@ -0,0 +1,15 @@
+#include <string.h>
```
bruno-garcia:
```suggestion
```

bruno-garcia:
@mitsuhiko unless you have objections/comments, I'll merge this.

PR #88, opened 2019-11-01
fixing a bug where gson will blowup with null value for objects

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fixing a bug where gson will blowup with null value for objects

because gson blows up with null values

## :pencil: Checklist

- [X] All tests passing


sentry-android-core/src/test/java/io/sentry/android/core/AndroidSerializerTest.kt
```diff
@@ -228,6 +229,13 @@ class AndroidSerializerTest {
         assertEquals(SentryLevel.DEBUG, actual.level)
     }
 
+    @Test
+    fun `when theres a null value, gson wont blow up`() {
+        val invoke = FileFromResources.invoke("event.json")
+        val deserializeEvent = serializer.deserializeEvent(invoke)
+        assertTrue(deserializeEvent is SentryEvent)
```
bruno-garcia: can we assert: `assertIsNull(deserializeEvent.user)` instead?
Test name, when event has null user, gson deserializes null.

PR #89, opened 2019-11-01
feat: pickup envelopes

bruno-garcia:
@marandaneto @mitsuhiko it's draft but it can give some idea where I'm going. Still need more investigation of the _event types_ raised by `FileObserver` etc.

The current implementation only works for Android due to the `FileObserver` bits.
It also means we'd need to discuss the design we originally had which dictates how files should be ordered by timestamp etc since it's not possible using the Android facility.


sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserverIntegration.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    protected String getPath(SentryOptions options) {
+      return options.getOutboxPath();
+    }
+  }
+
+  private static class JavaCachedEnvelopeFileObserverIntegration extends EnvelopeFileObserverIntegration {
+    @Override
+    protected String getPath(SentryOptions options) {
+      // TODO: Wherever we're caching events from the Java layer
```
bruno-garcia: Another PR to sort out the offline caching for the Java layer. We'll be reading files as they are written?
That means we need to write Envelopes and not do that when it already came from reading an envelope from disk.

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    logger.log(
+      SentryLevel.DEBUG,
+      "onEvent fired for EnvelopeFileObserver with event type %d on path: %s for file %s.",
+      eventType,
+      this.rootPath,
+      relativePath);
+
+    // TODO: Only some event types should be pass through?
```
bruno-garcia: For sure need to deal with the right events, they are:

> Value is either 0 or a combination of ACCESS, MODIFY, ATTRIB, CLOSE_WRITE, CLOSE_NOWRITE, OPEN, MOVED_FROM, MOVED_TO, CREATE, DELETE, DELETE_SELF, and MOVE_SELF

https://developer.android.com/reference/android/os/FileObserver.html

I guess since I delete them on each run, it'll raise another event immediately for that delete.

sentry-android-core/build.gradle.kts
```diff
@@ -79,6 +79,7 @@ dependencies {
 
     // libs
     implementation(Config.Libs.gson)
+    implementation("com.android.support:support-annotations:28.0.0")
```
bruno-garcia: Seems required to implement `FileObserver`.
If that's it, I'll just add it to `Config.Libs.support-annotations`

marandaneto: I'd not say this is required, but if so, prefer the androidx annotations
https://android.googlesource.com/platform/frameworks/support.git/+/refs/heads/androidx-annotation-release/annotation/annotation/src/main/java/androidx/annotation

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
```diff
@@ -0,0 +1,39 @@
...
+  @SuppressWarnings("deprecation")
+  EnvelopeFileObserver(String path, IEnvelopeSender envelopeSender, ILogger logger) {
+    super(path);
+    this.rootPath = Objects.requireNonNull(path, "File path is required.");
+    this.envelopeSender = Objects.requireNonNull(envelopeSender, "Envelope sender is required.");
+    this.logger = Objects.requireNonNull(logger, "Logger is required.");
+  }
+
+  @Override
+  public void onEvent(int eventType, @android.support.annotation.Nullable String relativePath) {
```
marandaneto: this is probably not required `@android.support.annotation.Nullable` but if so, try to use the one we already have which is `org.jetbrains:annotations`

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
marandaneto: if I am not wrong, only `CREATE` right, and when you cache to disk (an event) and it's already enqueued, you still get an event from the Observer, we will need to take care of it to not duplicate the events in the queue

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserverIntegration.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  public static EnvelopeFileObserverIntegration getCachedEnvelopeFileObserver() {
+    return new JavaCachedEnvelopeFileObserverIntegration();
+  }
+
+  @TestOnly
+  abstract String getPath(SentryOptions options);
+
+  private static class OutboxEnvelopeFileObserverIntegration
```
marandaneto: you can make it final too

sentry-android-core/src/main/java/io/sentry/android/core/NdkIntegration.java
```diff
@@ -0,0 +1,35 @@
...
+import io.sentry.core.SentryOptions;
+import java.lang.reflect.Method;
+
+final class NdkIntegration implements Integration {
+  private static boolean isNdkAvailable() {
+    return Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP;
+  }
+
+  @Override
+  public void register(IHub hub, SentryOptions options) {
```
marandaneto: does it make sense to implement `Integration` if we don't use `hub` at all?

marandaneto: maybe we may have another `interface` which is also an `Integration` only with `SentryOptions`, still respecting ISP principle.

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
```diff
@@ -0,0 +1,107 @@
...
+      if (envelope == null) {
+        logger.log(SentryLevel.ERROR, "Stream from path %s resulted in a null envelope.", path);
+      } else {
+        processEnvelope(envelope);
+      }
+    } catch (Exception e) {
+      logger.log(SentryLevel.ERROR, "Error processing envelope.", e);
+    } finally {
+      try {
+        if (stream != null) stream.close();
```
marandaneto: you may use `try-with-resources`

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
```diff
@@ -0,0 +1,107 @@
...
+            //              logger.log(
+            //                  SentryLevel.ERROR,
+            //                  "Item %d of has a different event id (%s) to the envelope header
+            // (s)",
+            //                  items,
+            //                  envelope.getHeader().getEventId(),
+            //                  event.getEventId());
+            //              continue;
+            //            }
+            hub.captureEvent(event);
```
marandaneto: why is it the `hub` instead of `Sentry.captureEvent`?

sentry-core/src/main/java/io/sentry/core/ISerializer.java
```diff
@@ -1,10 +1,13 @@
 package io.sentry.core;
 
 import java.io.IOException;
+import java.io.Reader;
 import java.io.Writer;
 
 public interface ISerializer {
   SentryEvent deserializeEvent(String envelope);
 
+  SentryEvent deserializeEvent(Reader eventReader);
```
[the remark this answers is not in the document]
marandaneto: agreed

sentry-sample/src/main/AndroidManifest.xml
```diff
@@ -2,7 +2,6 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     package="io.sentry.sample">
-  <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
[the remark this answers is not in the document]
marandaneto: `ACCESS_NETWORK_STATE` should be opt-in IMHO, the SDK may work without it

sentry-core/src/test/java/io/sentry/core/EnvelopeSenderTest.kt
```diff
@@ -0,0 +1,124 @@
...
+
+        fun getSut(): EnvelopeSender {
+            return EnvelopeSender(hub, envelopeReader, serializer, logger)
+        }
+    }
+
+    private val fixture = Fixture()
+
+    private fun getTempEnvelope(): String {
+        val testFile = this::class.java.classLoader.getResource("envelope-event-attachment.txt")
```
marandaneto: I've created the class `FileFromResources` to make it easier, we can generalize it or create a new method in there to get the bytes

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
```diff
@@ -0,0 +1,39 @@
+package io.sentry.android.core;
+
+import android.os.FileObserver;
+import io.sentry.core.*;
+import io.sentry.core.util.Objects;
+
+final class EnvelopeFileObserver extends FileObserver {
```
marandaneto: Java has the `WatchService` but it's only available API >= 26 for Android.
https://developer.android.com/reference/java/nio/file/WatchService

marandaneto: for `FileObserver` pay attention to that:

> Warning: If a FileObserver is garbage collected, it will stop sending events. To ensure you keep receiving events, you must keep a reference to the FileObserver instance from some other live object.

marandaneto: there are backports of `WatchService` for Android, like https://github.com/lukhnos/nnio but not sure how trustable it is

marandaneto: apache commons-vfs also has classes for watching a dir.

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
bruno-garcia: The file is deleted once we handle it so unless it's fired in parallel, it won't end up as duplicate events. But in regards it would be hitting the `FileNotFound` so we should opt-in to some event types. IIRC per spec it would be a file rename where it's not prefixed by `.`

sentry-android-core/build.gradle.kts
bruno-garcia: The annotation is part of the signature and if I deleted it, it wouldn't compile.
I can try getting rid of it again but I recall not building otherwise

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
bruno-garcia: Couldn't choose whether to add it or not, nor the package it comes from. See above, will try again before adding the package to `Config`

sentry-android-core/src/main/java/io/sentry/android/core/NdkIntegration.java
bruno-garcia: The idea of `Integration` is something that emit events, for that it requires `Hub` but in the case of NDK we're using reflection to get a type and initialize it and using the file system as a way to communicate with the SDK.

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
bruno-garcia: Don't think we can use `try-with-resource` is API 19 or higher

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
bruno-garcia: Testability is one reason. Sentry static is an entrypoint that will either clone the current hub on a new thread or use what's on TLS. I didn't see the need of doing that here, we can use the hub provided by the ctor.

sentry-core/src/test/java/io/sentry/core/EnvelopeSenderTest.kt
bruno-garcia: Could you please clarify the value it adds? Here I'm just reading the resource in a single line. Not sure what `FileFromResources` is doing, it has a loop and adds line breaks? I wouldn't want the data to be modified.

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
bruno-garcia: It won't be GC'ed because the reference is held by the integration.
Why use `WatchService` over `FileObserver`?

sentry-android-core/build.gradle.kts
marandaneto: `android.support` is deprecated and because of that, they made a tool called `jetifier` to migrate `android.support` . to `androidx` libs., but `androidx` is dependent of the framework so would be better to use `jetbrains-annotations` to be able to reuse/understand on the pure java layer.
ref https://developer.android.com/studio/command-line/jetifier

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
marandaneto: `In addition to the Java 8 language features and APIs above, Android Studio 3.0 and later extends support for try-with-resources to all Android API levels.`
ref https://developer.android.com/studio/write/java8-support

`AutoCloseable` is min API 19, try-with-resources is fine afaik

sentry-core/src/test/java/io/sentry/core/EnvelopeSenderTest.kt
marandaneto: In this case, you just need the bytes to write on disk, so you are right and might not be suitable.
but the idea of the class would be a `Utils` for working with `files/json` content so we DRY, nothing more.
not sure why I've added `.append("\n")` either, maybe just for better visibility while debugging, can be safely removed as this would batter for envelopes.

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
marandaneto: if it would be possible `WatchService` is pure-java, so not dependent on the Android FW and could be on sentry-core, but it's min. API 26 so :(

sentry-sample/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
bruno-garcia: Added back.

sentry-android-core/src/main/java/io/sentry/android/core/NdkIntegration.java
bruno-garcia: I've added a comment to clarify

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
bruno-garcia: OK, @metlos had to get rid of some try-with-resource on his HTTP PR, but I guess at the time we didn't know about this desugaring (I still didn't know).

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
bruno-garcia: Makes sense. For now let's target only `sentry-android` which requires API higher than 26 so I'd say `FileObserver` should be OK. After the 25th we can put all options back on the table. 
Once backend java becomes priority this will come up anyway. In fact, since there's no NDK there we might just not have this at all. Only the ReadOnStartup behavior

marandaneto: agreed

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
```diff
@@ -0,0 +1,40 @@
...
+    logger.log(
+        SentryLevel.DEBUG,
+        "onEvent fired for EnvelopeFileObserver with event type %d on path: %s for file %s.",
+        eventType,
+        this.rootPath,
+        relativePath);
+
+    // TODO: Only some event types should be pass through?
+
+    envelopeSender.processEnvelopeFile(this.rootPath + "/" + relativePath);
```
marandaneto: `File.separator` here

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserverIntegration.java
```diff
@@ -0,0 +1,57 @@
...
+      return options.getOutboxPath();
+    }
+  }
+
+  private static final class JavaCachedEnvelopeFileObserverIntegration
+      extends EnvelopeFileObserverIntegration {
+    @Override
+    protected String getPath(final SentryOptions options) {
+      // TODO: Wherever we're caching events from the Java layer
+      return options.getCacheDirPath() + "/cached";
```
marandaneto: File.separator here

PR #92, opened 2019-11-03
[deferred] Implement flush in SentryClient.

metlos:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Implement flushing of queued events in `SentryClient`.

Note that during the testing of this PR I found a race condition in the `RetryingThreadPoolExecutorTest` (just the test code, not the actual impl). It might be possible that the tests will hang when then stars align right. I ran the test suite over 300 times and encountered  1 such occurrence (which seems to be fixed by the changes in the `limits the queue size` test).

Flush is important for ensuring events are sent.

by unit tests

## :pencil: Checklist
- [x] All tests passing

### DIFF
--- a/sentry-core/src/test/java/io/sentry/core/SentryClientTest.kt
+++ b/sentry-core/src/test/java/io/sentry/core/SentryClientTest.kt
@@ -1,10 +1,12 @@
 package io.sentry.core
 
+import com.nhaarman.mockitokotlin2.any
 import com.nhaarman.mockitokotlin2.mock
 import com.nhaarman.mockitokotlin2.mockingDetails
 import com.nhaarman.mockitokotlin2.never
 import com.nhaarman.mockitokotlin2.times
 import com.nhaarman.mockitokotlin2.verify
+import com.nhaarman.mockitokotlin2.whenever
 import io.sentry.core.protocol.User
 import io.sentry.core.transport.AsyncConnection
 import kotlin.test.Ignore
@@ -12,6 +14,7 @@ import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
 import kotlin.test.assertTrue
+import org.junit.Before
 
 class SentryClientTest {
 
@@ -24,6 +27,11 @@ class SentryClientTest {
     }
 
     private val fixture = Fixture()
+
+    @Before
+    fun setup() {
+        whenever(fixture.connection.flush(any(), any())).thenReturn(mock())
+    }
 
     @Test
     fun `when fixture is unchanged, client is enabled`() {
--- a/sentry-core/src/test/java/io/sentry/core/transport/RetryingThreadPoolExecutorTest.kt
+++ b/sentry-core/src/test/java/io/sentry/core/transport/RetryingThreadPoolExecutorTest.kt
@@ -157,15 +157,156 @@ class RetryingThreadPoolExecutorTest {
         synchronized(jobBlocker) { jobBlocker.notify() }
         atLeastOneFinished.await()
 
+        // wait for the thread pool to realize that the job indeed finished
+        while (threadPool?.completedTaskCount == 0L) {
+            Thread.sleep(100)
+        }
+
         // now try to test that the above actually made room in the queue again
         val jobBlocker2 = CountDownLatch(1)
         val sync2 = CountDownLatch(1)
 
         f = threadPool?.submit { sync2.countDown(); jobBlocker2.await() }
-        sync2.await()
         assertFalse(f != null && f.isCancelled, "A task should be successfully enqueued after making a place in the queue")
+        sync2.await()
 
         synchronized(jobBlocker) { jobBlocker.notifyAll() }
         jobBlocker2.countDown()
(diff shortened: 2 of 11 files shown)


sentry-core/src/test/java/io/sentry/core/SentryClientTest.kt
```diff
@@ -1,17 +1,20 @@
...
 import com.nhaarman.mockitokotlin2.verify
+import com.nhaarman.mockitokotlin2.whenever
 import io.sentry.core.protocol.User
 import io.sentry.core.transport.AsyncConnection
 import kotlin.test.Ignore
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
 import kotlin.test.assertTrue
+import org.junit.Before
```
marandaneto: use `@BeforeTest` from `kotlin.test`

sentry-core/src/test/java/io/sentry/core/transport/RetryingThreadPoolExecutorTest.kt
```diff
@@ -157,15 +157,156 @@ class RetryingThreadPoolExecutorTest {
         synchronized(jobBlocker) { jobBlocker.notify() }
         atLeastOneFinished.await()
 
+        // wait for the thread pool to realize that the job indeed finished
+        while (threadPool?.completedTaskCount == 0L) {
```
marandaneto: it would be also possible with `kotlinx.coroutines` using `runBlocking`, but it's only al test, never mind.

metlos:
The flush is quite complicated because I thought that we should restrict the number of flushes that happen concurrently. The more I think about it, the less right it seems to be because it breaks the premise of "wait until all the currently enqueued jobs are finished". 

But that is hard to achieve as well, because it doesn't lend itself to easy implementation in the light of multiple concurrently running flushes + jobs added after a flush started finishing earlier than some of the jobs created before the flush (which the flush should track). I think the dotnet impl suffers from the same problem, but not sure. What do you think @bruno-garcia ?

I will take a look at this once more...

bruno-garcia:
It's indeed complicated. In .NET there's a in-memory queue and a single thread reading off of it and starting asynchronous network I/O. That makes it simpler already. Once an event (Sentry event) is flushed, an event (.NET event) is fired. On Flush (that can be called in parallel by as many callers as needed) will subscribe to that event and with `Interlock` increment a counter. Once subscribed, it takes the current items in the queue and asynchronously waits until _at least_ that number of messages are processed. At which point it completes the Task (aka Promise, Future) which will "unblock" the caller.

This does the job since the goal of `flush` is to make sure the caller gets an event it might have itself captured. It doesn't care about events that were queued _after_ it called `flush`.

Since for the Android implementation we'll need to store **all** events to disk to make sure we don't lose any in case the app is put to sleep, we only really need `close` for now.

Let's postpone this `flush` for now. Maybe it'll require redesigning the whole inner bits to fit it more easily.

metlos:
Ok, so here I assume that the Sentry server actually supports receiving events out of order (and reordering them based on their client-originating timestamps).

In .NET the flush works because the background worker is actually a single thread handling the whole event queue (as far as I looked).

So the question is - do we need a proper thread pool with potentially more than 1 thread sending the events? If yes, then I think we need to rethink the function and guarantees of `flush` otherwise the correct implementation is going to be quite expensive (to handle the "started-after-finished-before-flush" scenario that I tried to describe above).

I 100% agree this requires more thought and understanding before we commit to some implementation.

bruno-garcia:
> Ok, so here I assume that the Sentry server actually supports receiving events out of order (and reordering them based on their client-originating timestamps).

That's right.

> In .NET the flush works because the background worker is actually a single thread handling the whole event queue (as far as I looked).

For each event it might be a different thread since it takes from the pool, and awaits asynchronously for the I/O (`async`/`await`) but essentially a single event in-flight at any time, that's right. if it were to change the single thread would store the `Task` in a list and keep N in there while they are not complete. Just never seem to require that kind of implementation so far.

It's worth noting that Sentry will throttle any single project sending more than 100 events per second. At that rate you're likely just going to drain your quota and unlike these are all different events which will be actionable.

That said I do think it's fine to have the "queue" be in fact events being sent concurrently. Though that means 10 threads taken by the SDK to flush events out to Sentry and also introduces the complexity we're dealing with. 

>So the question is - do we need a proper thread pool with potentially more than 1 thread sending the events?

I think I just answered that with my random thoughts above, and the answer is no. Specially on Android. I believe we should make an effort not to overload the device with lots of I/O and threads running. We can capture an event, store it to disk and try to send it. If it fails because the device is offline, we'll try later (maybe on restart, maybe in a few minutes). Otherwise we'll drop the event. IMHO we can do 1 event at a time, single thread being used. Read it off of capped queue, one at a time.
What do you think?

I guess we could come to different conclusion if we were talking about a 40 core server running a single Spring Boot application which is online 99.999%

marandaneto:
as it is deferred, I'll close it.

PR #93, opened 2019-11-04
added try block to before callbacks and logged if an error has occurred

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
added try block to before callbacks and logged if an error has occurred

currently, if an error happens on before callbacks, App. may crash.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Hub.java
+++ b/sentry-core/src/main/java/io/sentry/core/Hub.java
@@ -4,7 +4,11 @@ import static io.sentry.core.ILogger.logIfNotNull;
 
 import io.sentry.core.protocol.SentryId;
 import io.sentry.core.util.Objects;
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.util.Deque;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.LinkedBlockingDeque;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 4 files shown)


sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -117,6 +117,34 @@ public SentryId captureEvent(SentryEvent event, @Nullable Scope scope) {
...
+      } catch (Exception e) {
+        logIfNotNull(
+            options.getLogger(),
+            SentryLevel.ERROR,
+            "The BeforeSend callback threw an exception. It will be added as breadcrumb and continue.",
+            e);
+
+        Breadcrumb breadcrumb = new Breadcrumb();
+        breadcrumb.setMessage("BeforeSend callback failed.");
+        breadcrumb.setCategory("SentryClient");
```
bruno-garcia: `category` is usually all lower case. Also, the[ docs has more info](https://docs.sentry.io/development/sdk-dev/event-payloads/breadcrumbs/):

> Optional. A dotted string indicating what the crumb is or from where it comes. Typically it is a module name or a descriptive string. For instance, ui.click could be used to indicate that a click happened in the UI or flask could be used to indicate that the event originated in the Flask framework.

Maybe something like?
```suggestion
        breadcrumb.setCategory("callback");
```

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -190,7 +194,7 @@ public void addBreadcrumb(Breadcrumb breadcrumb) {
       if (item != null) {
         SentryOptions.BeforeBreadcrumbCallback callback = options.getBeforeBreadcrumb();
         if (callback != null) {
-          breadcrumb = callback.execute(breadcrumb);
+            breadcrumb = executeBeforeBreadcrumb(callback, breadcrumb);
```
bruno-garcia:
```suggestion
          breadcrumb = executeBeforeBreadcrumb(callback, breadcrumb);
```

PR #96, opened 2019-11-04
fixed warnings detected by linting

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
fixed warnings detected by linting

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Hub.java
+++ b/sentry-core/src/main/java/io/sentry/core/Hub.java
@@ -13,13 +13,13 @@ import java.util.concurrent.LinkedBlockingDeque;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-public final class Hub implements IHub, Cloneable {
+public final class Hub implements IHub {
 
   private static final class StackItem {
     private volatile @NotNull ISentryClient client;
     private volatile @NotNull Scope scope;
 
-    public StackItem(@NotNull ISentryClient client, @NotNull Scope scope) {
+    StackItem(@NotNull ISentryClient client, @NotNull Scope scope) {
       this.client = client;
       this.scope = scope;
     }
@@ -46,7 +46,7 @@ public final class Hub implements IHub, Cloneable {
     this.lastEventId = SentryId.EMPTY_ID;
   }
 
-  static StackItem createRootStackItem(@NotNull SentryOptions options) {
+  private static StackItem createRootStackItem(@NotNull SentryOptions options) {
     Objects.requireNonNull(options, "SentryOptions is required.");
     Scope scope = new Scope(options.getMaxBreadcrumbs());
     ISentryClient client = new SentryClient(options);
@@ -173,9 +173,8 @@ public final class Hub implements IHub, Cloneable {
         }
       } catch (Exception e) {
         logIfNotNull(options.getLogger(), SentryLevel.ERROR, "Error while closing the Hub.", e);
-      } finally {
-        isEnabled = false;
       }
+      isEnabled = false;
     }
   }
 
@@ -294,22 +293,18 @@ public final class Hub implements IHub, Cloneable {
           "Instance is disabled and this 'withScope' call is a no-op.");
     } else {
       pushScope();
-      try {
-        StackItem item = stack.peek();
-        if (item != null) {
-          try {
-            callback.run(item.scope);
-          } catch (Exception e) {
-            logIfNotNull(
-                options.getLogger(), SentryLevel.ERROR, "Error in the 'withScope' callback.", e);
-          }
-        } else {
+      StackItem item = stack.peek();
+      if (item != null) {
+        try {
+          callback.run(item.scope);
+        } catch (Exception e) {
           logIfNotNull(
-              options.getLogger(), SentryLevel.FATAL, "Stack peek was null when withScope");
+              options.getLogger(), SentryLevel.ERROR, "Error in the 'withScope' callback.", e);
         }
-      } finally {
-        popScope();
+      } else {
+        logIfNotNull(options.getLogger(), SentryLevel.FATAL, "Stack peek was null when withScope");
       }
+      popScope();
     }
   }
 
--- a/sentry-core/src/main/java/io/sentry/core/SentryEnvelopeItem.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryEnvelopeItem.java
@@ -5,7 +5,7 @@ public final class SentryEnvelopeItem {
   // TODO: Can we have a slice or a reader here instead?
   private final byte[] data;
 
-  public SentryEnvelopeItem(SentryEnvelopeItemHeader header, byte[] data) {
+  SentryEnvelopeItem(SentryEnvelopeItemHeader header, byte[] data) {
     this.header = header;
     this.data = data;
   }
--- a/sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
+++ b/sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
@@ -19,7 +19,7 @@ public final class UncaughtExceptionHandlerIntegration
   private boolean isRegistered = false;
   private UncaughtExceptionHandler threadAdapter;
 
-  public UncaughtExceptionHandlerIntegration() {
+  UncaughtExceptionHandlerIntegration() {
     this(UncaughtExceptionHandler.Adapter.INSTANCE);
   }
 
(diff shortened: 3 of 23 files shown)


sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -173,9 +173,8 @@ public void close() {
         }
       } catch (Exception e) {
         logIfNotNull(options.getLogger(), SentryLevel.ERROR, "Error while closing the Hub.", e);
-      } finally {
-        isEnabled = false;
       }
+      isEnabled = false;
```
marandaneto: There is a severe performance penalty for using finalizers and this line of code could be done without a `finally`

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -294,22 +293,18 @@ public void withScope(@NotNull ScopeCallback callback) {
...
           logIfNotNull(
-              options.getLogger(), SentryLevel.FATAL, "Stack peek was null when withScope");
+              options.getLogger(), SentryLevel.ERROR, "Error in the 'withScope' callback.", e);
         }
-      } finally {
-        popScope();
+      } else {
+        logIfNotNull(options.getLogger(), SentryLevel.FATAL, "Stack peek was null when withScope");
       }
+      popScope();
```
marandaneto: this try/catch only wraps `peek` which only returns null if empty.
`popScope` should be out of a `finally` block and if needed, to have it's own try/catch.

> Uncaught exceptions inside a finalizer won't even print a warning

sentry-core/src/main/java/io/sentry/core/Hub.java
ninniuz: With `finalizers` you mean calling `Object.finalize()`? I don't know of any penalty in using a `finally` block.

marandaneto: @ninniuz yes, the `Object.finalize()`, but I could not explain correctly, I mixed up the meaning.
Usually for closing/cleaning is common to have a `finalizer`, but it's bad on Java.
In this case, as we don't have any `return`, the `finally` is just not necessary.

sentry-core/src/main/java/io/sentry/core/Hub.java
marandaneto: same here, I removed the quoting as I mixed up the meaning, but the change still applies.

bruno-garcia: I'm OK with the change but just to clarify:

I tend to leave stuff that must happen within the `try` block in a `finally` block also for clarity. 
We must keep in mind in future changes that a early `return` added in either `try` or `catch` blocks would break the code.

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeItem.java
```diff
@@ -5,7 +5,7 @@
   // TODO: Can we have a slice or a reader here instead?
   private final byte[] data;
 
-  public SentryEnvelopeItem(SentryEnvelopeItemHeader header, byte[] data) {
+  SentryEnvelopeItem(SentryEnvelopeItemHeader header, byte[] data) {
```
bruno-garcia: Doesn't this means this type can't be constructed outside of this package?
That means this isnt' just a refactor but it changes the API

sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
```diff
@@ -19,7 +19,7 @@
   private boolean isRegistered = false;
   private UncaughtExceptionHandler threadAdapter;
 
-  public UncaughtExceptionHandlerIntegration() {
+  UncaughtExceptionHandlerIntegration() {
```
bruno-garcia: Same as above. Now there's no public constructor. Is that what we want?

marandaneto: @bruno-garcia right now it's a default integration and could be only removed as it's added on `init()`.
I would see one removing from `SentryOptions.getIntegrations()`, for that you won't need a ctor.

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeItem.java
marandaneto: @bruno-garcia I only see it being used by `EnvelopeReader`, why would we expose it?

sentry-core/src/main/java/io/sentry/core/Hub.java
marandaneto: @bruno-garcia got it, but it's a `void` method, an early return will break compilation or tests.

marandaneto: @bruno-garcia got it, but it's a `void` method, an early return will break compilation or tests.

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeItem.java
bruno-garcia: Should we make the whole type internal?
Seems like we can get around with the whole Envelope thing internal for now.

Only once we have a JVM based integration we'd need to open this up.

sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
bruno-garcia: Right, so we don't want users of this package to be able to use this integration by themselves. Outside of `Sentry.init`. 
Happy with that. It's always easier to open it later as opposed to restricting access.

sentry-core/src/main/java/io/sentry/core/Hub.java
ninniuz: @bruno-garcia Honestly I do not like driving the business logic inside `finally` blocks as it somehow breaks readability and control flow. I only use it to clean up critical resources in a safe way.

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -294,22 +293,18 @@ public void withScope(@NotNull ScopeCallback callback) {
...
+          callback.run(item.scope);
+        } catch (Exception e) {
           logIfNotNull(
-              options.getLogger(), SentryLevel.FATAL, "Stack peek was null when withScope");
+              options.getLogger(), SentryLevel.ERROR, "Error in the 'withScope' callback.", e);
         }
-      } finally {
-        popScope();
+      } else {
+        logIfNotNull(options.getLogger(), SentryLevel.FATAL, "Stack peek was null when withScope");
```
ninniuz: I am seeing a lot of calls to `logIfNotNull` using `options.getLogger()` as the `ILogger` instance, which is marked as `@NotNull` and safely handled as a non-null value in `SentryOptions`.
What is the purpose of checking for nullability in the log call then?

sentry-core/src/main/java/io/sentry/core/Hub.java
ninniuz: @bruno-garcia I understand your point but I feel like driving business logic inside a `finally` block breaks readability and control flow. I tend to use it only to cleanup critical resources in a safe way.

sentry-core/src/main/java/io/sentry/core/Hub.java
ninniuz: What is the purpose of using `logIfNotNull` with `options.getLogger()` returns a `@NonNull` value?

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeItem.java
marandaneto: yep, right now as we have different packages over the modules, it might not be fully possible, maybe for Envelopes, but not for all the cases, we should work on that before the first release.

sentry-core/src/main/java/io/sentry/core/Hub.java
marandaneto: @ninniuz so basically this was supposed to be like an extension method, and I guess that time `logger` could be `null`.
I'd say we need to revisit the whole `log` thing to work as `.net`, a single `log` call that checks nullability, if it could be null, if `debug` is enabled and also if the default `logLevel` or the one set by the client is >= the given `logLevel`, we have this task already and it's a good point.

marandaneto: answered here https://github.com/getsentry/sentry-android/pull/96#discussion_r342470530

PR #97, opened 2019-11-04
feat: uncaught ex, crashed=true, handled=false

bruno-garcia:


bruno-garcia:
@HazAT @jan-auer @marandaneto Following up with our meeting wrt Threads panel expecting a crashed thread.
When uncaught exception is captured at least we'll mark the `crashed=true` on the right thread an mechanism `handled=false`.

Next we need to figure out the `errored=true` for handled exceptions.

sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
```diff
@@ -47,6 +45,24 @@ public SentryEvent process(SentryEvent event) {
       event.setExceptions(sentryExceptionFactory.getSentryExceptions(throwable));
     }
 
+    if (event.getThreads() == null) {
+      Long crashedThreadId = null;
+      List<SentryException> exceptions = event.getExceptions();
+      if (event.getExceptions() != null && exceptions.size() > 0) {
```
marandaneto: you may use `!exceptions.isEmpty()`

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
```diff
@@ -122,7 +122,8 @@ public void setThreads(List<SentryThread> threads) {
   }
 
   public List<SentryException> getExceptions() {
-    return exception.getValues();
+    // We dream of exception?.getValues() like any other language in 2019
```
marandaneto: Java has `Optional` but it's API 24.
but please remove this comment.

sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
```diff
@@ -58,8 +62,10 @@ public void uncaughtException(Thread thread, Throwable thrown) {
     logIfNotNull(options.getLogger(), SentryLevel.INFO, "Uncaught exception received.");
 
     try {
-      // TODO: Set Thread info to the scope?
-      this.hub.captureException(thrown);
+      Throwable throwable = getUnhandledThrowable(thread, thrown);
+      this.hub.captureException(throwable);
+      // Close the SDK to flush the event to disk before shutting down.
+      this.hub.close();
```
marandaneto: right now `flush` is not implemented but getting `options.getShutdownTimeout()`
`connection.close();` awaits termination for 1 minute, should `close` also respect `options.getShutdownTimeout()`?

sentry-core/src/main/java/io/sentry/core/exception/ExceptionMechanismThrowable.java
```diff
@@ -1,26 +1,32 @@
...
 import io.sentry.core.protocol.Mechanism;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * A throwable decorator that holds an {@link io.sentry.core.protocol.Mechanism} related to the
  * decorated {@link Throwable}.
  */
 @SuppressWarnings("serial")
 public final class ExceptionMechanismThrowable extends Throwable {
+  private static final long serialVersionUID = 142345454265713915L;
```
marandaneto: we are not serializing this class, but converting to our `SentryException`, why do we need that? if so, you may remove `@SuppressWarnings("serial")`

PR #98, opened 2019-11-04
Implementation of DiskCache and the ability to store events before sending

metlos:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
This adds the implementation of the `IEventCache` interface that stores the data to the filesystem.
Also it adds a simple ability of the `AsyncConnection` to store the events before attempting to send.
The implementation of the `DiskCache` is largely copied from `sentry-java`.

While a new config property governing the maximum of cached events is added in this PR, the user has no choice in selecting the implementation of the cache. Maybe this should be changed?

We need this to be able to persist the events before sending to enable crash/kill recovery.

Just unit tests, no testing on a real/simulated Android device

## :pencil: Checklist

- [x] All tests passing

## :crystal_ball: Next steps
- [ ] add ability to set the `IEventCache` implementation in `SentryOptions` as it is done for `ISerializer`?
- [ ] is sharing the threadpool for sending and storing events the right thing to do?

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/cache/DiskCache.java
+++ b/sentry-core/src/main/java/io/sentry/core/cache/DiskCache.java
@@ -1,0 +1,152 @@
+package io.sentry.core.cache;
+
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+import static io.sentry.core.ILogger.logIfNotNull;
+import static io.sentry.core.SentryLevel.DEBUG;
+import static io.sentry.core.SentryLevel.ERROR;
+import static io.sentry.core.SentryLevel.WARNING;
+
+import java.io.BufferedInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+import io.sentry.core.ISerializer;
+import io.sentry.core.SentryEvent;
+import io.sentry.core.SentryLevel;
+import io.sentry.core.SentryOptions;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A simple cache implementation storing the events to a disk, each event in a separater file in a configured directory.
+ */
+public final class DiskCache implements IEventCache {
+  /**
+   * File suffix added to all serialized event files.
+   */
+  public static final String FILE_SUFFIX = ".sentry-event";
+
+  @SuppressWarnings("CharsetObjectCanBeUsed")
+  private static final Charset UTF8 = Charset.forName("UTF-8");
+
+  private final File directory;
+  private final int maxSize;
+  private final ISerializer serializer;
+  private final SentryOptions options;
+
+  public DiskCache(SentryOptions options) {
+    this.directory = new File(options.getCacheDirPath());
+    this.maxSize = options.getCacheDirSize();
+    this.serializer = options.getSerializer();
+    this.options = options;
+
+    try {
+      //noinspection ResultOfMethodCallIgnored
+      directory.mkdirs();
+      checkDirectoryValid();
+    } catch (IllegalStateException e) {
+      throw e;
+    } catch (Exception e) {
+      throw new IllegalStateException("Failed to initialize the directory for caching Sentry events.", e);
+    }
+  }
+
+  @Override
+  public void store(SentryEvent event) {
+    if (getNumberOfStoredEvents() >= maxSize) {
+      logIfNotNull(options.getLogger(), SentryLevel.WARNING, "Disk cache full. Not storing event {}", event);
+      return;
+    }
+
+    File eventFile = getEventFile(event);
+    if (eventFile.exists()) {
+      logIfNotNull(options.getLogger(), DEBUG,
+        "Not adding Event to offline storage because it already exists: %s", eventFile.getAbsolutePath());
+      return;
+    } else {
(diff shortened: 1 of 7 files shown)


sentry-core/src/main/java/io/sentry/core/cache/DiskCache.java
```diff
@@ -0,0 +1,152 @@
...
+          format("Error while reading cached event from file %s", f.getAbsolutePath()), e);
+      }
+    }
+
+    return ret.iterator();
+  }
+
+  private File[] allEventFiles() {
+    checkDirectoryValid();
+    return requireNonNull(directory.listFiles((__, fileName) -> fileName.endsWith(FILE_SUFFIX)));
```
bruno-garcia: Can we order by oldest to the newest here?

sentry-core/src/main/java/io/sentry/core/cache/DiskCache.java
```diff
@@ -0,0 +1,152 @@
...
+      serializer.serialize(event, wrt);
+    } catch (Exception e) {
+      logIfNotNull(options.getLogger(), ERROR, "Error writing Event to offline storage: %s", event.getEventId());
+    }
+  }
+
+  @Override
+  public void discard(SentryEvent event) {
+    File eventFile = getEventFile(event);
+        if (eventFile.exists()) {
```
bruno-garcia: `exists` followed by `delete` seems redundant or racey. Is there a way to get`delete` to returns whether the file was there and deleted or didn't exist?

sentry-core/src/main/java/io/sentry/core/cache/DiskCache.java
```diff
@@ -0,0 +1,152 @@
...
+            logIfNotNull(options.getLogger(), DEBUG, "Discarding event from cache: %s",
+              eventFile.getAbsolutePath());
+
+            if (!eventFile.delete()) {
+                logIfNotNull(options.getLogger(), WARNING, "Failed to delete Event: %s", eventFile.getAbsolutePath());
```
bruno-garcia: Would the be `ERROR` instead? I assume we'd be trying to send this again in the future if not deleted.
Or maybe there's another worker moving our files (like did we spawn a second instance that reads cached events?)

sentry-core/src/main/java/io/sentry/core/cache/DiskCache.java
```diff
@@ -0,0 +1,152 @@
...
+  @Override
+  public void store(SentryEvent event) {
+    if (getNumberOfStoredEvents() >= maxSize) {
+      logIfNotNull(options.getLogger(), SentryLevel.WARNING, "Disk cache full. Not storing event {}", event);
+      return;
+    }
+
+    File eventFile = getEventFile(event);
+    if (eventFile.exists()) {
+      logIfNotNull(options.getLogger(), DEBUG,
```
bruno-garcia: I'd say this is at least a warning. Probably an error. I think that becase this shouldn't be happening and if it does we likely have a bug.

sentry-core/src/main/java/io/sentry/core/cache/DiskCache.java
metlos: Well, the check is there merely for correct reporting to the log. We handle the failure to delete by checking the return value of `eventFile.delete()` in the code below, but unfortunately the File API doesn't provide the reason for the failure. `java.nio.files.Files#delete` is better in that regard because it throws an exception, but I assume it is not supported on our minimum Android API level.

sentry-core/src/main/java/io/sentry/core/cache/DiskCache.java
```diff
@@ -0,0 +1,152 @@
...
+        logIfNotNull(options.getLogger(), ERROR,
+          format("Error while reading cached event from file %s", f.getAbsolutePath()), e);
+      }
+    }
+
+    return ret.iterator();
+  }
+
+  private File[] allEventFiles() {
+    checkDirectoryValid();
```
bruno-garcia: Do we need to run this before each access? Since this runs in the constructor (which means I/O in the ctor) can we get away with just doing that in the ctor?
Is the goal to report early if some other process changed the directory underneath us? Or disk is full or else?
I dunno if each of those does an I/O call but sounds like we'd be:
> isDirectory,canRead,canWrite,fileExists,fileWrite

Are these 6 I/O calls for each write?
And do we do the same for delete except the last one is _fileDelete instead of write_?

Not saying it's a problem but if the goal is to have better logging, maybe we only run the checks if logging is enabled? If there's no logging just attempt the operation we intended and .. I guess can't log the error.

sentry-core/src/main/java/io/sentry/core/cache/DiskCache.java
```diff
@@ -0,0 +1,152 @@
...
+
+    try {
+      //noinspection ResultOfMethodCallIgnored
+      directory.mkdirs();
+      checkDirectoryValid();
+    } catch (IllegalStateException e) {
+      throw e;
+    } catch (Exception e) {
+      throw new IllegalStateException("Failed to initialize the directory for caching Sentry events.", e);
+    }
```
marandaneto: I'd prefer to do this check only on `AndroidOptionsInitializer.initializeCacheDirs` maybe copying this piece of code in there, as the NDK needs this folder as well and it should be writable.
also at least for Android, if you `canWrite`, automatically you `canRead`, unless you root your device and change `chmod` manually.

sentry-core/src/main/java/io/sentry/core/cache/DiskCache.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  @Override
+  public void store(SentryEvent event) {
+    if (getNumberOfStoredEvents() >= maxSize) {
```
marandaneto: @bruno-garcia should we not drop the oldest and cache the new one in here?

PR #101, opened 2019-11-05
deep-copy to scope and its fields

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
copy by Serialization is too slow:
https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/SerializationUtils.java

Scope should clone its fields

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Breadcrumb.java
+++ b/sentry-core/src/main/java/io/sentry/core/Breadcrumb.java
@@ -1,17 +1,20 @@
 package io.sentry.core;
 
 import java.util.Date;
+import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import org.jetbrains.annotations.TestOnly;
 
 public final class Breadcrumb implements Cloneable, IUnknownPropertiesConsumer {
+
   private Date timestamp;
   private String message;
   private String type;
   private Map<String, String> data;
   private String category;
   private SentryLevel level;
-
-  @SuppressWarnings("UnusedVariable")
   private Map<String, Object> unknown;
 
   public Date getTimestamp() {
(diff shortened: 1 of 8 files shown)


sentry-core/src/main/java/io/sentry/core/Breadcrumb.java
```diff
@@ -66,4 +69,50 @@ public void setLevel(SentryLevel level) {
...
+    }
+
+    if (unknown != null) {
+      Map<String, Object> unknownClone = new HashMap<>();
+
+      for (Map.Entry<String, Object> item : unknown.entrySet()) {
+        if (item != null) {
+          unknownClone.put(
+              item.getKey(),
+              item.getValue()); // TODO: how do we clone an object that we dont know the shape of it
```
bruno-garcia: We can try to cast it to `Cloneable` and call clone though.

Otherwise just:
```suggestion
              item.getValue()); // NOTE: We shallow copy objects
```

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -82,7 +86,54 @@ public void setExtra(String key, String value) {
...
-    return (Scope) super.clone();
+  public Scope clone() throws CloneNotSupportedException {
+    Scope clone = (Scope) super.clone();
+    clone.level = level != null ? SentryLevel.valueOf(level.name().toUpperCase(Locale.ROOT)) : null;
+    clone.user = user != null ? user.clone() : null;
+    clone.fingerprint = fingerprint != null ? new ArrayList<>(fingerprint) : null;
+
+    if (breadcrumbs != null) {
+      Queue<Breadcrumb> breadcrumbsClone =
+          SynchronizedQueue.synchronizedQueue(new CircularFifoQueue<>(maxBreadcrumb));
```
bruno-garcia: Can we have this line as a private method? 
The intent is to have a single place where we define what concrete types are involved in creating breadcrumb list.
Right now it's here and in the constructor.

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -82,7 +86,54 @@ public void setExtra(String key, String value) {
...
+    }
+
+    if (extra != null) {
+      Map<String, Object> extraClone = new HashMap<>();
+
+      for (Map.Entry<String, Object> item : extra.entrySet()) {
+        if (item != null) {
+          extraClone.put(
+              item.getKey(),
+              item.getValue()); // TODO: how do we clone an object that we dont know the shape of it
```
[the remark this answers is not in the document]
bruno-garcia: The other SDKs don't deep copy at this point.
Here we can try to cast to Cloneable and if succeeds take the .clone() of it.
Data in Extra is expected to be serializable either way.

It's just important to have a new `List`, `Dictionary` so a call to `setExtra` on a new scope doesn't affect a previous one.

sentry-core/src/main/java/io/sentry/core/protocol/User.java
```diff
@@ -108,4 +109,46 @@ public void setOthers(Map<String, String> other) {
...
     this.unknown = unknown;
   }
+
+  @TestOnly
+  Map<String, Object> getUnknown() {
+    return unknown;
+  }
+
+  @Override
+  public User clone() throws CloneNotSupportedException {
```
bruno-garcia: can we have this implementation be on `<T> where T : IUnknownPropertiesConsumer` or something? Seems like the same code

sentry-core/src/main/java/io/sentry/core/Scope.java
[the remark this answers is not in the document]
marandaneto: @bruno-garcia `Object` has `clone()` as `protected` so won't be possible. Interface `Cloneable` doesn't have the `clone()` either, it's just an indicator.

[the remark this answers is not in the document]
marandaneto: will resolve it as new instances are already being created.

sentry-core/src/main/java/io/sentry/core/protocol/User.java
marandaneto: we are `Overriding` an `Object`'s method, it's just gonna make more complicated for something straightforward

PR #104, opened 2019-11-05
feat: Java ANR

bruno-garcia:
Added a ANR button to the sample and some support based on our 1.x SDK:

image

- [x] Write tests
- [ ] Fix thread marked as `crashed=true` which is not right considering the app wasn't killed:
- [x] Add AnrWatchDog license (see sentry-java)

image

Also, made me realize: Why are the package names not displayed? It's available in the JSON:

Wanted to open a draft PR but accidentally opened a final one.

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
@@ -30,6 +30,7 @@ public final class AnrIntegration implements Integration {
       anrWatchDog =
           new ANRWatchDog(
               options.getAnrTimeoutIntervalMills(),
+              options.isAnrReportInDebug(),
               error -> {
                 logIfNotNull(
                     options.getLogger(),
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
@@ -1,0 +1,52 @@
+package io.sentry.android.core;
+
+import static io.sentry.core.ILogger.logIfNotNull;
+
+import io.sentry.core.IHub;
+import io.sentry.core.Integration;
+import io.sentry.core.SentryLevel;
+import io.sentry.core.SentryOptions;
+import io.sentry.core.exception.ExceptionMechanismThrowable;
+import io.sentry.core.protocol.Mechanism;
+
+public final class AnrIntegration implements Integration {
+  private static ANRWatchDog anrWatchDog;
+
+  @Override
+  public void register(IHub hub, SentryOptions options) {
+    register(hub, (SentryAndroidOptions) options);
+  }
+
+  private void register(IHub hub, SentryAndroidOptions options) {
+    logIfNotNull(options.getLogger(), SentryLevel.DEBUG, "ANR enabled: %s", options.isAnrEnabled());
+
+    if (options.isAnrEnabled() && anrWatchDog == null) {
+      logIfNotNull(
+          options.getLogger(),
+          SentryLevel.DEBUG,
+          "ANR timeout in milliseconds: %d",
+          options.getAnrTimeoutIntervalMills());
+
+      anrWatchDog =
+          new ANRWatchDog(
+              options.getAnrTimeoutIntervalMills(),
+              error -> {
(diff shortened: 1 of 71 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -0,0 +1,33 @@
+package io.sentry.android.core;
+
+import io.sentry.core.SentryOptions;
+
+public final class SentryAndroidOptions extends SentryOptions {
+  private boolean anrEnabled = true; // TODO: Do we have ANR detection opt-out?
```
marandaneto: I'd say yes

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
```diff
@@ -15,6 +17,24 @@ private SentryAndroid() {}
...
+    init(context, new AndroidLogger());
+  }
+
+  static void init(@NotNull final Context context, @NotNull ILogger logger) {
+    try {
+      Sentry.init(
+          SentryAndroidOptions.class,
+          options -> AndroidOptionsInitializer.init(options, context, logger));
+    } catch (IllegalAccessException e) {
+      // This is awful. Should we have this all on the interface and let the caller deal with these?
```
marandaneto: Java has `Supplier` which helps on that, but only API 24

sentry-core/src/main/java/io/sentry/core/Sentry.java
```diff
@@ -29,13 +30,24 @@ public static void init() {
     init(new SentryOptions());
   }
 
-  public static void init(@NotNull OptionsConfiguration optionsConfiguration) {
+  // Used by integrations that define their own SentryOptions
+  public static <TOptions extends SentryOptions> void init(
+      @NotNull Class<TOptions> clazz, @NotNull OptionsConfiguration<TOptions> optionsConfiguration)
+      throws IllegalAccessException, InstantiationException, NoSuchMethodException,
+          InvocationTargetException {
+    TOptions options = clazz.getDeclaredConstructor().newInstance();
```
marandaneto: we don't need `getDeclaredConstructor()` most probably

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
marandaneto: Kotlin has `reified` btw.

quick question, why do we want to instantiate the options ourselves, would not be ok to allow them to instantiate their own `SentryAndroidOptions` and Sentry only get `<T extends SentryOptions>`? is there a reason to block it?

sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
```diff
@@ -0,0 +1,53 @@
...
+import io.sentry.core.SentryOptions;
+import io.sentry.core.exception.ExceptionMechanismThrowable;
+import io.sentry.core.protocol.Mechanism;
+
+public final class AnrIntegration implements Integration {
+  private static ANRWatchDog anrWatchDog;
+
+  @Override
+  public void register(IHub hub, SentryOptions options) {
+    register(hub, (SentryAndroidOptions) options);
```
marandaneto: that's ok, we could also make `Integration<T extends SentryOptions>` or don't we want that?

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
bruno-garcia: This was discussed offline. tl;dr; Allow passing the options instance to framework callbacks that will read configuration data into the object (i.e: env var, xml config files etc) and only after that, provide the instance to the user callback for mutation.

sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
bruno-garcia: The problem is that we keep them on a `List<Integration>` and each integration now would close the generic paramter with something different and would be a problem to have `List<Integration<TOptions>>` since they'd differ

PR #107, opened 2019-11-06
feat: Integration to send events on start

bruno-garcia:
Proposal for capturing events on startup as discussed this morning.

Blockers:
- [x] Merge the file watcher PR (#89) which has Serializer on Reader
- [x] Merge hint PR #106 to allow propagating the file was cached already
- [x] Merge event caching PR (#111) which writes files, contain the const with postfixes
- [x] Unit tests

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/EventCachedEventFireAndForgetIntegration.java
+++ b/sentry-core/src/main/java/io/sentry/core/EventCachedEventFireAndForgetIntegration.java
@@ -1,0 +1,125 @@
+package io.sentry.core;
+
+import static io.sentry.core.ILogger.logIfNotNull;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.charset.Charset;
+import java.util.concurrent.Executors;
+
+public final class EventCachedEventFireAndForgetIntegration implements Integration {
+  private static final Charset UTF_8 = Charset.forName("UTF-8");
+
+  @Override
+  public void register(IHub hub, SentryOptions options) {
+    String cachedDir = options.getCacheDirPath();
+    if (cachedDir == null) {
+      logIfNotNull(
+          options.getLogger(), SentryLevel.WARNING, "No cache dir path is defined in options.");
+      return;
+    }
+
+    File outbox = new File(options.getCacheDirPath());
+    try {
+      Executors.callable(
+              () -> {
+                SendCachedFiles(hub, options.getLogger(), options.getSerializer(), outbox);
+              })
+          .call();
+    } catch (Exception e) {
+      logIfNotNull(
+          options.getLogger(),
+          SentryLevel.ERROR,
+          "Failed trying to send cached events at %s",
+          outbox);
+    }
+  }
+
+  private static void SendCachedFiles(
+      IHub hub, ILogger logger, ISerializer serializer, File directory) {
+    if (!directory.exists()) {
+      logIfNotNull(
+          logger,
+          SentryLevel.INFO,
+          "Directory '%s' doesn't exist. Nothing no cached events to send.",
+          directory.getAbsolutePath());
+      return;
+    }
+    if (!directory.isDirectory()) {
+      logIfNotNull(
+          logger,
+          SentryLevel.ERROR,
+          "Cache dir %s is not a directory.",
+          directory.getAbsolutePath());
+      return;
+    }
+
+    logIfNotNull(
+        logger,
+        SentryLevel.DEBUG,
+        "Processing %d items from cache dir %s",
+        directory.length(),
+        directory.getAbsolutePath());
+
+    for (File file : directory.listFiles()) {
+      // TODO: postfix should ve a const, shared with the caching code (still to be merged)
+      if (!file.getName().endsWith(".sentry-event")) {
+        logIfNotNull(
+            logger,
+            SentryLevel.DEBUG,
+            "File '%s' doesn't match extension expected.",
+            file.getName());
+        continue;
+      }
+
+      if (!directory.isFile()) {
+        logIfNotNull(logger, SentryLevel.DEBUG, "'%s' is not a file.", directory.getAbsolutePath());
+        return;
+      }
+
+      if (!file.canRead()) {
+        logIfNotNull(logger, SentryLevel.WARNING, "File '%s' cannot be read.", file.getName());
+        safeDelete(file, "which can't be read but has expected file extension.", logger);
+        continue;
+      }
+
+      CachedEvent hint = new CachedEvent();
+      try (Reader reader =
+          new BufferedReader(new InputStreamReader(new FileInputStream(file), UTF_8))) {
+        SentryEvent event = serializer.deserializeEvent(reader);
+        // TODO: Run a sync capture? The SentryId returned could be the indication to whether delete
+        // or not.
+        // Hub doesn't throw so we wouldn't know if it was successful or not.
+        // Hint will be used by the caching layer to avoid persisting this event
+        hub.captureEvent(event /*, hint); // when hint PR is merged */);
+      } catch (FileNotFoundException e) {
+        logIfNotNull(logger, SentryLevel.ERROR, "File '%s' cannot be found.", file.getName(), e);
+      } catch (IOException e) {
+        logIfNotNull(logger, SentryLevel.ERROR, "I/O on file '%s' failed.", file.getName(), e);
+      } finally {
+        // Unless the transport marked this to be retried, it'll be deleted.
+        if (!hint.isResend()) {
+          safeDelete(file, "after trying to capture it", logger);
+        }
+      }
+    }
+  }
+
+  private static void safeDelete(File file, String errorMessageSuffix, ILogger logger) {
+    try {
+      file.delete();
+    } catch (Exception e) {
+      logIfNotNull(
+          logger,
+          SentryLevel.ERROR,
+          "Failed to delete '%s' " + errorMessageSuffix,
+          file.getName(),
+          e);
+    }
+  }
+}
--- a/sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
+++ b/sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
@@ -12,7 +12,6 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.nio.charset.Charset;
-
 import org.jetbrains.annotations.NotNull;
 
 final class SendCachedEvent {
--- a/sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
+++ b/sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
@@ -2,6 +2,7 @@ package io.sentry.core;
 
 import static io.sentry.core.ILogger.logIfNotNull;
 
+import io.sentry.core.cache.DiskCache;
 import io.sentry.core.util.Objects;
 import java.io.BufferedReader;
 import java.io.File;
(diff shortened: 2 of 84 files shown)


sentry-core/src/main/java/io/sentry/core/EventCachedEventFireAndForgetIntegration.java
```diff
@@ -0,0 +1,125 @@
...
+      if (!file.getName().endsWith(".sentry-event")) {
+        logIfNotNull(
+            logger,
+            SentryLevel.DEBUG,
+            "File '%s' doesn't match extension expected.",
+            file.getName());
+        continue;
+      }
+
+      if (!directory.isFile()) {
```
marandaneto: should it not be `file ` in here?

sentry-core/src/main/java/io/sentry/core/EventCachedEventFireAndForgetIntegration.java
```diff
@@ -0,0 +1,125 @@
...
+        hub.captureEvent(event /*, hint); // when hint PR is merged */);
+      } catch (FileNotFoundException e) {
+        logIfNotNull(logger, SentryLevel.ERROR, "File '%s' cannot be found.", file.getName(), e);
+      } catch (IOException e) {
+        logIfNotNull(logger, SentryLevel.ERROR, "I/O on file '%s' failed.", file.getName(), e);
+      } finally {
+        // Unless the transport marked this to be retried, it'll be deleted.
+        if (!hint.isResend()) {
+          safeDelete(file, "after trying to capture it", logger);
+        }
```
marandaneto: I'd do that out of the `finally` block as it's more business logic than closing resources

sentry-core/src/main/java/io/sentry/core/EventCachedEventFireAndForgetIntegration.java
```diff
@@ -0,0 +1,125 @@
...
+      logIfNotNull(
+          options.getLogger(), SentryLevel.WARNING, "No cache dir path is defined in options.");
+      return;
+    }
+
+    File outbox = new File(options.getCacheDirPath());
+    try {
+      Executors.callable(
+              () -> {
+                SendCachedFiles(hub, options.getLogger(), options.getSerializer(), outbox);
```
marandaneto: why don't you have a global variable caching what we need to use in this class in different methods, like `ILogger, ISerializer` and so on, just easier than passing them around, or just the `SentryOptions`, we are doing this way in the other integrations

sentry-core/src/main/java/io/sentry/core/EventCachedEventFireAndForgetIntegration.java
```diff
@@ -0,0 +1,125 @@
...
+            file.getName());
+        continue;
+      }
+
+      if (!directory.isFile()) {
+        logIfNotNull(logger, SentryLevel.DEBUG, "'%s' is not a file.", directory.getAbsolutePath());
+        return;
+      }
+
+      if (!file.canRead()) {
```
marandaneto: if you want to delete a file, the way how to check is `file.getParentFile().canWrite()`

sentry-core/src/main/java/io/sentry/core/EventCachedEventFireAndForgetIntegration.java
```diff
@@ -0,0 +1,125 @@
...
+        logIfNotNull(logger, SentryLevel.WARNING, "File '%s' cannot be read.", file.getName());
+        safeDelete(file, "which can't be read but has expected file extension.", logger);
+        continue;
+      }
+
+      CachedEvent hint = new CachedEvent();
+      try (Reader reader =
+          new BufferedReader(new InputStreamReader(new FileInputStream(file), UTF_8))) {
+        SentryEvent event = serializer.deserializeEvent(reader);
+        // TODO: Run a sync capture? The SentryId returned could be the indication to whether delete
```
marandaneto: I'd say only to enqueue it, as `EventSender` has `IEventCache` and it's business logic to `eventCache.discard(event)` if so, I'd say we fix the business logic in there if we need to.

sentry-core/src/main/java/io/sentry/core/EventCachedEventFireAndForgetIntegration.java
metlos: The problem with `SentryOptions` for me is that it is mutable - this enables logging to be changed at runtime, which is good, but I'm not sure swapping out the serializer (or cache if it were returned from the options) is something we want to support/enable. So while you can set the serializer on the options it doesn't affect the client that already exists - that client will always use the serializer that it was instantiated with. This eliminates problems like inability to deserialize an event after it was successfully serialized using a different serializer.

bruno-garcia: What @metlos said and also, I want to make it easier to pull the whole logic out of this class so it can be reused not only from this "FireAndForget" approach.

sentry-core/src/main/java/io/sentry/core/EventCachedEventFireAndForgetIntegration.java
bruno-garcia: Good point. We do need to read it too. But I guess if we can't delete it, maybe we ignore the file otherwise we would be sending this file on each app start.

sentry-core/src/main/java/io/sentry/core/EventCachedEventFireAndForgetIntegration.java
bruno-garcia: I need to bypass that logic. Before it discards, it stores and we dont' want this file to be stored again.

sentry-core/src/main/java/io/sentry/core/EventCachedEventFireAndForgetIntegration.java
bruno-garcia: Good catch!

sentry-core/src/main/java/io/sentry/core/EventCachedEventFireAndForgetIntegration.java
bruno-garcia: We need to make sure a file that we fail processing gets deleted.

sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    logIfNotNull(
+        logger,
+        SentryLevel.DEBUG,
+        "Processing %d items from cache dir %s",
+        directory.length(),
+        directory.getAbsolutePath());
+
+    for (File file : directory.listFiles()) {
```
marandaneto: instead of the `if` below, you may use `listFiles(FilenameFilter filter)` which also returns `File[]`

sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
```diff
@@ -0,0 +1,115 @@
...
+        logIfNotNull(logger, SentryLevel.ERROR, "File '%s' cannot be found.", file.getName(), e);
+      } catch (IOException e) {
+        logIfNotNull(logger, SentryLevel.ERROR, "I/O on file '%s' failed.", file.getName(), e);
+      } catch (Exception e) {
+        logIfNotNull(
+            logger, SentryLevel.ERROR, "Failed to capture cached event.", file.getName(), e);
+        hint.setResend(false);
+      } finally {
+        // Unless the transport marked this to be retried, it'll be deleted.
+        if (!hint.isResend()) {
```
marandaneto: I'd prefer to have it out of the `finally` block.
`finally` for java at least, it's more common to be used for cleaning resources.

also, there's this thing:

> Note: If the JVM exits while the try or catch code is being executed, then the finally block may not execute. Likewise, if the thread executing the try or catch code is interrupted or killed, the finally block may not execute even though the application as a whole continues.

sentry-core/src/test/java/io/sentry/core/SendCachedEventTest.kt
```diff
@@ -0,0 +1,99 @@
...
+      options.isDebug = true
+      options.setLogger(logger)
+    }
+
+    fun getSut(): SendCachedEvent {
+      return SendCachedEvent(serializer, hub, logger)
+    }
+  }
+
+    val tempDirectory = Files.createTempDirectory("send-cached-event-test")
```
marandaneto: I'd declare a `File` as `lateinit`, init it using `@BeforeTest` and delete it using `@AfterTest`, so you have control if the `File` exists or not in your tests

sentry-core/src/test/java/io/sentry/core/SendCachedEventTest.kt
```diff
@@ -0,0 +1,99 @@
...
+import java.nio.file.Files
+import kotlin.test.Test
+import kotlin.test.assertFalse
+
+class SendCachedEventTest {
+  private class Fixture {
+    var hub: IHub? = mock()
+      var logger: ILogger? = mock()
+      var serializer: ISerializer? = mock()
+    var options = SentryOptions()
```
marandaneto: why are all of them nullable? `type?` if they get a fresh instance with `mock()`

sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
bruno-garcia: Right, we can't guarantee finally will be called. But I do want this code to be executed either if it comes out of try or catch, and that's what `finally` is there for. I'm not sure why not use it?
I do want to 'clean up' a resource. If we don't get rid of this file, it'll be sent forever.

sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
```diff
@@ -0,0 +1,115 @@
...
+import org.jetbrains.annotations.TestOnly;
+
+final class SendCachedEvent {
+  private static final Charset UTF_8 = Charset.forName("UTF-8");
+  @TestOnly static final String CACHED_EVENT_SUFFIX = ".sentry-event";
+  private final ISerializer serializer;
+  private final IHub hub;
+  private final ILogger logger;
+
+  SendCachedEvent(ISerializer serializer, IHub hub, ILogger logger) {
```
marandaneto: I'd also use the jetbrains annotations `@NotNull`

sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
```diff
@@ -0,0 +1,115 @@
...
+      } finally {
+        // Unless the transport marked this to be retried, it'll be deleted.
+        if (!hint.isResend()) {
+          safeDelete(file, "after trying to capture it", logger);
+        }
+      }
+    }
+  }
+
+  private static void safeDelete(File file, String errorMessageSuffix, ILogger logger) {
```
marandaneto: why is it `static`?

sentry-core/src/test/java/io/sentry/core/SendCachedEventTest.kt
bruno-garcia: Often I set each one null and write a test to see how the mock behaves during construction time

sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
bruno-garcia: Doesn't need access to the objects state. Would there be a reason to make it a member instance?

sentry-core/src/test/java/io/sentry/core/SendCachedEventTest.kt
bruno-garcia: Didn't do it now because I'm rushing everything*

sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
bruno-garcia: I started with the filter but thought of writing logs instead. The directory is expected to be used solely by Sentry's SDK event caching. But if it's somehow mixed up with other stuff, would be nice to know when debugging something is weird.

marandaneto: that makes total sense, agreed

PR #110, opened 2019-11-06
closing integrations, so it wont leak observers or handlers

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
closing integrations, so it won't leak file observers or handlers

integrations were leaking observers and/or handlers

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
+++ b/sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
@@ -87,7 +87,8 @@ public final class UncaughtExceptionHandlerIntegration
 
   @Override
   public void close() {
-    if (defaultExceptionHandler != null) {
+    if (defaultExceptionHandler != null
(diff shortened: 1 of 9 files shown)


sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
```diff
@@ -83,4 +84,11 @@ static Throwable getUnhandledThrowable(Thread thread, Throwable thrown) {
     mechanism.setType("UncaughtExceptionHandler");
     return new ExceptionMechanismThrowable(mechanism, thrown, thread);
   }
+
+  @Override
+  public void close() {
+    if (defaultExceptionHandler != null) {
+      threadAdapter.setDefaultUncaughtExceptionHandler(defaultExceptionHandler);
```
bruno-garcia: This is problematic. We need to check if we are still the current handler before we go ahead and override the current one.
After we set ourselves there, another integration might have set themselves and wrapped us.

PR #111, opened 2019-11-07
Implementation of DiskCache and the ability to store events before sending

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
This adds the implementation of the `IEventCache` interface that stores the data to the filesystem.
Also it adds a simple ability of the `AsyncConnection` to store the events before attempting to send.
The implementation of the `DiskCache` is largely copied from `sentry-java`.

While a new config property governing the maximum of cached events is added in this PR, the user has no choice in selecting the implementation of the cache. Maybe this should be changed?

We need this to be able to persist the events before sending to enable crash/kill recovery.

Just unit tests, no testing on a real/simulated Android device

## :pencil: Checklist

- [x] All tests passing

## :crystal_ball: Next steps
- [ ] add ability to set the `IEventCache` implementation in `SentryOptions` as it is done for `ISerializer`?
- [ ] is sharing the threadpool for sending and storing events the right thing to do?

by Manoel:
Continuation of https://github.com/getsentry/sentry-android/pull/98

### DIFF
--- a/sentry-android-core/src/test/java/io/sentry/android/core/EnvelopeFileObserverIntegrationTest.kt
+++ b/sentry-android-core/src/test/java/io/sentry/android/core/EnvelopeFileObserverIntegrationTest.kt
@@ -7,12 +7,28 @@ import com.nhaarman.mockitokotlin2.verify
 import io.sentry.core.Hub
 import io.sentry.core.SentryOptions
 import java.io.File
+import java.nio.file.Files
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import org.junit.runner.RunWith
 
 @RunWith(AndroidJUnit4::class)
 class EnvelopeFileObserverIntegrationTest {
+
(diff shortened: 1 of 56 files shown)


sentry-android-core/src/test/java/io/sentry/android/core/EnvelopeFileObserverIntegrationTest.kt
```diff
@@ -7,12 +7,28 @@ import com.nhaarman.mockitokotlin2.verify
...
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import org.junit.runner.RunWith
 
 @RunWith(AndroidJUnit4::class)
 class EnvelopeFileObserverIntegrationTest {
+
+    private lateinit var file: File
```
ninniuz: Would have been easier with a `TemporaryFolder` JUnit rule

PR #117, opened 2019-11-08
added errored field and fixes

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
added errored field and fixes

we need an `errored` to mark as the current thread but necessarily crashed.
stacktraces were out of order.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/SentryStackTraceFactory.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryStackTraceFactory.java
@@ -2,6 +2,7 @@ package io.sentry.core;
 
 import io.sentry.core.protocol.SentryStackFrame;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import org.jetbrains.annotations.Nullable;
 
@@ -45,6 +46,7 @@ final class SentryStackTraceFactory {
         }
       }
     }
+    Collections.reverse(sentryStackFrames);
 
     return sentryStackFrames;
   }
--- a/sentry-core/src/main/java/io/sentry/core/SentryThreadFactory.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryThreadFactory.java
@@ -53,7 +53,9 @@ final class SentryThreadFactory {
     sentryThread.setDaemon(thread.isDaemon());
     sentryThread.setState(thread.getState().name());
     if (crashedThreadId != null) {
+      // TODO: do we still need crashed?
       sentryThread.setCrashed(crashedThreadId == thread.getId());
+      sentryThread.setErrored(sentryThread.isCrashed());
     }
     sentryThread.setCurrent(thread == currentThread);
 
--- a/sentry-core/src/main/java/io/sentry/core/SentryThreadFactory.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryThreadFactory.java
@@ -56,8 +56,6 @@ final class SentryThreadFactory {
       sentryThread.setCrashed(crashedThreadId == thread.getId());
     }
     sentryThread.setErrored(thread == currentThread);
-    //    sentryThread.setCurrent(thread == currentThread); TODO: do we need that? this should be if
-    // its the UI thread or not
 
     List<SentryStackFrame> frames = sentryStackTraceFactory.getStackFrames(stackFramesElements);
 
--- a/sentry-core/src/main/java/io/sentry/core/SentryThreadFactory.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryThreadFactory.java
@@ -32,9 +32,9 @@ final class SentryThreadFactory {
     Map<Thread, StackTraceElement[]> threads = Thread.getAllStackTraces();
     List<SentryThread> result = new ArrayList<>();
 
+    Thread currentThread = Thread.currentThread();
     for (Map.Entry<Thread, StackTraceElement[]> item : threads.entrySet()) {
-      result.add(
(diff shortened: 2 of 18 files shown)


sentry-core/src/main/java/io/sentry/core/SentryThreadFactory.java
```diff
@@ -53,7 +53,9 @@ private SentryThread getSentryThread(
     sentryThread.setDaemon(thread.isDaemon());
     sentryThread.setState(thread.getState().name());
     if (crashedThreadId != null) {
+      // TODO: do we still need crashed?
       sentryThread.setCrashed(crashedThreadId == thread.getId());
+      sentryThread.setErrored(sentryThread.isCrashed());
```
marandaneto: is it here the right place to mark? or the `else` case? it's a bit confusing as we have `crashed, daemon, current and errored`

sentry-core/src/test/java/io/sentry/core/SentryExceptionFactoryTest.kt
```diff
@@ -52,6 +52,7 @@ class SentryExceptionFactoryTest {
 
         assertEquals("message", queue.first.value)
         assertEquals("cause", queue.last.value)
+        // TODO: is it right? or it should be the cause(1) and then the exception(2)?
```
marandaneto: @bruno-garcia do you know how it should be the right order of exceptions?

sentry-core/src/main/java/io/sentry/core/SentryStackTraceFactory.java
```diff
@@ -45,6 +46,7 @@ public SentryStackTraceFactory(
         }
       }
     }
+    Collections.reverse(sentryStackFrames);
```
marandaneto: Jan said that stack traces were out of order.

sentry-core/src/main/java/io/sentry/core/SentryThreadFactory.java
```diff
@@ -55,7 +55,9 @@ private SentryThread getSentryThread(
     if (crashedThreadId != null) {
       sentryThread.setCrashed(crashedThreadId == thread.getId());
     }
-    sentryThread.setCurrent(thread == currentThread);
+    sentryThread.setErrored(thread == currentThread);
+    //    sentryThread.setCurrent(thread == currentThread); TODO: do we need that? this should be if
```
marandaneto: @bruno-garcia `current` should be if this thread is `foreground` or not, so we will depend on the Android FW.

`Looper.getMainLooper().getThread() == Thread.currentThread()` means `current` = true if I got it right.

sentry-sample/src/main/AndroidManifest.xml
```diff
@@ -34,7 +34,7 @@
     <meta-data android:name="io.sentry.anr.timeout-interval-mills" android:value="2000" />
 
     <!-- Raise ANR events even if the debugger is attached -->
-    <meta-data android:name="io.sentry.anr.report-debug" android:value="true" />
+    <meta-data android:name="io.sentry.anr.report-debug" android:value="false" />
```
marandaneto: this is annoying if we are debugging and have breakpoints, it should be disabled by default.

sentry-core/src/main/java/io/sentry/core/SentryThreadFactory.java
bruno-garcia: That means UI threat. Foreground would be isDaemon=false.
Could be useful to also flag the UI threat on mobile I assume

sentry-sample/src/main/AndroidManifest.xml
bruno-garcia: Can also just comment out the whole line (it's disabled by default) e, leave it with a comment (to cause ANR while debugging)

sentry-core/src/test/java/io/sentry/core/SentryExceptionFactoryTest.kt
bruno-garcia: It's reversed from the logical order cuz python

sentry-core/src/main/java/io/sentry/core/SentryThreadFactory.java
bruno-garcia: Sorry I'm talking about isDaemon which has nothing to do with current. Definition of current is muddy imo. By looking at the name I think of the thread where the error happened. Hence the code. If an exception is thrown in a background there. I'd Mark that one as current and show that on the UI.
I believe we're going for 'erroed' to display in the UI which is fine

marandaneto: @bruno-garcia `Foreground would be isDaemon=false` is not correct for Android, that's why we need to compare the `Looper.getMainLooper().getThread()` for Android.

sentry-sample/src/main/AndroidManifest.xml
marandaneto: you are right, I will comment then and keep as it was, so one will know how to enable it.

sentry-core/src/main/java/io/sentry/core/SentryThreadFactory.java
bruno-garcia: I guess the concept doesn't make much sense then since it's a thread that avoid the process from ending. Doesn't make much sense in mobile anyway.on Windows you can keep a process running if u have a foreground thread alive even if the UI thread exited. I assume that doesn't happen in mobile

PR #119, opened 2019-11-08
setting correct names to package and native attributes (stack frames)

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
setting correct names to package and native attributes (stack frames)

package and native are reserved words in java, but we need them like this when serialized/deserialized.

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
getters and setters have the correct name, I will write an adapter later on, but let's fix it like this for now because it won't be a breaking change.

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/protocol/SentryStackFrame.java
+++ b/sentry-core/src/main/java/io/sentry/core/protocol/SentryStackFrame.java
@@ -1,5 +1,6 @@
 package io.sentry.core.protocol;
 
+import com.google.gson.annotations.SerializedName;
 import io.sentry.core.IUnknownPropertiesConsumer;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


sentry-core/src/main/java/io/sentry/core/protocol/SentryStackFrame.java
```diff
@@ -18,8 +19,13 @@
   private String absPath;
   private String contextLine;
   private Boolean inApp;
-  private String _package; // TODO: _package as its a reserverd word
-  private Boolean _native; // TODO: _native as its a reserverd word
+
+  @SerializedName(value = "package")
```
marandaneto: @bruno-garcia the solution here is not ideal, an adapter would be better, but I'd like to fix all the open bugs asap.
later could be replaced with an adapter without a breaking change.

bruno-garcia: Wfm

PR #121, opened 2019-11-09
feat: publish to bintray

bruno-garcia:
While doing this I found we have to publish snapshot builds to artifactory instead.

The goal of this PR is to be able to publish packages via the main gradle file which will publish jar for sentry-core and arr for the Android libraries. It'll also ignore the sample project.

### DIFF
--- a/sentry-core/build.gradle.kts
+++ b/sentry-core/build.gradle.kts
@@ -1,8 +1,15 @@
+import com.jfrog.bintray.gradle.tasks.BintrayPublishTask
+import com.jfrog.bintray.gradle.tasks.BintrayUploadTask
+import org.gradle.api.publish.PublishingExtension
+
 plugins {
     `java-library`
+    `maven-publish`
+    signing
     kotlin("jvm")
     jacoco
     id("net.ltgt.errorprone")
+    id(Config.PublishPlugins.bintrayPlugin)
 }
 
 dependencies {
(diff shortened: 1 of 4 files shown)


sentry-core/build.gradle.kts
```diff
@@ -1,8 +1,15 @@
+import com.jfrog.bintray.gradle.tasks.BintrayPublishTask
+import com.jfrog.bintray.gradle.tasks.BintrayUploadTask
+import org.gradle.api.publish.PublishingExtension
+
 plugins {
     `java-library`
+    `maven-publish`
```
bruno-garcia: @marandaneto I guess merging #70 will affect this bit

marandaneto: not a problem

sentry-core/build.gradle.kts
```diff
@@ -49,4 +56,60 @@ tasks {
...
+            setPublications(project.name)
+        }
+    }
+}
+
+publishing {
+    publications {
+        register(project.name, MavenPublication::class) {
+            if (project.hasProperty("android")) {
+                artifact("$buildDir/outputs/aar/${project.name}-release.aar")
```
marandaneto: this is the sentry-core module, we don't need android stuff in here

sentry-sample/build.gradle.kts
```diff
@@ -3,7 +3,7 @@ import org.jetbrains.kotlin.config.KotlinCompilerVersion
 plugins {
     id("com.android.application")
     kotlin("android")
-//    id("io.sentry.android.gradle") how to add sentry gradle plugin
+    id(Config.PublishPlugins.bintrayPlugin)
```
marandaneto: this should go into the modules that would be published, not in the sample app.

sentry-core/build.gradle.kts
bruno-garcia: Right, this is all very provisional. It will all move to the main gradle file

sentry-sample/build.gradle.kts
bruno-garcia: right, just playing around. Draft PR

bruno-garcia: I'll add a description of the goal

bruno-garcia:
Closed in favor of #129

PR #124, opened 2019-11-10
feat: Don't minify exception types

bruno-garcia:
Automatically include a rule to avoid minification of exceptions

See https://github.com/getsentry/sentry-java/issues/711#issuecomment-550443950

### DIFF
--- a/sentry-android-core/proguard-rules.pro
+++ b/sentry-android-core/proguard-rules.pro
@@ -24,6 +24,9 @@
 -keep class * implements com.google.gson.JsonSerializer
 -keep class * implements com.google.gson.JsonDeserializer
 
+# Prevent proguard from minifying exception type names
+-keep public class * extends java.lang.Exception
+
 # Prevent R8 from leaving Data object members always null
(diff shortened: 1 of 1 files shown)


sentry-android-core/proguard-rules.pro
```diff
@@ -24,6 +24,9 @@
 -keep class * implements com.google.gson.JsonSerializer
 -keep class * implements com.google.gson.JsonDeserializer
 
+# Prevent proguard from minifying exception type names
+-keep public class * extends java.lang.Exception
```
marandaneto: what's about
`-keep class * extends java.lang.RuntimeException` as well?
so we cover checked and unchecked exceptions.

I'd also remove `public` as it could be an internal exception right

bruno-garcia: Makes sense.

bruno-garcia:
```suggestion
-keep class * extends java.lang.Exception
```

PR #125, opened 2019-11-10
exception classes extend RuntimeException not Throwable

marandaneto:


sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
```diff
@@ -41,8 +41,8 @@ private void register(IHub hub, SentryAndroidOptions options) {
                 Mechanism mechanism = new Mechanism();
                 mechanism.setType("ANR");
                 mechanism.setHandled(false);
-                ExceptionMechanismThrowable throwable =
-                    new ExceptionMechanismThrowable(mechanism, error, Thread.currentThread());
+                ExceptionMechanism throwable =
```
bruno-garcia: We have it in the protocol the `ExceptionMechanism` which is called `Mechanism`. In C# it's actually called `ExceptionMechanism` so this name is a bit confusing.

Maybe MechanismException if this is in fact an `Exception`?

sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
```diff
@@ -82,7 +82,7 @@ static Throwable getUnhandledThrowable(Thread thread, Throwable thrown) {
     Mechanism mechanism = new Mechanism();
     mechanism.setHandled(false);
     mechanism.setType("UncaughtExceptionHandler");
-    return new ExceptionMechanismThrowable(mechanism, thrown, thread);
+    return new ExceptionMechanism(mechanism, thrown, thread);
```
bruno-garcia: This part became confusing

https://docs.sentry.io/development/sdk-dev/event-payloads/exception/#exception-mechanism

sentry-core/src/main/java/io/sentry/core/exception/ExceptionMechanism.java
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * A {@link Throwable} that decorates another with a Sentry {@link Mechanism}.
+   *
+   * @param mechanism The {@link Mechanism}.
+   * @param throwable The {@link java.lang.Throwable}.
+   * @param thread The {@link java.lang.Thread}.
+   */
+  public ExceptionMechanism(
+      @Nullable Mechanism mechanism, @Nullable Throwable throwable, @Nullable Thread thread) {
```
bruno-garcia: Before this was a throwable wrapping a throwable.
Now this is an exception that wraps a throwable.
Would we want this to wrap an exception instead?
```suggestion
      @Nullable Mechanism mechanism, @Nullable Exception exception, @Nullable Thread thread) {
```

Also, considering this is a decorator, we could name it `MechanismDecoratorException` or something along these lines?

bruno-garcia: This file was added? What happened with the original `ExceptionMechanismThrowable`? Could it be a rename operation (delete the original file)?

sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
marandaneto: ExceptionMechanismException now, so same pattern as before

sentry-core/src/main/java/io/sentry/core/exception/ExceptionMechanism.java
marandaneto: guess the IDE did something wrong, I just renamed it but fixed it.

PR #135, opened 2019-11-11
fix: Capture envelope when native is done writing

bruno-garcia:


sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java
```diff
@@ -13,7 +13,13 @@
 
   private static native void initSentryNative(SentryOptions options);
 
+  private static native void verificationEventNative();
```
marandaneto: this is only for testing, right?

bruno-garcia: Yes. Also, often on the onboarding in Sentry there's a `verification step` which is often tricky and requires telling the user a bunch of code to write. 
Perhaps we could facilitate by having a method that will create such event.

But for now let's bring this one in for testing

PR #136, opened 2019-11-12
ref: use hints

bruno-garcia:
Use `Hint` to opt-in or out to some SDK behavior to better support scenarios like sending cached events/envelopes, anr, uncaught exception


sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
```diff
@@ -75,7 +77,7 @@ public void sendCachedFiles(@NotNull File directory) {
         continue;
       }
 
-      CachedEvent hint = new CachedEvent();
+      SendCachedEventHint hint = new SendCachedEventHint();
```
marandaneto: you may use `Retryable`

sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
```diff
@@ -92,4 +99,27 @@ public void close() {
...
+
+    UncaughtExceptionHint(long timeoutMills) {
+      this.timeoutMills = timeoutMills;
+      this.latch = new CountDownLatch(1);
+    }
+
+    public void waitFlush() {
+      try {
+        latch.await(timeoutMills, TimeUnit.MILLISECONDS);
+      } catch (InterruptedException e) {
```
marandaneto: we may log the error

sentry-core/src/main/java/io/sentry/core/transport/AsyncConnection.java
```diff
@@ -159,17 +167,25 @@ public void run() {
...
-              options.getLogger().log(SentryLevel.DEBUG, message);
+              options.getLogger().log(SentryLevel.ERROR, message);
             }
 
             throw new IllegalStateException(message);
           }
         } catch (IOException e) {
           eventCache.store(event);
+          // Failure due to IO is allowed to retry the event
+          if (hint instanceof io.sentry.core.hints.Retryable) {
```
marandaneto: we already have a `Retryable` object, let's merge it but keep in mind to change this latter.

sentry-core/src/main/java/io/sentry/core/transport/AsyncConnection.java
```diff
@@ -159,17 +167,25 @@ public void run() {
...
+            ((io.sentry.core.hints.Retryable) hint).setRetry(true);
+          }
           throw new IllegalStateException("Sending the event failed.", e);
         }
       } else {
         eventCache.store(event);
+        // If transportGate is blocking from sending, allowed to retry
+        if (hint instanceof io.sentry.core.hints.Retryable) {
```
marandaneto: same here

PR #138, opened 2019-11-12
ref: uncaught ex

bruno-garcia:


sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
```diff
@@ -20,20 +20,21 @@ public void register(@NotNull IHub hub, @NotNull SentryOptions options) {
...
+            } catch (Exception e) {
+              logIfNotNull(
+                  options.getLogger(),
+                  SentryLevel.ERROR,
+                  "Failed trying to send cached events at %s",
+                  e,
+                  outbox);
+            }
+          });
+      es.shutdown();
```
bruno-garcia: @marandaneto @metlos calling shutdown here will **always** wait for the call above to complete? Even if a thread wasn't schedule yet at the time shutdown was called?

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
```diff
@@ -20,20 +20,21 @@ public void register(@NotNull IHub hub, @NotNull SentryOptions options) {
...
-                  logIfNotNull(
-                      options.getLogger(),
-                      SentryLevel.ERROR,
-                      "Failed trying to send cached events at %s",
-                      e,
-                      outbox);
-                }
-              })
-          .call();
+      ExecutorService es = Executors.newFixedThreadPool(1);
```
marandaneto: you may use `newSingleThreadExecutor` which is almost the same, but if a failure happens, a new thread takes place

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
```diff
@@ -20,20 +20,21 @@ public void register(@NotNull IHub hub, @NotNull SentryOptions options) {
...
+      es.execute(
```
marandaneto: I think `submit(...)` would be the correct one

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
marandaneto: > method is called on an executor service, it stops accepting new tasks, waits for previously submitted tasks to execute, and then terminates the executor.

I'd say yes, as the `submit()` will be sync. and the execution async.

marandaneto: its also common to combine `shutdown()` and `awaitTermination`

metlos: `ExecutorService#shutdown()` initiates an "orderly" shutdown of the executor where it stops accepting new tasks but lets the enqueued tasks finish. But the call merely **initiates** the shutdown and doesn't wait for it to be complete. That's what `awaitTermination` is for as pointed out by @marandaneto .

marandaneto: as it's a fire and forget class, only `shutdown()` is ok

bruno-garcia: This is the `FireAndForget` thingy so I don't want to awaitTermination.
I just want it to use a background thread-from a pool or not, I don't care where it runs.
I also don't want to leak anything hence the `shutdown` call but after reading the docs, the `initiates an orderly... accepts more` I wasn't clear if once `sumbit` returns it's considered accepted. Or if it still asyncrhonously does something that will lead to it being accepted. And hance having a race between `submit and shutdown`.

bruno-garcia: Just read Manoel's reply! Thanks

PR #140, opened 2019-11-12
feat: Unknown frames are not in app

bruno-garcia:
Since we add the app's namespace as `inAppInclude`, I propose default all frames as InApp=false.

Sentry should anyway show the whole stacktrace if *all frames* are inApp=false so if we somehow missed it, it would show the whole thing.
The default should be good enough as in only showing the app's stack frames and hiding everything else. If no app frames are included, everything is shown.

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/SentryStackTraceFactory.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryStackTraceFactory.java
@@ -4,7 +4,7 @@ import io.sentry.core.protocol.SentryStackFrame;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.*;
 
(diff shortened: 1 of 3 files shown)


sentry-core/src/main/java/io/sentry/core/SentryStackTraceFactory.java
```diff
@@ -4,7 +4,7 @@
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.*;
```
marandaneto: why? :(

bruno-garcia: My FUCKING IDE

bruno-garcia:
```suggestion
import org.jetbrains.annotations.Nullable;
import org.jetbrains.annotations.TestOnly;
```

PR #141, opened 2019-11-12
fix: inapp

bruno-garcia:
No frame was marked as inApp because inAppInclude list was null and passed as an argument to the factory before the Android package could add (first the instance of the list) the inApp frames

image


bruno-garcia:
@marandaneto we need some integration tests eh? That `init` the SDK and verify some behavior of the thing stitched together

marandaneto:
> @marandaneto we need some integration tests eh? That `init` the SDK and verify some behavior of the thing stitched together

yeah, that would be good, let's merge it for now.

PR #142, opened 2019-11-13
ref: Sample ANR

bruno-garcia:
Fire it in a shorter interval and block for a bit longer to fire every time

### DIFF
--- a/sentry-sample/src/main/AndroidManifest.xml
+++ b/sentry-sample/src/main/AndroidManifest.xml
@@ -30,8 +30,8 @@
 
     <meta-data android:name="io.sentry.debug" android:value="true" />
 
-    <!-- Raise a ANR event if UI thread blocked for over 2 seconds -->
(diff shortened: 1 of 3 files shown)


sentry-sample/src/main/AndroidManifest.xml
```diff
@@ -30,8 +30,8 @@
 
     <meta-data android:name="io.sentry.debug" android:value="true" />
 
-    <!-- Raise a ANR event if UI thread blocked for over 2 seconds -->
-    <meta-data android:name="io.sentry.anr.timeout-interval-mills" android:value="2000" />
+    <!-- To change the time used to watch for ANR. By default it's 5 seconds (5000 below as it's in milliseconds) -->
```
marandaneto: @bruno-garcia 5s the App. is gonna die and nothing is gonna happen.
I'd suggest like 4s or 4.5s

marandaneto: I've changed ANR default's value to 4s

PR #143, opened 2019-11-13
fix: clone scope

bruno-garcia:


sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -408,7 +408,16 @@ public IHub clone() {
     // Clone will be invoked in parallel
     Hub clone = new Hub(this.options, null);
     for (StackItem item : this.stack) {
-      clone.stack.push(item);
+      Scope clonedScope = null;
+      try {
+        clonedScope = item.scope.clone();
+      } catch (CloneNotSupportedException e) {
+        // TODO: Why do we need to deal with this? We must guarantee clone is possible here!
```
marandaneto: because the docs say:

>  * Note that this interface does <i>not</i> contain the <tt>clone</tt> method.
>  * Therefore, it is not possible to clone an object merely by virtue of the
>  * fact that it implements this interface.  Even if the clone method is invoked
>  * reflectively, there is no guarantee that it will succeed.

previously I've made it into `Scope.clone` class and you've said to blow up and take care about it in here, so we could log the error.

bruno-garcia: Then on a second though: We should handle the errors at the individual levels of the scope and do a partial clone then, I believe. It's better than losing all the scope data because 1 object added by the user of the API can't be cloned.

What I meant in the TODO was for us to make sure we can clone our own objects, so there shouldn't be the possibility of `CloneNotSupportedException`

PR #145, opened 2019-11-13
wip

bruno-garcia:
Still needs the hints and surely a refactor but first stab at sending envelopes on start (signal handler)

Event written by `sentry-native` is there, fails to read but succeeds deleting (?!)

> java.io.FileNotFoundException: /data/user/0/io.sentry.sample/cache/sentry/outbox/17f84131-5044-48de-7b83-a1e71e77f1a3: open failed: EACCES (Permission denied)

Mainly need to figure this out before reviewing and merging. Following a refactor.

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/DirectoryProcessor.java
+++ b/sentry-core/src/main/java/io/sentry/core/DirectoryProcessor.java
@@ -1,0 +1,58 @@
+package io.sentry.core;
+
+import static io.sentry.core.ILogger.logIfNotNull;
+
+import java.io.File;
+import org.jetbrains.annotations.Nullable;
+
+public abstract class DirectoryProcessor {
+
+  private ILogger logger;
+
+  protected DirectoryProcessor(@Nullable ILogger logger) {
+
+    this.logger = logger;
+  }
+
+  void processDirectory(File directory) {
+    if (!directory.exists()) {
+      logIfNotNull(
+          logger,
+          SentryLevel.WARNING,
+          "Directory '%s' doesn't exist. No cached events to send.",
+          directory.getAbsolutePath());
+      return;
+    }
+    if (!directory.isDirectory()) {
+      logIfNotNull(
+          logger,
+          SentryLevel.ERROR,
+          "Cache dir %s is not a directory.",
+          directory.getAbsolutePath());
+      return;
+    }
+
+    File[] listFiles = directory.listFiles();
+    if (listFiles == null) {
+      logIfNotNull(logger, SentryLevel.ERROR, "Cache dir %s is null.", directory.getAbsolutePath());
+      return;
+    }
+
+    File[] filteredListFiles = directory.listFiles((d, name) -> isRelevantFileName(name));
+
+    logIfNotNull(
+        logger,
+        SentryLevel.DEBUG,
+        "Processing %d items from cache dir %s",
+        filteredListFiles != null ? filteredListFiles.length : 0,
+        directory.getAbsolutePath());
+
+    for (File file : listFiles) {
+      processFile(file);
+    }
+  }
+
+  protected abstract void processFile(File file);
+
+  protected abstract boolean isRelevantFileName(String fileName);
+}
--- a/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
+++ b/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
@@ -11,7 +11,7 @@ import java.io.Reader;
 import java.nio.charset.Charset;
 import org.jetbrains.annotations.NotNull;
 
-public final class EnvelopeSender implements IEnvelopeSender {
+public final class EnvelopeSender extends DirectoryProcessor implements IEnvelopeSender {
 
   @SuppressWarnings("CharsetObjectCanBeUsed")
   private static final Charset UTF_8 = Charset.forName("UTF-8");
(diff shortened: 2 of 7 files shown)


bruno-garcia:
@marandaneto Got it out of Draft. Not ready yet but we should start discussing this.
I already want to do some refactor but at this point I just want to get this working properly, write tests to verify the behavior and then refactor.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidSerializer.java
```diff
@@ -44,7 +44,7 @@ private Gson provideGson() {
         .registerTypeAdapter(SentryLevel.class, new SentryLevelSerializerAdapter(logger))
         .registerTypeAdapter(SentryLevel.class, new SentryLevelDeserializerAdapter(logger))
         .registerTypeAdapter(Contexts.class, new ContextsDeserializerAdapter(logger))
-        .registerTypeAdapterFactory(UnknownPropertiesTypeAdapterFactory.get())
+//        .registerTypeAdapterFactory(UnknownPropertiesTypeAdapterFactory.get())
```
bruno-garcia:
```suggestion
        .registerTypeAdapterFactory(UnknownPropertiesTypeAdapterFactory.get())
```

sentry-core/src/main/java/io/sentry/core/DirectoryProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+import static io.sentry.core.ILogger.logIfNotNull;
+
+import java.io.File;
+import org.jetbrains.annotations.Nullable;
+
+public abstract class DirectoryProcessor {
+
+  private ILogger logger;
+
+  protected DirectoryProcessor(@Nullable ILogger logger) {
```
marandaneto: we have `NoOpLogger` so it won't be null right

sentry-core/src/main/java/io/sentry/core/DirectoryProcessor.java
```diff
@@ -0,0 +1,58 @@
+package io.sentry.core;
+
+import static io.sentry.core.ILogger.logIfNotNull;
+
+import java.io.File;
+import org.jetbrains.annotations.Nullable;
+
+public abstract class DirectoryProcessor {
```
marandaneto: can it be private package?

sentry-core/src/main/java/io/sentry/core/DirectoryProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private ILogger logger;
+
+  protected DirectoryProcessor(@Nullable ILogger logger) {
+
+    this.logger = logger;
+  }
+
+  void processDirectory(File directory) {
```
marandaneto: add @NotNull

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
```diff
@@ -1,5 +1,6 @@
 package io.sentry.core;
 
+import io.sentry.core.hints.*;
```
marandaneto: The last few commits were always with wildcards, not good, I'd fix it.

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
```diff
@@ -99,6 +120,63 @@ private void processEnvelope(SentryEnvelope envelope) throws IOException {
...
+          logger, ERROR, "Exception while awaiting for flush in UncaughtExceptionHint", e);
+      }
+    }
+
+    public void newCountDown() {
+      latch = new CountDownLatch(1);
+    }
+
+    @Override
+    public void flushed() {
```
marandaneto: is it `flushed` a good name? what's about only `flush`?

sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
```diff
@@ -16,98 +16,63 @@
...
+    super(logger);
     this.serializer = Objects.requireNonNull(serializer, "Serializer is required.");
     this.hub = Objects.requireNonNull(hub, "Hub is required.");
     this.logger = Objects.requireNonNull(logger, "Logger is required.");
   }
 
-  public void sendCachedFiles(@NotNull File directory) {
-    if (!directory.exists()) {
+  @Override
+  protected void processFile(File file) {
```
marandaneto: @NotNull annotation

sentry-core/src/main/java/io/sentry/core/DirectoryProcessor.java
bruno-garcia: fixed: `NotNull`

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
bruno-garcia: I've tried to change this multiple times.
Based on this:
https://intellij-support.jetbrains.com/hc/en-us/community/posts/206203659-Turn-off-Wildcard-imports-

It's here:

image
It's always adding it now though

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
bruno-garcia: `flush` seems like the action of flushing.
I need `markFlushed` maybe a better name then

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
ninniuz: I usually set these to 999 to avoid star imports

bruno-garcia: It's adding it for a single imports for me. And no change to that setting makes any difference.
Nothing on `.editorconfig`, no idea what's going on

marandaneto: I think we should force it on `.editorconfig`, maybe nothing means default value which is just a few start imports

PR #146, opened 2019-11-13
fixed ANR typo

marandaneto:


sentry-sample/src/main/java/io/sentry/sample/MainActivity.java
```diff
@@ -65,7 +65,7 @@ protected void onCreate(Bundle savedInstanceState) {
         .setOnClickListener(
             view -> {
               // Try cause ANR by blocking for 2.5 seconds.
-              // By default the SDK sends an event if blocked by at least 5 seconds.
+              // By default the SDK sends an event if blocked by at least 4 seconds.
```
bruno-garcia: Old SDK is 5. Should we move to 4?
Or this is because the config is currently4?

marandaneto: I've changed the config to 4s, but forgot to change the comment.
the reason is, newer android versions may crash after 5s ANR, so we won't have the time to capture it.
would be nice to have this buffer of 1s to actually do something and report as non-fatal.

PR #148, opened 2019-11-13
fix: envelope header event id

bruno-garcia:


sentry-core/src/main/java/io/sentry/core/protocol/SentryId.java
```diff
@@ -27,6 +28,19 @@ public String toString() {
     return uuid.toString().replace("-", "");
   }
 
+  @Override
+  public boolean equals(Object o) {
```
marandaneto: I think you can use `compareTo(UUID val)` and you don't need to override anything

bruno-garcia: I'll use it internally to compare the UUIDs, but `SentryId` I belive should override it. One might accidentally use equals here (as I did). I'd say it's probably the only type worth overriding so far.

Will do as you suggested: `return uuid.compareTo(sentryId.uuid) == 0;`

PR #149, opened 2019-11-14
WIP - fire and forget envelopes

marandaneto:
findings of debugging the current impl.

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/UnknownPropertiesTypeAdapterFactory.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/UnknownPropertiesTypeAdapterFactory.java
@@ -104,8 +104,10 @@ final class UnknownPropertiesTypeAdapterFactory implements TypeAdapterFactory {
       }
       // First convert the above JSON tree to an object
       final T object = typeAdapter.fromJsonTree(jsonObjectToParse);
-      // And do the post-processing
-      object.acceptUnknownProperties(unknownProperties);
+      if (!unknownProperties.isEmpty()) {
+        // And do the post-processing
+        object.acceptUnknownProperties(unknownProperties);
+      }
       return object;
     }
 
--- a/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
+++ b/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
@@ -111,7 +111,7 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
               continue;
             }
             hub.captureEvent(event, hint);
-            hint.waitFlush();
+            hint.waitFlush(); // why do we need to waitFlush? is not ok to be async?
             hint.newCountDown();
             logger.log(SentryLevel.DEBUG, "Item %d is being captured.", items);
           }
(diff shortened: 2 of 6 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/UnknownPropertiesTypeAdapterFactory.java
```diff
@@ -104,8 +104,10 @@ public T read(final JsonReader in) {
       }
       // First convert the above JSON tree to an object
       final T object = typeAdapter.fromJsonTree(jsonObjectToParse);
-      // And do the post-processing
-      object.acceptUnknownProperties(unknownProperties);
+      if (!unknownProperties.isEmpty()) {
```
marandaneto: @bruno-garcia this will eliminate the `unknown` field error for now as we have all the fields already mapped on our data classes, but deserialization still should be fixed.

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
```diff
@@ -111,7 +111,7 @@ private void processEnvelope(SentryEnvelope envelope, CachedEnvelopeHint hint)
               continue;
             }
             hub.captureEvent(event, hint);
-            hint.waitFlush();
+            hint.waitFlush(); // why do we need to waitFlush? is not ok to be async?
```
marandaneto: @bruno-garcia why don't we just enqueue the events and the rest of the flow keeps the same?

sentry-core/src/main/java/io/sentry/core/transport/AsyncConnection.java
```diff
@@ -176,7 +176,8 @@ private void flush() {
             throw new IllegalStateException(message);
           }
         } catch (IOException e) {
-          eventCache.store(event);
+          //          eventCache.store(event); TODO: its already saving at first, do we still need
```
marandaneto: @bruno-garcia I believe this should be gone as we already save at the beginning.

sentry-core/src/main/java/io/sentry/core/protocol/Mechanism.java
```diff
@@ -12,6 +12,7 @@
   private Map<String, Object> meta;
   private Map<String, Object> data;
   private final transient Thread thread;
+  private Boolean synthetic;
```
marandaneto: https://github.com/getsentry/sentry-native/pull/89#discussion_r346324745

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -61,11 +61,13 @@ public SentryId captureEvent(SentryEvent event, @Nullable Scope scope, @Nullable
 
     logIfNotNull(options.getLogger(), SentryLevel.DEBUG, "Capturing event: %s", event.getEventId());
 
+    // TODO: CachedEnvelopeHint is an instance of Cached, so its not applying processors at all,
+    // probably a bug only for hard crash?
```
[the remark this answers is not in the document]
bruno-garcia: We a can't apply everything. As discussed offline.
It went from broken to broken.

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
bruno-garcia: We need to know if it succeeded or not to delete the file.
If it fails due to the device being offline we don't want to delete it.
Since this is a background thread anyway, it's fine to sync with the transport

bruno-garcia:
I've brought ur changes into the branch: https://github.com/getsentry/sentry-android/pull/145

sentry-core/src/main/java/io/sentry/core/SentryClient.java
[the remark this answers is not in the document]
marandaneto: didn't get it, let's discuss offline

PR #153, opened 2019-11-15
fix proguard

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
added more exceptions to proguard

NDK cant be init. on release mode

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
move all classes to protocol package, so I dont need to do:
-keep class io.sentry.core.** { *; }
-keep class io.sentry.core.protocol.** { *; }

but only:
-keep class io.sentry.core.protocol.** { *; }

### DIFF
--- a/sentry-android-core/proguard-rules.pro
+++ b/sentry-android-core/proguard-rules.pro
@@ -12,9 +12,10 @@
 #-keep class com.google.gson.stream.** { *; }
 
 # Application classes that will be serialized/deserialized over Gson
--keep class io.sentry.core.** { <fields>; }
--keep class io.sentry.core.protocol.** { <fields>; }
+-keep class io.sentry.core.** { *; }
+-keep class io.sentry.core.protocol.** { *; }
 -keepclassmembers enum * { *; }
+-keep class io.sentry.android.core.** { *; }
 
 # Prevent proguard from stripping interface information from TypeAdapter, TypeAdapterFactory,
 # JsonSerializer, JsonDeserializer instances (so they can be used in @JsonAdapter)
--- a/sentry-android-ndk/proguard-rules.pro
+++ b/sentry-android-ndk/proguard-rules.pro
@@ -1,10 +1,16 @@
 ##---------------Begin: proguard configuration for NDK  ----------
 
--keep class io.sentry.ndk.** { <fields>; }
+-keep class io.sentry.android.ndk.** { *; }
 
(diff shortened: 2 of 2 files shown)


sentry-android-core/proguard-rules.pro
```diff
@@ -12,9 +12,10 @@
 #-keep class com.google.gson.stream.** { *; }
 
 # Application classes that will be serialized/deserialized over Gson
--keep class io.sentry.core.** { <fields>; }
--keep class io.sentry.core.protocol.** { <fields>; }
+-keep class io.sentry.core.** { *; }
```
marandaneto: @bruno-garcia I'd like to remove this exception, so lib may be optimized, but for that, we need to move some data classes from io.sentry.core to io.sentry.core.protocol eg Breadcrumb

sentry-android-core/proguard-rules.pro
```diff
@@ -12,9 +12,10 @@
...
+-keep class io.sentry.core.protocol.** { *; }
 -keepclassmembers enum * { *; }
+-keep class io.sentry.android.core.** { *; }
```
marandaneto: here may be only `SentryAndroidOptions`

sentry-android-ndk/proguard-rules.pro
```diff
@@ -1,10 +1,16 @@
...
 
--keep class io.sentry.ndk.** { <fields>; }
+-keep class io.sentry.android.ndk.** { *; }
 
 # For native methods, see http://proguard.sourceforge.net/manual/examples.html#native
 -keepclasseswithmembernames,includedescriptorclasses class * {
     native <methods>;
 }
 
+# if issues with proguard, disable it
```
marandaneto: can be removed, but its only a comment anyway

sentry-android-core/proguard-rules.pro
bruno-garcia: Can we just add the exact classes instead of moving them?
I mean since you need to import each individual class in Java, we can move them, doesn't matter.
Some types stayed in the top level namespace in .NET to to available without adding a bunch of `import`

marandaneto: jep, also all the public classes should be there.
lets merge it for now

PR #157, opened 2019-11-18
add method to captureMessage with level

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
add method to captureMessage with level

captureMessage didn't support level yet.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/ISentryClient.java
+++ b/sentry-core/src/main/java/io/sentry/core/ISentryClient.java
@@ -30,10 +30,7 @@ public interface ISentryClient {
     Message sentryMessage = new Message();
     sentryMessage.setFormatted(message);
     event.setMessage(sentryMessage);
-
-    if (scope == null || scope.getLevel() == null) {
-      event.setLevel(level);
-    }
+    event.setLevel(level);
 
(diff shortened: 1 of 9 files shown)


sentry-core/src/main/java/io/sentry/core/ISentryClient.java
```diff
@@ -25,18 +25,23 @@ default SentryId captureEvent(SentryEvent event, @Nullable Object hint) {
...
-  }
-
-  default SentryId captureMessage(String message, @Nullable Scope scope) {
+  default SentryId captureMessage(String message, SentryLevel level, @Nullable Scope scope) {
     SentryEvent event = new SentryEvent();
     Message sentryMessage = new Message();
     sentryMessage.setFormatted(message);
     event.setMessage(sentryMessage);
+
+    if (scope == null || scope.getLevel() == null) {
```
bruno-garcia: The default should be just dumb helper to what a user would do which is create an event and set the level.
The scope dealing is stuff happening under the hoods (MainEventProcessor) and if we were to change that decision (scope > event for levels) we would change in 1 place,

PR #160, opened 2019-11-21
execute beforeBreadcrumb also for scope

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
execute beforeBreadcrumb also for scope

beforeBreadcrumb was not being executed when breadcrumbs were added directly to the scope

unit tests

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
refactor it, we have a design flaw here, I don't like the solution, but we need to have it done till tomorrow.

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Scope.java
+++ b/sentry-core/src/main/java/io/sentry/core/Scope.java
@@ -1,6 +1,8 @@
 package io.sentry.core;
 
 import io.sentry.core.protocol.User;
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -62,6 +71,33 @@ public void setFingerprint(List<String> fingerprint) {
...
 
   public void addBreadcrumb(@NotNull Breadcrumb breadcrumb) {
+    addBreadcrumb(breadcrumb, true);
+  }
+
+  void addBreadcrumb(@NotNull Breadcrumb breadcrumb, boolean executeBeforeBreadcrumb) {
+    if (executeBeforeBreadcrumb && beforeBreadcrumbCallback != null) {
+      try {
+        breadcrumb =
+            beforeBreadcrumbCallback.execute(breadcrumb, null); // TODO: whats about hint here?
```
bruno-garcia: We should take hint also

marandaneto: @bruno-garcia how? if the user has called `scope.addBreadcrumb(...)` where do we get it from?

bruno-garcia: We'd need to add an overload that takes hint.

PR #161, opened 2019-11-22
adding nativeBundle plugin

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
adding 3rd plugin that copies headers and extracts fat aar.

one will need to:

Apply the plugin in the App:

`apply plugin: 'com.ydq.android.gradle.native-aar.import'`

Reference the extracted binaries in their `CMakeLists.txt`:

```
include (${ANDROID_GRADLE_NATIVE_BUNDLE_PLUGIN_MK})
target_link_libraries(myapp ${ANDROID_GRADLE_NATIVE_MODULES})
```

Include the header in their c files:

`#include <sentry.h>`

it's not possible to include sentry.h from native code at the moment, that will make it possible.

deploying to maven local and including sentry.h and capturing an issue.

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
create a sample of how to include the 3rd plugin and include the headers


sentry-android-ndk/build.gradle.kts
```diff
@@ -102,6 +105,10 @@ android {
     }
 
     ndkVersion = "20.1.5948944"
+
+    nativeBundleExport {
+        headerDir = "${project.projectDir}/sentry-native/include"
```
marandaneto: what will make the `aar` to have the header `jni/include/sentry.h`

marandaneto: if you use sym link, this might be changed to `/sentry-native-local/` btw

sentry-sample/src/main/cpp/native-sample.cpp
```diff
@@ -1,7 +1,8 @@
 #include <jni.h>
 #include <android/log.h>
+//#include <sentry.h>
```
marandaneto: fails on compile time, sentry.h not found

build.gradle.kts
```diff
@@ -24,6 +24,7 @@ buildscript {
         classpath(kotlin(Config.BuildPlugins.kotlinGradlePlugin, version = Config.kotlinVersion))
         classpath(Config.QualityPlugins.errorpronePlugin)
         classpath(Config.Deploy.novodaBintray)
+        classpath("com.ydq.android.gradle.build.tool:nativeBundle:1.0.3")
```
marandaneto: plugin to export c headers

bruno-garcia:
@marandaneto can we try on the sample? to consume the bits?

marandaneto:
> @marandaneto can we try on the sample? to consume the bits?

the sample won't work because it's referencing the project directly and not a maven dependency.
this will only work with "maven dependencies", that's why it exists.

for this sample, as they are under the same project, we would need to reference the project in the CMakeLists.txt manually as we do on sentry-android-ndk/sentry-native.
we basically need to copy the behavior, using a relative path.
See: https://github.com/howardpang/androidNativeBundle/issues/1

bruno-garcia:
is it possible to run a maven install on local maven on build, and on the sample consume from maven local (as opposed to a reference to a project?)

marandaneto:
> is it possible to run a maven install on local maven on build, and on the sample consume from maven local (as opposed to a reference to a project?)

yes, I thought to do that, but then we lose debugging.
The tooling is shit :(

PR #164, opened 2019-11-23
import native libs statically

marandaneto:
just a test, the idea is to test the native-shared libs


marandaneto:
in favor of https://github.com/getsentry/sentry-android/pull/161

PR #179, opened 2019-12-03
adding scope methods to sentry static class

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
adding scope methods to sentry static class

making migration easier, otherwise one will need to call `Sentry.configureScope(...)`

I'll do unit tests

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Hub.java
+++ b/sentry-core/src/main/java/io/sentry/core/Hub.java
@@ -31,7 +31,7 @@ public final class Hub implements IHub {
   private volatile @NotNull SentryId lastEventId;
   private final @NotNull SentryOptions options;
   private volatile boolean isEnabled;
-  private final Deque<StackItem> stack = new LinkedBlockingDeque<>();
+  private final @NotNull Deque<StackItem> stack = new LinkedBlockingDeque<>();
 
   public Hub(@NotNull SentryOptions options) {
     this(options, createRootStackItem(options));
@@ -227,7 +227,7 @@ public final class Hub implements IHub {
   }
 
   @Override
-  public void setLevel(SentryLevel level) {
+  public void setLevel(@Nullable SentryLevel level) {
     if (!isEnabled()) {
       logIfNotNull(
           options.getLogger(),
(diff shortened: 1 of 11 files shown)


bruno-garcia:
I'd like to record my thinking here for @fdrabek and @HazAT .
Adding these method to `Sentry.` is not ideal. The reason is, now there's even more ways to do simple stuff, but this is more unclear how it works.

```
Sentry.setLevel();
Sentry.pushScope();
Sentry.setLevel();
```

It looks even more like magic. What's happening under the hood that the ambient data propagates down?

Also, not **all** methods are on `Sentry.` which means we still need `configureScope`. This makes it extra confusing. If we're going this direction, I'd say we add everything to `Sentry.` and dump `configureScope` all together (which is not my favorite option anyway).

The argument to do this in JS was that passing a cloujure was "too advance" for our users. That one never made much sense.
But the real motive was to make it easier to port from old users. One on in Java has requested that yet. We could simply provide a migration guide.

There was no `pushScope`. There's anyway a new concept. We need to introduce users to the new concept anyway.

Instead of poluting the API I believe we should figure out a way to get this to work with `async` frameworks because `Sentry.setLevel` will not work at all. Neither is our current scope management based on thread-local-storage.

HazAT:
@bruno-garcia 

> But the real motive was to make it easier to port from old users. One on in Java has requested that yet. We could simply provide a migration guide.

True, no one requested it yet but the number of people potentially not updating because they can't simply string replace 
```java
eventBuilder.withExtra(entry.getKey(), entry.getValue());
// with
Sentry.setExtra(entry.getKey(), entry.getValue());
```
is something that's hard to measure. Just one person writing could mean, a 100 users didn't even bother to write an issue and another 1000 users complety discarded the idea of upgrading.

I agree with your side of adding more API is generally always bad I think we need to take the hit in order to make it simpler, even it's just for a few users.

Also, those functions were so people don't need to even understand what a `scope` is. We don't want to force anyone learning how powerful a `scope` can be, if they have a problem they can't solve with a simple `setTag`, then it's time they move to learn about scopes and not earlier.


All things aside, I think it would make sense to have a general discussion about the unified API, specifically talking about what we've learned in the last 1-2 years about it. It's for sure we would make some things a bit different.

PR #183, opened 2019-12-04
SDK final review

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
final review of the public API.

review access modifiers, package structure, etc..

## :pencil: Checklist

- [X] All tests passing


sentry-core/src/main/java/io/sentry/core/cache/DiskCache.java
```diff
@@ -23,12 +23,14 @@
...
 import java.util.Iterator;
 import java.util.List;
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
 /**
  * A simple cache implementation storing the events to a disk, each event in a separater file in a
  * configured directory.
  */
+@ApiStatus.Internal // TODO: check if its only internal
```
marandaneto: @bruno-garcia can it be used by DI as well?

sentry-core/src/main/java/io/sentry/core/transport/AsyncConnection.java
```diff
@@ -15,11 +15,13 @@
 import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.TestOnly;
 
 /** A connection to Sentry that sends the events asynchronously. */
+@ApiStatus.Internal // TODO: check if its only internal
```
marandaneto: @bruno-garcia can it be used by DI as well?

bruno-garcia: The Hub ctor takes SentryOptions so for now lets keep the rest of the API internal so we can move things around a bit.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -13,13 +13,15 @@
...
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
 
 /**
  * An implementation of the {@link ITransport} interface that sends the events to the Sentry server
  * over HTTP(S) in UTF-8 encoding.
  */
-@Open // TODO: make it final and disable nopen check for testing
+@Open
+@ApiStatus.NonExtendable // only not final because of testing
```
bruno-garcia: Could the tests rely on `ITransport` and make this class final?

marandaneto: not actually, because it needs to extend and override the `open` method otherwise `Connection refused` will be thrown on every test

marandaneto: the other way would be to create a fake/mock http server, I've used for okhttp, but not for `HttpURLConnection`, will have a look.

PR #185, opened 2019-12-05
workaround android_id

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
generating android_id if Settings.Secure.ANDROID_ID is buggy or non existent.

https://android-developers.googleblog.com/2011/03/identifying-app-installations.html

debugging all the cases, this class has not been tested yet (unit tests), will do it later on.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -785,7 +785,23 @@ public final class DefaultAndroidEventProcessor implements EventProcessor {
   private String getAndroidId() {
     // Android 29 has changed and -> Avoid using hardware identifiers, find another way in the
     // future
-    return Settings.Secure.getString(context.getContentResolver(), Settings.Secure.ANDROID_ID);
+    String androidId =
+        Settings.Secure.getString(context.getContentResolver(), Settings.Secure.ANDROID_ID);
+
+    //    https://android-developers.googleblog.com/2011/03/identifying-app-installations.html
+    if (androidId == null
+        || androidId.isEmpty()
+        || androidId.toLowerCase(Locale.ROOT).contentEquals("9774d56d682e549c")) {
+      try {
+        androidId = Installation.id(context);
+      } catch (RuntimeException e) {
+        log(SentryLevel.ERROR, "Could not generate device Id.", e);
+
+        androidId = UUID.randomUUID().toString();
+      }
+    }
+
+    return androidId;
   }
 
   private String[] getProGuardUuids() {
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -797,7 +797,7 @@ public final class DefaultAndroidEventProcessor implements EventProcessor {
       } catch (RuntimeException e) {
         log(SentryLevel.ERROR, "Could not generate device Id.", e);
 
-        androidId = UUID.randomUUID().toString();
+        return null;
       }
     }
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/Installation.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/Installation.java
@@ -1,0 +1,43 @@
+package io.sentry.android.core;
+
+import android.content.Context;
+import java.io.File;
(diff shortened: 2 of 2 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -785,7 +785,23 @@ public User getUser() {
...
+    //    https://android-developers.googleblog.com/2011/03/identifying-app-installations.html
+    if (androidId == null
+        || androidId.isEmpty()
+        || androidId.toLowerCase(Locale.ROOT).contentEquals("9774d56d682e549c")) {
+      try {
+        androidId = Installation.id(context);
+      } catch (RuntimeException e) {
+        log(SentryLevel.ERROR, "Could not generate device Id.", e);
+
+        androidId = UUID.randomUUID().toString();
```
bruno-garcia: Would it make sense to return `null` here instead? We'll potentially artificially increase the cardinality of the device id without any added benefit to the client. I'll look like there're more users then in fact are and cost us more to process the event.

sentry-android-core/src/main/java/io/sentry/android/core/Installation.java
```diff
@@ -0,0 +1,45 @@
...
+    RandomAccessFile f = new RandomAccessFile(installation, "r");
+    byte[] bytes = new byte[(int) f.length()];
+    f.readFully(bytes);
+    f.close();
+    return new String(bytes, UTF_8);
+  }
+
+  private static void writeInstallationFile(File installation) throws IOException {
+    FileOutputStream out = new FileOutputStream(installation);
+    String id = UUID.randomUUID().toString();
```
bruno-garcia: I'd suggest returning this string to avoid having the read I/O done right after creating the file

sentry-android-core/src/main/java/io/sentry/android/core/Installation.java
```diff
@@ -0,0 +1,45 @@
...
+  private static final Charset UTF_8 = Charset.forName("UTF-8");
+
+  private Installation() {}
+
+  public static synchronized String id(Context context) {
+    if (sID == null) {
+      File installation = new File(context.getFilesDir(), INSTALLATION);
+      try {
+        if (!installation.exists()) writeInstallationFile(installation);
+        sID = readInstallationFile(installation);
```
bruno-garcia: The id can be returned from the `writeInstallationFile` instead of doing I/O here

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -785,7 +785,23 @@ public User getUser() {
...
-    return Settings.Secure.getString(context.getContentResolver(), Settings.Secure.ANDROID_ID);
+    String androidId =
+        Settings.Secure.getString(context.getContentResolver(), Settings.Secure.ANDROID_ID);
+
+    //    https://android-developers.googleblog.com/2011/03/identifying-app-installations.html
+    if (androidId == null
+        || androidId.isEmpty()
+        || androidId.toLowerCase(Locale.ROOT).contentEquals("9774d56d682e549c")) {
+      try {
+        androidId = Installation.id(context);
```
bruno-garcia: This code is now potentially doing blocking I/O on the calling thread (which might be the UI thread).

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: yeah agreed, just thought it would make sense to have a sort of ID

sentry-android-core/src/main/java/io/sentry/android/core/Installation.java
marandaneto: good catch

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: yeah, it is, I'm gonna fix it together with all the other IO calls later on.

marandaneto: btw this will only be executed for a very small portion of the devices, the bug happens only for a specific brand and API version.

PR #193, opened 2019-12-09
adding rules to multidex

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
https://developer.android.com/studio/build/multidex
as we have 2 classes which are init. by reflection, let's add to multiindex rules and hopefully prevent some errors.

honestly, this is just a try on how to avoid problems with multidex, not sure if it's gonna be an issue at all, but 2 liners won't hurt.

I didn't test it, but so far we didn't experience an error either, see Description/Motivation.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/build.gradle.kts
+++ b/sentry-android-core/build.gradle.kts
@@ -35,6 +35,7 @@ android {
         getByName("debug")
         getByName("release") {
             consumerProguardFiles("proguard-rules.pro")
+            multiDexKeepProguard = file("multidex-config.pro")
         }
     }
 
--- a/sentry-android-core/multidex-config.pro
+++ b/sentry-android-core/multidex-config.pro
@@ -1,0 +1,5 @@
+##---------------Begin: multidex configuration for NDK  ----------
+
+-keep class io.sentry.android.core.SentryAndroidOptions
+
+##---------------End: multidex configuration for NDK  ----------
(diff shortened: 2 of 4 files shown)


ninniuz:
Are those rules exported together with the `consumer-proguard` rules? If they are not, I doubt they will have any effect on the consuming app.

You might want, instead, to add this info to the integration docs for applications that use Sentry SDK/NDK IMHO

marandaneto:
@ninniuz it’s the very same mechanism, look at https://developer.android.com/studio/build/multidex#keep so AGP does it automatically like consumer proguard

ninniuz:
It's just I do not seem to find any reference of this mechanism in the docs and never tried myself. 
If those rules are found in the output Sentry SDK `aar` then I am with you, it is likely that AGP is able to collect such rules from dependencies and use them when building app dex files, otherwise I don't think a consuming app will be able to use your multidex rules.

marandaneto:
> It's just I do not seem to find any reference of this mechanism in the docs and never tried myself.
> If those rules are found in the output Sentry SDK `aar` then I am with you, it is likely that AGP is able to collect such rules from dependencies and use them when building app dex files, otherwise I don't think a consuming app will be able to use your multidex rules.

that's actually true and I've not used as well, I've just remembered the use case and found the way on how we should solve it, I'll check if it actually embedded a file like proguard.txt at the final library, if the case is only for Apps., then we should add it to the docs for sure, thanks again, always very helpful @ninniuz

marandaneto:
it turns out that AGP doesn't merge this file

PR #200, opened 2019-12-12
throw an exception if DSN is not set

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
throw an exception if DSN is not set.

if DSN is not set, user needs to know right away, it might be a misconfiguration.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
@@ -56,7 +56,7 @@ final class ManifestMetadataReader {
 
         String dsn = metadata.getString(DSN_KEY, null);
         if (dsn == null) {
-          logIfNotNull(options.getLogger(), SentryLevel.FATAL, "DSN is required.");
+          logIfNotNull(options.getLogger(), SentryLevel.FATAL, "DSN is required. Use empty string to disable SDK.");
         } else if (dsn.isEmpty()) {
           logIfNotNull(
               options.getLogger(), SentryLevel.DEBUG, "DSN is empty, disabling sentry-android");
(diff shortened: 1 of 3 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -55,10 +55,15 @@ static void applyMetadata(Context context, SentryAndroidOptions options) {
         options.setAnrTimeoutIntervalMills(anrTimeoutIntervalMills);
 
         String dsn = metadata.getString(DSN_KEY, null);
-        if (dsn != null) {
+        if (dsn == null) {
+          logIfNotNull(options.getLogger(), SentryLevel.FATAL, "DSN is required.");
```
bruno-garcia: A suggestion:

```suggestion
          logIfNotNull(options.getLogger(), SentryLevel.FATAL, "DSN is required. Use empty string to disable SDK.");
```

sentry-core/src/main/java/io/sentry/core/Sentry.java
```diff
@@ -50,7 +50,9 @@ public static void init(@NotNull OptionsConfiguration<SentryOptions> optionsConf
 
   private static synchronized <T extends SentryOptions> void init(@NotNull T options) {
     String dsn = options.getDsn();
-    if (dsn == null || dsn.isEmpty()) {
+    if (dsn == null) {
+      throw new IllegalArgumentException("DSN is required");
```
marandaneto:
```suggestion
      throw new IllegalArgumentException("DSN is required. Use empty string to disable SDK.");
```

PR #201, opened 2019-12-12
added executor for caching values out of the main thread

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
added executor for caching values out of the main thread.

right now every capture exception or message will do IO on the main thread and data is static, so it may be cached.

executing and StrictMode.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -74,7 +74,7 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
     this.options = Objects.requireNonNull(options, "The SentryOptions is required.");
 
     ExecutorService executorService = Executors.newSingleThreadExecutor();
-    contextData = executorService.submit(this::loadContextData);
+    contextData = executorService.submit(() -> loadContextData());
 
     executorService.shutdown();
   }
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -59,6 +59,12 @@ import org.jetbrains.annotations.Nullable;
 
 final class DefaultAndroidEventProcessor implements EventProcessor {
 
+  private static final String PROGUARD_UUID = "proGuardUuids";
+  private static final String ROOTED = "rooted";
+  private static final String ANDROID_ID = "androidId";
+  private static final String KERNEL_VERSION = "kernelVersion";
+  private static final String EMULATOR = "emulator";
+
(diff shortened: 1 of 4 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -59,12 +64,43 @@
...
+
   public DefaultAndroidEventProcessor(Context context, SentryOptions options) {
     this.context =
         Objects.requireNonNull(
             context != null ? context.getApplicationContext() : null,
             "The application context is required.");
     this.options = Objects.requireNonNull(options, "The SentryOptions is required.");
+
+    ExecutorService executorService = Executors.newSingleThreadExecutor();
+    contextData = executorService.submit(() -> loadContextData());
```
marandaneto: could do a method reference here, but ran into a bug on lint :( https://groups.google.com/forum/#!topic/lint-dev/wf2QAqPvSs8

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -59,12 +64,43 @@
...
+    String[] proGuardUuids = getProGuardUuids();
+    if (proGuardUuids != null) {
+      map.put("proGuardUuids", proGuardUuids);
+    }
+
+    map.put("rooted", isRooted());
+
+    String androidId = getAndroidId();
+    if (androidId != null) {
+      map.put("androidId", androidId);
```
bruno-garcia: What about having the map keys as constants? To get compiler checks here and make sure they don't go out of sync?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -794,7 +859,15 @@ private String getApplicationName() {
   public User getUser() {
     User user = new User();
 
-    user.setId(getAndroidId());
+    try {
+      Object androidId = contextData.get().get("androidId");
+
+      if (androidId != null) {
+        user.setId((String) androidId);
```
bruno-garcia: We've been discussing adding this also to `context.device.id` which is not in the protocol yet but could be align the approach with iOS  since neither have a consistent way to give an id to the device.
Not necessarily for this PR though. /cc @fetzig

marandaneto:
> This is great stuff!

file storage stats are still IO, we could do it async, but we will need to do await anyway, not sure if its worth it.
file storage stats would need to be checked every event anyway as it could be the source of the exception.

marandaneto:
> This is great stuff!

file storage stats are still IO, we could do it async, but we will need to do await anyway, not sure if its worth it.
file storage stats would need to be checked every event anyway as it could be the source of the exception.

PR #203, opened 2020-01-02
transport may be set on options

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
transport may be set on options.

transport should be possible to be set on options as well, unified API.

unit tests.

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
socket tagging because of https://github.com/square/okhttp/issues/3537
but it's just a warning and an okhttp issue, not ours, so not sure if we should do this workaround only for the sake of warning free. but why? I agree with https://github.com/coil-kt/coil/issues/46


sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -35,10 +36,16 @@ public boolean isEnabled() {
     this(options, null);
   }
 
-  @TestOnly
   public SentryClient(SentryOptions options, @Nullable Connection connection) {
     this.options = options;
     this.enabled = true;
+
+    ITransport transport = options.getTransport();
+    if (transport instanceof NoOpHttpTransport) {
```
bruno-garcia: Could we just leave it `null` instead of having `NoOp` assigned to options?

sentry-core/src/main/java/io/sentry/core/SentryOptions.java
```diff
@@ -228,6 +231,14 @@ public void addInAppInclude(@NotNull String include) {
     inAppIncludes.add(include);
   }
 
+  public @NotNull ITransport getTransport() {
+    return transport;
```
bruno-garcia: Callers would need to be aware of `NoOp` as a type if they'd want to see what's assigned here (can't infer `null` means nothing was set yet)

marandaneto: yeah, it makes sense.

PR #206, opened 2020-01-03
added android transport gate

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
added an android transport gate to check if the device is online or not, so the event only will try to send if it actually has a connected network.

check if the device has a connected network before trying to send the event, right now even if it's not connected anywhere, we are trying to send, so wasting resources.

unit tests on the setup, but it's hard to mock the whole network conf. of the android framework.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/util/CheckConnectivity.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/util/CheckConnectivity.java
@@ -1,0 +1,54 @@
+package io.sentry.android.core.util;
+
+import android.Manifest;
+import android.content.Context;
+import android.net.ConnectivityManager;
+import io.sentry.core.ILogger;
+import io.sentry.core.SentryLevel;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@ApiStatus.Internal
+public final class CheckConnectivity {
+
+  private CheckConnectivity() {}
+
+  /**
+   * Check whether the application has internet access at a point in time.
+   *
+   * @return true if the application has internet access
+   */
+  public static @Nullable Boolean isConnected(@NotNull Context context, @NotNull ILogger logger) {
+    ConnectivityManager connectivityManager =
+        (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+    if (connectivityManager == null) {
+      logger.log(SentryLevel.INFO, "ConnectivityManager is null and cannot check network status");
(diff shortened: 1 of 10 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/util/CheckConnectivity.java
```diff
@@ -0,0 +1,54 @@
...
+import io.sentry.core.ILogger;
+import io.sentry.core.SentryLevel;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@ApiStatus.Internal
+public final class CheckConnectivity {
+
+  private CheckConnectivity() {}
```
bruno-garcia: What about we name the class like:
```suggestion
  private ConectivityChecker() {}
```

Or `ConectivitySomethingElse` since `CheckConectivity` sounds like an action or method

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -45,8 +46,13 @@ public SentryClient(SentryOptions options, @Nullable Connection connection) {
       options.setTransport(transport);
     }
 
-    if (connection == null) {
+    ITransportGate transportGate = options.getTransportGate();
+    if (transportGate == null) {
+      transportGate = () -> true;
+      options.setTransportGate(transportGate);
```
bruno-garcia: Can we get away with setting this? All user defined callbacks have triggered so sounds a bit late.
We can simply keep `transportGate` as a field and ignore `options.getTransportGate` within this type.

sentry-android-core/src/main/java/io/sentry/android/core/util/CheckConnectivity.java
marandaneto: yeah, it makes sense, sometimes I feel writing extension methods hehe

sentry-core/src/main/java/io/sentry/core/SentryClient.java
marandaneto: not sure if I got it, are you saying to not have this in the `options`? the thing is, it should come from the android module because of the `Context`, so it's either in the options or parameters all the way

bruno-garcia: I see. Makes sense. I missed the fact the value is transport from here to the Transport (if that implementation is interested in this).

PR #208, opened 2020-01-07
Integration refactoring

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [X] Refactoring

## :scroll: Description
Integration refactoring and enhacements.

Currently, integrations call the main hub always but it should call the Sentry static method.
Also, it's not checking if the integration is available for the client.

not yet.

## :pencil: Checklist

- [X] All tests passing


sentry-core/src/main/java/io/sentry/core/HubWrapper.java
```diff
@@ -0,0 +1,22 @@
+package io.sentry.core;
+
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
+
+public abstract class HubWrapper implements IHub {
```
marandaneto: maybe a better name for this class?!

bruno-garcia: The same concept exists in [C# and there's it's called `HubAdapter`](https://github.com/getsentry/sentry-dotnet/blob/385a1e4b9f550c1382f14ddb720a1fa85a26e80e/src/Sentry/Extensibility/HubAdapter.cs). I personally prefer Adapter over Wrapper but they are two names for the same pattern.

marandaneto:
closed in favor of https://github.com/getsentry/sentry-android/pull/245

PR #209, opened 2020-01-08
added event processors to scope

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
added event processors to scope

the current scope doesn't support event processors, but it's part of the unified API.

unit tests.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/SentryClient.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryClient.java
@@ -79,7 +79,12 @@ public final class SentryClient implements ISentryClient {
       // Event has already passed through here before it was cached
       // Going through again could be reading data that is no longer relevant
       // i.e proguard id, app version, threads
-      applyScope(event, scope, hint);
+      event = applyScope(event, scope, hint);
+
+      if (event == null) {
+        // event dropped by the scope event processors
+        return SentryId.EMPTY_ID;
+      }
     } else {
       logIfNotNull(
           options.getLogger(),
@@ -112,7 +117,7 @@ public final class SentryClient implements ISentryClient {
     return event.getEventId();
   }
 
-  private void applyScope(SentryEvent event, @Nullable Scope scope, @Nullable Object hint) {
+  private SentryEvent applyScope(SentryEvent event, @Nullable Scope scope, @Nullable Object hint) {
     if (scope != null) {
       if (event.getTransaction() == null) {
         event.setTransaction(scope.getTransaction());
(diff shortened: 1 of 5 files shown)


sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -150,6 +150,10 @@ private void applyScope(SentryEvent event, @Nullable Scope scope) {
       if (scope.getLevel() != null) {
         event.setLevel(scope.getLevel());
       }
+
+      for (EventProcessor item : scope.getEventProcessors()) {
+        event = item.process(event, hint);
```
bruno-garcia: Where do we apply the event processors from the Options? Maybe we keep them next to each other?
Also, note that we need to bail if a processor returns `null`, instead of passing it to the next processor.
That means the implementor decided to drop the event

marandaneto: > Where do we apply the event processors from the Options? Maybe we keep them next to each other?
we do it on `captureEvent` itself, they are in different methods right now because of the `Hint != Cached` check, but they are executed one after the other (EP scopes 1st and EP options 2nd).
> Also, note that we need to bail if a processor returns `null`, instead of passing it to the next processor.
good catch, will do it.
> That means the implementor decided to drop the event

sentry-core/src/test/java/io/sentry/core/SentryClientTest.kt
```diff
@@ -372,6 +373,30 @@ class SentryClientTest {
...
+    fun `when scope has event processors, they should be applied`() {
+        val event = SentryEvent()
+        val scope = createScope()
+        val processor = mock<EventProcessor>()
+        scope.addEventProcessor(processor)
+
+        val sut = fixture.getSut()
+
+        sut.captureEvent(event, scope)
+        verify(processor, times(1)).process(eq(event), anyOrNull())
```
philipphofmann: You could remove `times(1)` here, because `Mockito.verify()` with only one parameter anyway uses `times(1)`.

marandaneto: @jonesphillip good catch :) time to replace `, times(1)` for nothing hehe, thanks!

philipphofmann: Thanks :) I opened up a PR to get rid of all of them in your repository: https://github.com/getsentry/sentry-android/pull/211

PR #216, opened 2020-01-11
added some unit tests

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
added some unit tests

trying to add some unit tests more, our coverage is not great yet, but some classes are hard to test due framework or static access.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransportGate.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransportGate.java
@@ -5,6 +5,7 @@ import io.sentry.android.core.util.ConnectivityChecker;
 import io.sentry.core.ILogger;
 import io.sentry.core.transport.ITransportGate;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.TestOnly;
 
 final class AndroidTransportGate implements ITransportGate {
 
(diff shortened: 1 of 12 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransportGate.java
```diff
@@ -18,7 +19,11 @@
 
   @Override
   public boolean isSendingAllowed() {
-    Boolean connected = ConnectivityChecker.isConnected(context, loger);
+    return isConnected(ConnectivityChecker.isConnected(context, loger));
```
bruno-garcia: I realized this change didn't introduce it but there's a typo on `loger`

sentry-android-core/src/test/java/io/sentry/android/core/AndroidTransportGateTest.kt
```diff
@@ -8,20 +8,38 @@ import kotlin.test.BeforeTest
...
     @BeforeTest
     fun `set up`() {
         context = ApplicationProvider.getApplicationContext()
+        transportGate = AndroidTransportGate(context, mock())
     }
 
     @Test
-    fun `isConnected won't throw exception`() {
-        val transportGate = AndroidTransportGate(context, mock())
+    fun `isSendingAllowed won't throw exception`() {
```
bruno-garcia: Maybe instead of `does not throw ` it could be `is not null`.
Since we can assert something here.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransportGate.java
marandaneto: yep, fixing it, thx

sentry-android-core/src/test/java/io/sentry/android/core/AndroidTransportGateTest.kt
marandaneto: good catch

PR #223, opened 2020-01-14
added a few java docs (Sentry, Hub and SentryClient)

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
added a few java docs.

we are poor of javadocs.

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
more and more javadocs

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Sentry.java
+++ b/sentry-core/src/main/java/io/sentry/core/Sentry.java
@@ -16,6 +16,10 @@ public final class Sentry {
 
   private static volatile @NotNull IHub mainHub = NoOpHub.getInstance();
 
+  /**
+   * Returns the current (thread’s) hub, if none, clones the mainHub and returns it.
+   * @return the hub
+   */
   private static @NotNull IHub getCurrentHub() {
     IHub hub = currentHub.get();
     if (hub == null) {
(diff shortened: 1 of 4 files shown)


sentry-core/src/main/java/io/sentry/core/Sentry.java
```diff
@@ -69,110 +93,224 @@ public static void init(@NotNull OptionsConfiguration<SentryOptions> optionsConf
     hub.close();
   }
 
+  /**
+   * Close the SDK
+   */
   public static synchronized void close() {
-    IHub hub = getCurrentHub();
+    IHub hub = getCurrentHub(); // TODO: should we not close all the hubs?
```
marandaneto: @bruno-garcia

bruno-garcia:
@marandaneto Ideally all public types their members. 

I'd say at least all public API from `Sentry`, `SentryOptions`, `SentryClient`, `Hub`, `SentryEvent` and `Scope`. Anything else major you'd suggest?

sentry-core/src/main/java/io/sentry/core/Sentry.java
bruno-garcia: This is a gap in the API. `Close` only closes the current Hub and this is the behavior in all SDKs AFAIK.
@HazAT was there any discussion over this issue?

We could hold weak references and dispatch close to all known hubs or something like that.
@mitsuhiko might have opinions here.

marandaneto:
> CI is unhappy somehow

because of #224 already fixed it

sentry-core/src/main/java/io/sentry/core/Sentry.java
HazAT: I think it's fine just to close the current hub/client.
I would expect to call close doing it for all hubs.

PR #226, opened 2020-01-14
clone method - race condition free

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
clone race condition free.

because the clone methods might give some race conditions.

just ran the tests.

## :pencil: Checklist

- [X] All tests passing


sentry-core/src/main/java/io/sentry/core/Breadcrumb.java
```diff
@@ -92,14 +88,13 @@ public void acceptUnknownProperties(Map<String, Object> unknown) {
...
   @Override
   public Breadcrumb clone() throws CloneNotSupportedException {
-    Breadcrumb clone = (Breadcrumb) super.clone();
-
-    clone.timestamp = timestamp != null ? (Date) timestamp.clone() : null;
+    final Breadcrumb clone = (Breadcrumb) super.clone();
 
-    if (data != null) {
-      Map<String, String> dataClone = new ConcurrentHashMap<>();
+    final Map<String, String> dataRef = data;
```
bruno-garcia: Technically it would be safe to do this if the concrete type we're assigning to `Map<String, String>` was thread safe. It's the case for when the SDK sets the map but if the setter is public, someone can set another type which isn't.

One way would be to make it `ConcurrentHashMap<String, String>` instead:

```suggestion
    final ConcurrentHashMap<String, String> dataRef = data;
```

Since we'll be reading from it while potentially another thread will be writing to it.

marandaneto: @bruno-garcia actually it would make more sense to have only `set(key, value)` like Scope, what do you think? so we would this problem, good catch.

bruno-garcia: Makes sense to me. That's how's in .NET too.

PR #229, opened 2020-01-15
more device context (deviceId, connectionType and language)

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
more device context.

more device context, like deviceId, connectionType and language.

executing and a single unit test.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -14,6 +14,7 @@ import android.os.BatteryManager;
 import android.os.Build;
 import android.os.Environment;
 import android.os.LocaleList;
+import android.os.Looper;
 import android.os.StatFs;
 import android.os.SystemClock;
 import android.provider.Settings;
@@ -32,6 +33,7 @@ import io.sentry.core.protocol.DebugMeta;
 import io.sentry.core.protocol.Device;
 import io.sentry.core.protocol.OperatingSystem;
 import io.sentry.core.protocol.SdkVersion;
+import io.sentry.core.protocol.SentryThread;
 import io.sentry.core.protocol.User;
 import io.sentry.core.util.Objects;
 import java.io.BufferedReader;
@@ -158,6 +160,12 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
       }
       if (event.getContexts().getApp() == null) {
         event.getContexts().setApp(getApp(packageInfo));
+      }
+    }
+
+    if (event.getThreads() != null) {
+      for (SentryThread thread : event.getThreads()) {
+        thread.setCurrent(Looper.getMainLooper().getThread().getId() == thread.getId());
       }
     }
   }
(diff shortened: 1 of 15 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -160,6 +162,12 @@ private void processNonCachedEvent(SentryEvent event) {
         event.getContexts().setApp(getApp(packageInfo));
       }
     }
+
+    if (event.getThreads() != null) {
+      for (SentryThread thread : event.getThreads()) {
+        thread.setCurrent(Looper.getMainLooper().getThread().getId() == thread.getId());
+      }
+    }
```
marandaneto: it seems I've used the wrong base branch

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
```diff
@@ -35,20 +37,81 @@ private ConnectivityChecker() {}
...
+      }
     }
-    logger.log(SentryLevel.INFO, "NetworkInfo is null and cannot check network status");
     return null;
   }
+
+  private static @Nullable ConnectivityManager getConnectivityManager(
+      @NotNull Context context, @NotNull ILogger logger) {
+    ConnectivityManager connectivityManager =
+        (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
```
bruno-garcia: Can this cast ever throw?

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
```diff
@@ -23,4 +23,9 @@ class ConnectivityCheckerTest {
     fun `isConnected won't throw exception`() {
         ConnectivityChecker.isConnected(context, mock())
     }
+
+    @Test
+    fun `getConnectionType won't throw exception`() {
+        ConnectivityChecker.getConnectionType(context, mock())
```
bruno-garcia: Ideally the tests would assert something.
A 'does not throw' test is a last resort thing when the code isnt' testable (I guess like in this case).
Would it be possible to get it to return something and assert non null?

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
```diff
@@ -290,6 +293,30 @@ public void setScreenHeightPixels(Integer screenHeightPixels) {
...
+
+  public String getLanguage() {
+    return language;
+  }
+
+  public void setLanguage(String language) {
+    this.language = language;
+  }
+
+  public String getConnectionType() {
```
bruno-garcia: Is connection_type on that list of protocol additions? I recall id and language there

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
marandaneto: no, because of `Context.CONNECTIVITY_SERVICE`

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
marandaneto: I agree, but unless I mock the world, I cannot guarantee that the JVM will return something.
but definitely it's on my radar to test all the android FW related classes later on.

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
marandaneto: yes, it is

PR #235, opened 2020-01-16
update ndk for new sentry-native version

Swatinem:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Adapt to API changes in `sentry-native`, and temporarily point to a specific commit that builds.

So far only checked that it builds

## :pencil: Checklist

- [ ] All tests passing


Swatinem:
> and see if the events get sent?

got that working now. will coordinate with @jan-auer to further discuss when which part of code should land.

sentry-android-ndk/build.gradle.kts
```diff
@@ -36,6 +36,7 @@ android {
 
         externalNativeBuild {
             cmake {
+                targets.add("sentry-android")
```
marandaneto: that's not right, I'll need to duplicate this line elsewhere, this should be fixed on source origin.

marandaneto: that's not right, I'll need to duplicate this line elsewhere, this should be fixed on source origin.

PR #237, opened 2020-01-16
few improvements identified by a pre-release list

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
few improvements identified by a pre-release list.

to meet the unified API.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Scope.java
+++ b/sentry-core/src/main/java/io/sentry/core/Scope.java
@@ -59,7 +59,8 @@ public final class Scope implements Cloneable {
     this.user = user;
   }
 
-  public @NotNull List<String> getFingerprint() {
+  @NotNull
+  List<String> getFingerprint() {
     return fingerprint;
   }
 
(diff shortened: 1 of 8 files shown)


sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -93,6 +94,8 @@ void addBreadcrumb(@NotNull Breadcrumb breadcrumb, boolean executeBeforeBreadcru
       }
 
       if (breadcrumb == null) {
+        //        options.getLogger().log(SentryLevel.INFO, "Breadcrumb was dropped by scope
```
bruno-garcia: Sounds fair to have this log. Perhaps in Debug mode instead?
Or remove dead code?

marandaneto: actually we don't have access to options object on this class, that's why, it's a TODO.
I'll try to fix this "workaround" with the Scope class and Before BreadCrumb before GA, it's a design flaw.

PR #245, opened 2020-01-22
Refactor integrations to match unified API

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
Refactor integrations to match unified API.

Integrations should use always the current hub.
Default integrations should have an opt-out flag.
Default integrations should run only once.

not yet, I'll write unit tests for everything, this is just a draft as a POC

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
write unit tests, convert EventProcessors to Integrations

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
@@ -45,7 +45,7 @@ final class AnrIntegration implements Integration, Closeable {
   @TestOnly
   void reportANR(IHub hub, final @NotNull ILogger logger, ApplicationNotResponding error) {
     if (!hub.isIntegrationEnabled(AnrIntegration.class)) {
-      logger.log(SentryLevel.WARNING, "AnrIntegration is not enabled to the current hub.");
+      logger.log(SentryLevel.INFO, "AnrIntegration is not enabled for the current hub.");
       return;
     }
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
@@ -44,6 +44,11 @@ final class AnrIntegration implements Integration, Closeable {
 
   @TestOnly
   void reportANR(IHub hub, final @NotNull ILogger logger, ApplicationNotResponding error) {
+    if (!hub.isIntegrationEnabled(AnrIntegration.class)) {
+      logger.log(SentryLevel.WARNING, "AnrIntegration is not enabled to the current hub.");
+      return;
+    }
+
     logger.log(SentryLevel.INFO, "ANR triggered with message: %s", error.getMessage());
 
     Mechanism mechanism = new Mechanism();
--- a/sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
@@ -2,6 +2,7 @@ package io.sentry.android.core;
 
 import android.os.FileObserver;
 import io.sentry.core.IEnvelopeSender;
+import io.sentry.core.IHub;
 import io.sentry.core.ILogger;
 import io.sentry.core.SentryLevel;
 import io.sentry.core.util.Objects;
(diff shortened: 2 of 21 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
```diff
@@ -44,6 +44,11 @@ private void register(IHub hub, SentryAndroidOptions options) {
 
   @TestOnly
   void reportANR(IHub hub, final @NotNull ILogger logger, ApplicationNotResponding error) {
+    if (!hub.isIntegrationEnabled(AnrIntegration.class)) {
+      logger.log(SentryLevel.WARNING, "AnrIntegration is not enabled to the current hub.");
```
bruno-garcia: Not sure WARN is a good choice here since it's a legit case.
The API allows you to bind a new client to the hub and capture errors. It's intentional that integrations that are not set to that hub would not run in that case.

Can we have?:
```suggestion
      logger.log(SentryLevel.INFO, "AnrIntegration is not enabled to the current hub.");
```

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
```diff
@@ -30,6 +34,13 @@ public void onEvent(int eventType, @Nullable String relativePath) {
       return;
     }
 
+    if (!hub.isIntegrationEnabled(EnvelopeFileObserverIntegration.class)) {
+      logger.log(
+          SentryLevel.WARNING,
```
bruno-garcia:
```suggestion
          SentryLevel.INFO,
```

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
```diff
@@ -30,6 +34,13 @@ public void onEvent(int eventType, @Nullable String relativePath) {
...
+          "EnvelopeFileObserverIntegration is not enabled to the current hub.");
```
bruno-garcia:
```suggestion
          "EnvelopeFileObserverIntegration is not enabled for the current hub.");
```

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -484,4 +504,39 @@ public void flush(long timeoutMills) {
...
+        options
+            .getLogger()
+            .log(SentryLevel.ERROR, "Error while getSentryClient: " + e.getMessage(), e);
+      }
+    }
+    return sentryClient;
+  }
+
+  @Override
+  public <T extends Integration> boolean isIntegrationEnabled(Class<T> integration) {
```
bruno-garcia: This needs to take the actual instance. So when we do `new AnrIntegration` we're matching against this actual object.

If you had 3 instances of the same type in the list, it would return true to all of them and 3 integrations would run.

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
```diff
@@ -35,6 +35,15 @@ public void register(@NotNull IHub hub, @NotNull SentryOptions options) {
       ExecutorService es = Executors.newSingleThreadExecutor();
       es.submit(
           () -> {
+            if (!hub.isIntegrationEnabled(SendCachedEventFireAndForgetIntegration.class)) {
```
bruno-garcia: As the other comment. I know we've discussed using the type as a possibility but it has gotchas. I believe should use reference comparison to make sure it's the one instance added.

sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
marandaneto: makes sense, thanks

sentry-core/src/main/java/io/sentry/core/Hub.java
marandaneto: that's true, but on the other hand, if we do by reference, a user could `new UncaughtExceptionHandlerIntegration` and have 2 global hooks, unless I protect those integrations with a static field, what do you think?

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
marandaneto: answer: https://github.com/getsentry/sentry-android/pull/245#discussion_r370108607

sentry-core/src/main/java/io/sentry/core/Hub.java
marandaneto: @HazAT do you have options here?

HazAT: @marandaneto is right, we just check the class not the instance. The reason is that we don't want global stuff to run twice. So we need to make sure that every Client still get the instance of the integration (in case the integration has different options then for another client) but only if `register` was never called for this class before we should do it.

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -27,12 +28,23 @@
...
+          integration.register(HubAdapter.getInstance(), options);
+        }
+      }
+    }
+
     for (Integration integration : options.getIntegrations()) {
-      integration.register(this, options);
+      integration.register(HubAdapter.getInstance(), options);
     }
   }
```
marandaneto: @bruno-garcia @HazAT right now it won't `defaultIntegrations.register` twice, is it the correct behavior or this should only be done for integrations that are global hooks? by coincidence, right now all the `defaultIntegrations` are global hooks, so it makes sense to run only once, but if we add a new `defaultIntegration` that is not a global hook, it won't be able to register again.

sentry-core/src/main/java/io/sentry/core/Hub.java
bruno-garcia: The integration instance must see if itself is in the current hub, not its type?

@mitsuhiko last we discussed it was about the instance, so now I'm confused.
Isn't there a difference between _global integrations_ and non global?

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
bruno-garcia: Will resolve this since we'll settle on the behavior on the other comment.

bruno-garcia:
After a lot of discussions here (@HazAT, @rhcarvalho, @untitaker) we decided not to go on with this change.

It would be nice to keep `HubAdapter` around also for sentry-core DI scenarios though.

Sorry for going back and forth.

marandaneto:
@bruno-garcia, so just to see if I got this right. The whole thing keep as it is, but it would be nice to have the HubAdapter calling the static method and that’s all?
Thanks

bruno-garcia:
Yeah because on DI scenarios, we'll register in the container `IHub` with the implementation `HubAdapter.getInstance`. This way it's possible to consume the SDK by taking `IHub` in the ctor.
Does this make sense?

marandaneto:
in favor of https://github.com/getsentry/sentry-android/pull/250

PR #247, opened 2020-01-24
feat: Hub mode configurable

bruno-garcia:
A Hub per Thread makes sense in server applications.
That's not the case for Mobile or desktop apps. Where it's a single 'user session' for the whole process, UI thread is scarce resource and lots of work are queued to worker threads.

This PR introduces a flag to control the behavior of the `Sentry` static class. By default, it'll continue to clone the `mainHub` per thread. If the flag is flipped (which happens in the Android SDK), a single Hub (and its scope) is shared. Meaning the breadcrumbs added in the UI thread as well as worker threads making HTTP requests also end up on outgoing events.

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
@@ -56,13 +56,13 @@ public final class SentryAndroid {
       @NotNull ILogger logger,
       @NotNull Sentry.OptionsConfiguration<SentryAndroidOptions> configuration) {
     try {
-      Sentry.setGlobalHubMode(true);
       Sentry.init(
           OptionsContainer.create(SentryAndroidOptions.class),
           options -> {
             AndroidOptionsInitializer.init(options, context, logger);
             configuration.configure(options);
-          });
+          },
+          true);
     } catch (IllegalAccessException e) {
       logger.log(SentryLevel.FATAL, "Fatal error during SentryAndroid.init(...)", e);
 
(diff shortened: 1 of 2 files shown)


marandaneto:
@bruno-garcia Happy with the changes, but what’s about having isGlobalHubMode on the SentryOptions and being able to flip the flag only on init method which is supposed to be called only once, so we avoid the problems mentioned by @metlos?

bruno-garcia:
The goal of this flat is to be flipped before initializing, and only once. Agree this isn't very nice.
We could add it to Options but that is also used to initialize a Hub and a SentryClient and this flag is really only useful to the `Sentry` static class.

Maybe we need to discuss this change more. The problem here is that on Android we want a single Hub. By default `configureScope` mutates a global object and any event sent out includes all data from that scope.

On server apps we would want hubs to fork by default. Thoughts?

marandaneto:
@bruno-garcia yep got it, my only concern would be setGlobalHubMode as a public method, that’s why I am not sure if it’s a good idea, I would be happy being private package at least, maybe even one of the Sentry.init methods with an overload so we can pass this Boolean there, or again, from options, but the big gain is to not have a public api for this that may be changed on runtime, so we can change in the future, what do you think?

bruno-garcia:
If it could be package private it would be great but `sentry-android` needs to call it.
It's a public API and I don't believe it needs to be documented (besides javadocs).

bruno-garcia:
@mitsuhiko this is what we're merging wrt 'global' hub in mobile. note push and pop scope as no-op.

PR #248, opened 2020-01-27
fix nativec build

marandaneto:
test


marandaneto:
in favor of https://github.com/getsentry/sentry-android/pull/251

PR #249, opened 2020-01-28
overloads for global hub mode

marandaneto:


sentry-core/src/main/java/io/sentry/core/Sentry.java
```diff
@@ -49,7 +51,16 @@ public static boolean isEnabled() {
...
-    init(new SentryOptions());
+    init(GLOBAL_HUB_DEFAULT_VALUE);
+  }
+
+  /**
+   * Initializes the SDK
+   *
+   * @param globalHubMode the globalHubMode
+   */
+  public static void init(boolean globalHubMode) {
```
bruno-garcia: @marandaneto can we remove this overload? I'd rather we have this flag less exposed. It'd be fine to only be able to flip it on the overload with all parameters.

PR #261, opened 2020-02-06
Chore/craft

bruno-garcia:


bruno-garcia:
@marandaneto still missing the `bintray` target on craft which I'll start next.
The problem is splitting the `bintrayUpload` gradle target so that it takes the already built artifacts, instead of running `gradle build bintrayUpload` and publishing new artifacts built locally.

bruno-garcia:
When adding the code to publish artifacts to zeus I realized the following:

image

Packages are not self contained. And the name of the pom file doesn't include the name of the package it relates to.

So we could rename each of the files by prepending the package name but I'm not yet sure there's any easy way to upload the `.jar` and `.aar` with these `pom` files to bintray.

I noticed we're not publishing anything to maven central yet. Whatever we do here has to be with the end result of having this stuff on jcenter and maven central.

buildSrc/src/main/java/Config.kt
```diff
@@ -49,11 +49,9 @@ object Config {
     object Sentry {
         val SENTRY_CLIENT_NAME = "sentry.java.android"
         val group = "io.sentry"
-//        TODO: change version to publish new version
-        val version = "2.0.0-rc03"
+        val version = "2.0.0-rc03" // updated by craft
         val description = "SDK for sentry.io"
-//        TODO: change version code to publish new version, follow the pattern of `version`
-        val buildVersionCode = 20014
+        val buildVersionCode = 20014 // updated by craft
```
marandaneto: I'd recommend moving these properties away from here to https://github.com/getsentry/sentry-android/blob/master/gradle.properties

scripts/bump-version.sh
```diff
@@ -0,0 +1,13 @@
...
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $SCRIPT_DIR/..
+
+OLD_VERSION="$1"
+NEW_VERSION="$2"
+
+# Add the new version
+sed -i '' -e "s/val version = \"\(.*\)\" \/\/ updated by craft/val version = \"$NEW_VERSION\" \/\/ updated by craft/g" buildSrc/src/main/java/Config.kt
+# Increment the buildVersionCode
+perl -pi -e 's{val buildVersionCode = (\d+) // updated by craft}{$n=$1+1; "val buildVersionCode = $n // updated by craft"}e' buildSrc/src/main/java/Config.kt
```
marandaneto: same as a comment above, upgrading versions on https://github.com/getsentry/sentry-android/blob/master/gradle.properties instead of `buildSrc/src/main/java/Config.kt`

marandaneto: `buildVersionCode` could be a hash from current date + incremental value, so it's easier to track. like sentry's `semverish ` suggest, what do you think?

scripts/zeus.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -eux
+
+yarn global add @zeus-ci/cli
+
+zeus upload -t "application/zip+aar" ./*/build/outputs/aar/*release.aar
+zeus upload -t "application/zip+jar" ./*/build/libs/*release.jar
```
marandaneto: what's about poms? eg
`build/publications/release/pom-default.xml`
or even `*javadocs.jar`, `*sources.jar`?

bruno-garcia: The jars are all match the pattern. The pom is not because their names are all the same. I mentioned on slack. Wither we rename or find another way

buildSrc/src/main/java/Config.kt
bruno-garcia: What is the benefit?

marandaneto: not much tbh, it's a key/value props file and it'd be easier to change with a CLI tool

scripts/zeus.sh
marandaneto: true about the pom files, but not for sources and javadocs files as they have a `*-sources.jar` and `*-javadocs.jar` pattern if I recall it well.

bruno-garcia:
@marandaneto Are we publishing using the artifacts?
Do I need to bundle them in a zip or something and upload to zeus/GCP?

PR #262, opened 2020-02-10
feat: Take sampleRate from metadata

bruno-garcia:


sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -44,6 +46,14 @@ static void applyMetadata(Context context, SentryAndroidOptions options) {
         options.getLogger().log(SentryLevel.DEBUG, "isAnrEnabled read: %s", isAnrEnabled);
         options.setAnrEnabled(isAnrEnabled);
 
+        if (options.getSampleRate() == null) {
+          Double sampleRate = metadata.getDouble(SAMPLE_RATE, -1);
+          options.getLogger().log(SentryLevel.DEBUG, "sampleRate read: %s", sampleRate);
+          if (sampleRate != -1) {
+            options.setSampleRate(sampleRate);
+          }
+        }
```
marandaneto: what's about:
`options.setSampleRate(metadata.getDouble(SAMPLE_RATE, options.getSampleRate()));`

bruno-garcia: Throws when getSampleRate is null. I need to know when no value was defined at all so I used `-1` for that

bruno-garcia: Also, I think we need to change the approach on the other one. 

Right now we're overriding any value set on option with the value from metadata. It should be the way I did here where we check first if options doesn't have any settings, and only then read from metadata. 

Separate PR for that though

fdrabek:
Can we please also add release and environment?

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
marandaneto: > Also, I think we need to change the approach on the other one.
> 
> Right now we're overriding any value set on option with the value from metadata. It should be the way I did here where we check first if options doesn't have any settings, and only then read from metadata.
> 
> Separate PR for that though

options callback is triggered afterward, so no values are replaced unless one bypasses it.

PR #263, opened 2020-02-11
Support mills timestamp format

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
Support mills timestamp format.
eg 1581410911.988
which is {seconds}.{mills}

JS sends mills timestamp format and we need to support hybrid SDKs.

unit test.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/adapters/DateDeserializerAdapter.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/adapters/DateDeserializerAdapter.java
@@ -27,7 +27,15 @@ public final class DateDeserializerAdapter implements JsonDeserializer<Date> {
     try {
       return json == null ? null : DateUtils.getDateTime(json.getAsString());
     } catch (Exception e) {
-      logger.log(SentryLevel.ERROR, "Error when deserializing Date", e);
+      logger.log(
+          SentryLevel.ERROR,
+          "Error when deserializing UTC timestamp format, might be mills timestamp format.",
+          e);
+    }
+    try {
+      return DateUtils.getDateTimeWithMillsPrecision(json.getAsString());
+    } catch (Exception e) {
+      logger.log(SentryLevel.ERROR, "Error when deserializing mills timestamp format.", e);
     }
     return null;
   }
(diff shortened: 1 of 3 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/adapters/DateDeserializerAdapter.java
```diff
@@ -27,7 +27,15 @@ public Date deserialize(JsonElement json, Type typeOfT, JsonDeserializationConte
     try {
       return json == null ? null : DateUtils.getDateTime(json.getAsString());
     } catch (Exception e) {
-      logger.log(SentryLevel.ERROR, "Error when deserializing Date", e);
+      logger.log(
+          SentryLevel.ERROR,
```
HazAT: I wouldn't log `ERROR` here, maybe `DEBUG` since we handled the error gracefully.

bruno-garcia: This is just sdk internal logging. I'd say better log as error and let it stand out. Debug is super noisy and we might miss this error in the middle of the noise in a debug session. At least a warn but tbh this is an error.
We're logging and not throwing and this is gracefully dealing with errors imho

marandaneto: I'd agree with @bruno-garcia , I've just changed because there are 2 input types, UTC and mills timestamp, so I've logged as DEBUG the 1st one, because  there are 2 tries, so even being an error, it might be the expected case, the 2nd case is logged as an ERROR, because if it failed in the 1st and 2nd case, we should know where it comes from

bruno-garcia: oh, my bad I missed that.
It would be ERROR only if we failed to deserialize both then, I guess. if it's valid in either format it's business as usual?

bruno-garcia: Agree with @marandaneto's approach here

PR #264, opened 2020-02-11
fix: breacrumb.data to string,object, Add LOG level

HazAT:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Add fix that breadcrumb.data is `<String, Object>` instead of `<String, String>`
Add `LOG` as log level.

Support for Hybrid SDKs

unit test.

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
🚢

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidLogger.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidLogger.java
@@ -24,10 +24,6 @@ final class AndroidLogger implements ILogger {
   public void log(SentryLevel level, String message, Throwable throwable) {
 
     switch (level) {
-      case LOG:
-      case DEBUG:
-        Log.d(tag, message, throwable);
-        break;
       case INFO:
         Log.i(tag, message, throwable);
         break;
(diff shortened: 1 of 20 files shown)


sentry-core/src/main/java/io/sentry/core/SentryLevel.java
```diff
@@ -2,6 +2,7 @@
 
 /** the SentryLevel */
 public enum SentryLevel {
+  LOG,
```
bruno-garcia: @HazAT This isn't in the docs eh?
Could you please clarify what this level is?

Oh, I assume it's some JS thing leaking in? Can we leave the value be mapped as unknown and travel through the serialization/deserialization instead?

marandaneto: @bruno-garcia yes, this is JS leaking in.
this is a value and not an unknown key, as we have a strong typed `enum`, it won't work, unless we add a generic value to `SentryLevel` and assume it for all the unknown values.

bruno-garcia: `LogLevel.Log` really makes no sense imho.
But sounds like the most pragmatic way to solve this though.

Can we have a javadoc on that explain where it comes from?

HazAT: @bruno-garcia Added a javadoc

sentry-android-core/src/main/java/io/sentry/android/core/AndroidLogger.java
```diff
@@ -39,39 +36,26 @@ public void log(SentryLevel level, String message, Throwable throwable) {
...
       case INFO:
         return Log.INFO;
       case WARNING:
         return Log.WARN;
       case FATAL:
         return Log.ASSERT;
-      case ERROR:
+      case LOG:
+      case DEBUG:
       default:
```
bruno-garcia: Sorry just spotted this and afaik default is error for issue events. 
But we should not be adding level by default so the server infers that

marandaneto: this is only used for calling the correct logging, not related to sentryEvent.level nor breadcrumb.level

bruno-garcia: My bad.

PR #267, opened 2020-02-11
feat: attach threads/stacktraces

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
feat: attach stacktraces

attach stacktraces is disabled by default, but it should have a flag to opt-in.
right now is enabled by default and there's no opt-out.

unit test.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
+++ b/sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
@@ -19,7 +19,8 @@ public final class MainEventProcessor implements EventProcessor {
         new SentryStackTraceFactory(options.getInAppExcludes(), options.getInAppIncludes());
 
     sentryExceptionFactory = new SentryExceptionFactory(sentryStackTraceFactory);
-    sentryThreadFactory = new SentryThreadFactory(sentryStackTraceFactory);
+    sentryThreadFactory =
+        new SentryThreadFactory(sentryStackTraceFactory, this.options.isAttachStacktrace());
   }
 
   MainEventProcessor(
@@ -71,15 +72,8 @@ public final class MainEventProcessor implements EventProcessor {
       event.setDist(options.getDist());
     }
 
-    if (event.getThreads() == null) {
-      // that means user has called Sentry.captureMessage
-      if (event.getMessage() != null && event.getExceptions() == null) {
-        if (options.isAttachStacktrace()) {
-          event.setThreads(sentryThreadFactory.getCurrentThreads());
-        }
-      } else {
-        event.setThreads(sentryThreadFactory.getCurrentThreads());
-      }
+    if (event.getThreads() == null && options.isAttachThreads()) {
+      event.setThreads(sentryThreadFactory.getCurrentThreads());
     }
   }
 }
--- a/sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
+++ b/sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
@@ -72,7 +72,14 @@ public final class MainEventProcessor implements EventProcessor {
     }
 
     if (event.getThreads() == null) {
-      event.setThreads(sentryThreadFactory.getCurrentThreads());
+      // that means user has called Sentry.captureMessage
+      if (event.getMessage() != null && event.getExceptions() == null) {
+        if (options.isAttachStacktrace()) {
+          event.setThreads(sentryThreadFactory.getCurrentThreads());
+        }
+      } else {
+        event.setThreads(sentryThreadFactory.getCurrentThreads());
+      }
     }
   }
 }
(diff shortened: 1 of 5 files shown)


bruno-garcia:
One might come and ask not to have all threads sent at all. Which would happen for captureException even if attachStacktrace false.
Maybe we can add a new setting called `attachAllTreads` (with a better name) to control that?

And not have attachStacktrace at all?

sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
```diff
@@ -72,7 +72,14 @@ private void processNonCachedEvent(SentryEvent event) {
...
     if (event.getThreads() == null) {
-      event.setThreads(sentryThreadFactory.getCurrentThreads());
+      // that means user has called Sentry.captureMessage
+      if (event.getMessage() != null && event.getExceptions() == null) {
+        if (options.isAttachStacktrace()) {
+          event.setThreads(sentryThreadFactory.getCurrentThreads());
+        }
+      } else {
+        event.setThreads(sentryThreadFactory.getCurrentThreads());
+      }
```
marandaneto: I'd say that it's a good guess of detecting a `Sentry.captureMessage` call but it could be a custom `Sentry.captureEvent` call with message and exceptions and in this case, if `attachStacktrace` is `disabled`, it still reports threads.
what do you think @bruno-garcia

marandaneto:
> One might come and ask not to have all threads sent at all. Which would happen for captureException even if attachStacktrace false.
> Maybe we can add a new setting called `attachAllTreads` (with a better name) to control that?
> 
> And not have attachStacktrace at all?

I'd agree that either we change its name from `attachStacktrace` to `attachThreads` or to still send threads, but without stacktraces, not sure which one makes more sense to the final user.

I believe `captureException` should not be affected at all, as the unified API defines only for messages.
https://docs.sentry.io/error-reporting/configuration/?platform=javascript#attach-stacktrace

@HazAT do you have opinions?

HazAT:
Let's leave it like this, for now, it's very rare that a user would like to have all threads mostly they are interested in the thread they called this from.

bruno-garcia:
@marandaneto Adding Threads (themselves and their stack traces) adds a lot of data and I can see people might be asking an opt-out for that.
In the case of Exception, that has a stacktrace itself, so if someone opts out to sending threads, it'll always get the exception stacktrace (as other SDKs work).
attachStacktrace is usually to add a single stack trace, for the current thread (the one called `captureMessage`). If we have a single options: attachThreads (default to `true`), we give the behavior of `attachStacktraces` for free but a way to also opt-out of it.

@HazAT Right now we're adding all threads. I believe our guideline says we should opt-in to features like this and let users opt-out if they don't want it. I'm proposing not adding `attachStacktrace` option but only `attachThreads` instead to be able to turn it off. Essentially we're attaching stacktraces for all events anyway unless user opts out.

marandaneto:
> @marandaneto Adding Threads (themselves and their stack traces) adds a lot of data and I can see people might be asking an opt-out for that.

I think it should be opt-in and not opt-out, as you just stated, it's a lot of data and mostly irrelevant if it's just a `captureMessage`, but I understand that the guidelines say the opposite, so I'm fine with it too.

> In the case of Exception, that has a stacktrace itself, so if someone opts out to sending threads, it'll always get the exception stacktrace (as other SDKs work).
> attachStacktrace is usually to add a single stack trace, for the current thread (the one called `captureMessage`). If we have a single options: attachThreads (default to `true`), we give the behavior of `attachStacktraces` for free but a way to also opt-out of it.
> 
> @HazAT Right now we're adding all threads. I believe our guideline says we should opt-in to features like this and let users opt-out if they don't want it. I'm proposing not adding `attachStacktrace` option but only `attachThreads` instead to be able to turn it off. Essentially we're attaching stacktraces for all events anyway unless user opts out.

So I agree that we need an `attachThreads` too, so users may opt-out of it, same for `attachStacktrace`, however `attachStacktrace` is disabled by default on the unified API docs, so we should decide about the default state of both and if we make difference between `captureMessage` and `captureException`

bruno-garcia:
> however attachStacktrace is disabled by default on the unified API docs

We can leave it off by default too, but the attachThreads (non "unified") enabled by default.
It's nice to give a rich event by default, users will see it on the first event coming in, they can always come and turn it off. If we don't turn this on, most people won't go through the settings and check what they are and turn them on.

marandaneto:
> > however attachStacktrace is disabled by default on the unified API docs
> 
> We can leave it off by default too, but the attachThreads (non "unified") enabled by default.
> It's nice to give a rich event by default, users will see it on the first event coming in, they can always come and turn it off. If we don't turn this on, most people won't go through the settings and check what they are and turn them on.

implemented as discussed, the different thing is, it's detached to `captureMessage` and/or `captureEvent`/`captureException`, based on the options, it will attach or not threads/stacktraces.
`attachThreads` enabled by default, `attachStacktraces` disabled though.

PR #269, opened 2020-02-14
fix: preview package manager

bruno-garcia:


sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -227,10 +227,10 @@ private SdkVersion getSdkVersion() {
     sdkVersion.setName("sentry.java.android");
     String version = BuildConfig.VERSION_NAME;
     sdkVersion.setVersion(version);
-    sdkVersion.addPackage("sentry-core", version);
-    sdkVersion.addPackage("sentry-android-core", version);
+    sdkVersion.addPackage("maven:sentry-core", version);
```
marandaneto: so it's missing `maven:sentry-android` here, right? I remember asking it but not sure why

bruno-garcia: good point, yes we are

bruno-garcia: Well, we don't know though. It's missing only if it was installed. Which we don't know here, do we? How can we figure out if that package was added? Maybe meta-data merging?

marandaneto: we could through reflection or adding something to manifest and check it on runtime or the backend side does the grouping and will know if `android-core` `android-ndk` is there, but yes on the client side would be easier and less error-prone.

bruno-garcia: agreed, will raise for that

PR #277, opened 2020-02-21
set current threadId when theres no mechanism to rely on

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
set current threadId when theres no mechanism to rely on.

exception had no threadId if it was not a exception mechanism and this makes the thread selector not happy on the dashboard.

unit test and debugging it.

## :pencil: Checklist

- [X] All tests passing


marandaneto:
> makes sense to send it anyway but ideally the front end wouldn't rely on having thread id at all and just show the exception component.

yep I agree, but if you do send exception + threads, it should be like this so the current behavior would work without changes on the frontend.

PR #281, opened 2020-02-24
session and envelopes

bruno-garcia:
Didn't run a single line of code yet. Supposed to build at least.
I'm hoping we can have some discussions over some stuff, then open small PRs addressing individual parts of this feature.

This PR will likely never be merged!


sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -42,6 +45,10 @@
   /** Scope's SentryOptions */
   private final @NotNull SentryOptions options;
 
+  // TODO Consider: Scope clone doesn't clone sessions
+  private volatile @Nullable Session session;
+  private final @NotNull Object sessionLock = new Object();
```
marandaneto: what's about `ReentrantLock`

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -1,5 +1,7 @@
 package io.sentry.core;
 
+import com.sun.xml.internal.messaging.saaj.soap.Envelope;
```
marandaneto: wrong import here eh?

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -112,6 +116,21 @@ public SentryId captureEvent(SentryEvent event, @Nullable Scope scope, @Nullable
       return SentryId.EMPTY_ID;
     }
 
+    SentryEvent finalEvent = event;
```
marandaneto: should it be `final` then?

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -123,6 +142,31 @@ public SentryId captureEvent(SentryEvent event, @Nullable Scope scope, @Nullable
...
   }
 
+  @Override
+  public void captureSession(@NotNull Session session) throws IOException {
+    if (!options.isEnableSessionTracking()) {
+      options.getLogger().log(SentryLevel.WARNING, "Session tracking is disabled in the options.");
+      return;
+    }
+
+    if (session.getRelease() != null) {
```
marandaneto: `==` or am I missing something?

sentry-core/src/main/java/io/sentry/core/SentryEnvelope.java
```diff
@@ -28,7 +34,19 @@ public SentryEnvelope(
...
   public SentryEnvelope(SentryId sentryId, Iterable<SentryEnvelopeItem> items) {
-    header = new SentryEnvelopeHeader(sentryId, null);
+    header = new SentryEnvelopeHeader(sentryId);
     this.items = items;
   }
+
+  // Single item envelope with an envelope-allocated id
+  public SentryEnvelope(SentryEnvelopeItem item) {
+    header = new SentryEnvelopeHeader();
+    ArrayList<SentryEnvelopeItem> items = new ArrayList<>(1);
```
marandaneto: use `List` instead of `ArrayList`

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeader.java
```diff
@@ -1,21 +1,31 @@
...
 
 import io.sentry.core.protocol.SentryId;
+
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-final class SentryEnvelopeHeader {
-  private final SentryId eventId;
-  private final String auth;
+public final class SentryEnvelopeHeader {
```
marandaneto: we could annotate with `@ApiStatus.Internal`

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeItemHeader.java
```diff
@@ -1,16 +1,25 @@
...
 
+  // TODO: Looks like a type here that defaults to String for unknown values would be ideal
   public String getType() {
     return type;
   }
 
   public int getLength() {
+    if (getLength != null) {
+      return getLength();
+    }
```
marandaneto: let's rename `getLength` field, it looks an endless loop method

build.gradle.kts
```diff
@@ -57,9 +57,9 @@ allprojects {
                 TestLogEvent.FAILED)
             dependsOn("cleanTest")
         }
-        withType<JavaCompile> {
-            options.compilerArgs.addAll(arrayOf("-Xlint:all", "-Werror", "-Xlint:-classfile"))
-        }
+//        withType<JavaCompile> {
+//            options.compilerArgs.addAll(arrayOf("-Xlint:all", "-Werror", "-Xlint:-classfile"))
+//        }
```
marandaneto: not needed anymore?

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -1,12 +1,15 @@
 package io.sentry.core;
 
 import io.sentry.core.protocol.User;
+
+import java.lang.ref.WeakReference;
```
marandaneto: why/where do we need this?

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -123,6 +142,31 @@ public SentryId captureEvent(SentryEvent event, @Nullable Scope scope, @Nullable
     return event.getEventId();
   }
 
+  @Override
+  public void captureSession(@NotNull Session session) throws IOException {
+    if (!options.isEnableSessionTracking()) {
+      options.getLogger().log(SentryLevel.WARNING, "Session tracking is disabled in the options.");
+      return;
+    }
```
marandaneto: this should be in the `Hub` eh

sentry-core/src/main/java/io/sentry/core/SentryEnvelope.java
```diff
@@ -28,7 +34,19 @@ public SentryEnvelope(
...
+    items.add(item);
+    this.items = items;
+  }
+
+  public static SentryEnvelope fromSession(ISerializer serializer, Session session) throws IOException {
```
marandaneto: can be `final`

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeItem.java
```diff
@@ -1,20 +1,81 @@
...
+          serializer.serialize(session, writer);
+          writer.flush();
+          return stream.toByteArray();
+        }
+    });
+
+    SentryEnvelopeItemHeader itemHeader = new SentryEnvelopeItemHeader(
+      "session",
+      () -> cachedItem.getBytes().length,
+      "application/json",
```
marandaneto: ideally, a `const`, have seen it somewhere.

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
```diff
@@ -304,4 +304,20 @@ public DebugMeta getDebugMeta() {
...
+  public boolean isCrashed() {
+    if (level == SentryLevel.FATAL) {
+      return true;
+    }
+    if (exception != null) {
+      for (SentryException e: exception.getValues()) {
+        if (e.getMechanism() != null && !e.getMechanism().isHandled()) {
+          return true;
+        }
+      }
```
marandaneto: does it make sense to add a `transient` field to` SentryEvent` if the event is handled or not? so when adding exceptions to the event, as we know when it's not handled, we could read directly from there, we would spare this `for`, but with the overhead of having in both places.

sentry-core/src/main/java/io/sentry/core/Session.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.Date;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public final class Session {
+
+  public enum State {
+    // TODO: What about 'Started' 'Ended'? Ok is not very clear
+    Ok,
+    Exited,
+    Crashed
```
marandaneto: let's make all uppercase, so its easier to cast json<-> enum, otherwise we would need to `capitalize`, we also keep the same pattern as `SentryLevel`

sentry-core/src/main/java/io/sentry/core/Session.java
```diff
@@ -0,0 +1,99 @@
+package io.sentry.core;
+
+import java.util.Date;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public final class Session {
+
+  public enum State {
+    // TODO: What about 'Started' 'Ended'? Ok is not very clear
```
marandaneto: yep, the wording could be improved IMHO

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -148,6 +149,108 @@ public TransportResult send(final SentryEvent event) throws IOException {
...
+      // need to also close the input stream of the connection
+      connection.getInputStream().close();
+      options.getLogger().log(DEBUG, "Envelope sent %s successfully.", envelope.getHeader().getEventId());
+      return TransportResult.success();
+    } catch (IOException e) {
+      long retryAfterMs = HTTP_RETRY_AFTER_DEFAULT_DELAY_MS;
+      final String retryAfterHeader = connection.getHeaderField("Retry-After");
+      if (retryAfterHeader != null) {
+        // TODO: Protocol undefined yet but ...
+        String rateLimitDetailString = connection.getHeaderField("X-Sentry-Rate-Limit");
```
marandaneto: wait `Retry-After` and `X-Sentry-Rate-Limit`, I need to update myself about `X-Sentry-Rate-Limit`

sentry-core/src/main/java/io/sentry/core/SentryClient.java
bruno-garcia: just trash

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeItemHeader.java
bruno-garcia: might be since callables need to be invoked with `.call()`

build.gradle.kts
bruno-garcia: just to get it to compile

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
bruno-garcia: Maybe something to discuss in the TSC. So far it's an agreement based on either case

sentry-core/src/main/java/io/sentry/core/Session.java
bruno-garcia: Need to discuss with Armin

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
bruno-garcia: It's in `Rate Limiting Updates` internal doc

bruno-garcia:
@marandaneto I've "resolved" the issues that are not required to be addressed to get this to work since we might shuffle everythign around anyway.
Once we agree on the approach we're taking and we have somethig that actually works, we can do a proper code review pass. 

Let's use the `Draft` PR mode to flag that. Not planning to merge a PR that comments out the warnings as errors :)

sentry-core/src/main/java/io/sentry/core/transport/AsyncConnection.java
```diff
@@ -83,6 +88,43 @@ public void send(final SentryEvent event, final @Nullable Object hint) throws IO
...
+    for (SentryEnvelopeItem item : envelope.getItems()) {
+      if (item.getHeader() != null && transport.isRetryAfter(item.getHeader().getType())) {
+        if (dropItems == null) {
+          dropItems = new ArrayList<>();
+        }
+        dropItems.add(item);
+      }
+    }
+
+    if (dropItems != null) {
```
bruno-garcia: Keep the same order otherwise session end could come before sessions start

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeader.java
```diff
@@ -1,21 +1,30 @@
...
 
 import io.sentry.core.protocol.SentryId;
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-final class SentryEnvelopeHeader {
-  private final SentryId eventId;
-  private final String auth;
+public final class SentryEnvelopeHeader {
+  private final @NotNull SentryId eventId;
```
bruno-garcia:
```suggestion
  // Event Id must be set if the envelope holds an event, or an item that is related to the event (e.g: attachments, user feedback)
  private final @NotNull SentryId eventId;
```

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -123,6 +140,30 @@ public SentryId captureEvent(SentryEvent event, @Nullable Scope scope, @Nullable
...
+
+    if (session.getRelease() != null) {
```
```suggestion
    if (session.getRelease() == null) {
```

sentry-core/src/main/java/io/sentry/core/SentryClient.java
bruno-garcia: Agreed this makes sense to be in the Hub. Perhaps on both sides since this is public API now (users can call `client.captureSession(new Session)`. Ideally this wouldn't be public though, if we can yank it from the interface would be better

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -112,6 +113,22 @@ public SentryId captureEvent(SentryEvent event, @Nullable Scope scope, @Nullable
...
 
+    SentryEvent finalEvent = event;
+    scope.withSession(
+        s -> {
+          if (finalEvent.isCrashed()) {
+            if (s != null) {
+              s.addError();
+            }
+          }
+          if (finalEvent.getEnvironment() != null) {
```
bruno-garcia: Don't think we can do this. Session data can't change other than `state` and `error count`.
So I guess we can only close the session here and hope that the environment and release was set during session start. Which should be the case since guideline is to restart session if either is changed.

bruno-garcia: We don't want this to run if the event is cached (reading from cached events after app restart):

Move this stuff to line 86

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -355,4 +359,63 @@ public void removeExtra(@NotNull String key) {
   public void addEventProcessor(@NotNull EventProcessor eventProcessor) {
     eventProcessors.add(eventProcessor);
   }
+
+  // Atomic operations on session
+  public void withSession(@NotNull IWithSession sessionCallback) {
```
bruno-garcia: Ideally something like:

```suggestion
  @Api.Internal
  public void withSession(@NotNull IWithSession sessionCallback) {
```

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -148,13 +149,59 @@ public boolean isEnabled() {
...
+      try {
+        item.client.captureSession(pair.getCurrent());
+      } catch (IOException e) {
+        options
+            .getLogger()
+            .log(SentryLevel.ERROR, "Error while capturing session: " + e.getMessage(), e);
+      }
+      if (pair.getPrevious() != null) {
+        try {
+          item.client.captureSession(pair.getPrevious());
```
bruno-garcia: We could have `captureSession(Iterator<Session> sessions)` and capture both sessions at once.

sentry-core/src/main/java/io/sentry/core/transport/AsyncConnection.java
marandaneto: it doesn't matter as long as the sequence is right

bruno-garcia: That's right

PR #282, opened 2020-02-25
Enable the initialization using argument passing style in addition to the callback style configuration.

metlos:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Currently, Sentry only allows for callback-style initialization using `OptionsConfiguration` interface passed into the `Sentry.init()` methods. This PR makes available to users also the initialization method that merely accepts the `SentryOptions` instance.

Additionally, I've removed the unnecessary generic parameter on the `init` method that was made public, because it naturally accepts any subclass of `SentryOptions` just by inheritance rules.

The callback-based initialization is limiting in the injection-based environments like Spring or CDI where instances are produced and injected rather than modified through a callback.

N/A

## :pencil: Checklist

- [x] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Sentry.java
+++ b/sentry-core/src/main/java/io/sentry/core/Sentry.java
@@ -118,10 +118,9 @@ public final class Sentry {
    *
    * @param options options the SentryOptions
    * @param globalHubMode the globalHubMode
-   * @param <T> a class that extends SentryOptions or SentryOptions itself.
    */
-  private static synchronized <T extends SentryOptions> void init(
-      @NotNull T options, boolean globalHubMode) {
+  public static synchronized void init(
+      @NotNull SentryOptions options, boolean globalHubMode) {
     String dsn = options.getDsn();
     if (dsn == null) {
       throw new IllegalArgumentException("DSN is required. Use empty string to disable SDK.");


bruno-garcia:
@metlos if we do this there's no way anymore to run code on the object before we give the user a chance to change things. So to know in what is framework config vs user config becomes more complicated.
Is there a way to just give these frameworks the object we created instead of letting them give us the instance?

metlos:
> if we do this there's no way anymore to run code on the object before we give the user a chance to change things. So to know in what is framework config vs user config becomes more complicated.
Is there a way to just give these frameworks the object we created instead of letting them give us the instance?

So basically you need to distinguish between Sentry framework config vs user config? So what about just using 2 different objects for that?

bruno-garcia:
@metlos everything gets set to an instance of `SentryOptions` but if a user programatically sets a value, that must take presedence (override) anything set by some framework integration.

So if the frameworks read a JSON config and binding the values to `SentryOptions` we need that to happen before applying the user defined callback (programaticaly defined settings).

metlos:
Ok, let's keep this PR open and I'll try to come up with a reasonably "natural" of doing things in Spring that would not require this change.

marandaneto:
closed in favor of https://github.com/getsentry/sentry-android/blob/feat/sentry-java/sentry-core/src/main/java/io/sentry/core/Sentry.java#L127-L135

PR #291, opened 2020-03-04
apply scope's data if event is cached but handled

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
apply scope's data if event is cached but handled

right now events that are cached never apply the scope's data.
an event can be cached but handled, we'd like to apply scope's data.
an event can be cached, handled, but we shouldn't apply scope's data as the App. has been killed later on.

## :pencil: Checklist

- [ ] All tests passing


marandaneto:
in favor of #322

PR #299, opened 2020-03-13
ANR report should have 'was active=yes' on the dashboard

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
if ANR (was active)=yes on the dashboard.

the thread that reports ANRs is `ANRWatchDog`, so right now it's gonna be always `was active=no`.
as we have an `ExceptionMechanismException` that takes a `thread`, main thread in the case, we can fix this easily.

to discuss: because of the change in this PR, the selected thread on the thread selector (dashboard) will be wrong, as the thread that reported the ANR is not the thread that has a stack trace.

tests already exist.

## :pencil: Checklist

- [X] All tests passing


sentry-core/src/main/java/io/sentry/core/SentryThreadFactory.java
```diff
@@ -79,7 +82,8 @@ public SentryThreadFactory(final @NotNull SentryStackTraceFactory sentryStackTra
       }
 
       for (Map.Entry<Thread, StackTraceElement[]> item : threads.entrySet()) {
-        result.add(getSentryThread(currentThread, item.getValue(), item.getKey()));
+        result.add(
+            getSentryThread(currentThread, item.getValue(), item.getKey(), mechanismThreadIds));
```
bruno-garcia: I guess the only suggestion would be to remove this parameter and pass the decision, And lift the code that does the decision

```suggestion
            getSentryThread(isCrashed, item.getValue(), item.getKey()));
```

PR #302, opened 2020-03-17
automatic starting and ending session

marandaneto:


sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -96,6 +96,9 @@ public boolean isEnabled() {
                 session -> {
                   if (session != null) {
                     // if we do that on the client, session start will call also a session update
+                    // but I guess this should be in the client anyway, otherlise it will trigger
+                    // captureSession even if
+                    // its a normal event
                     item.client.captureSession(
```
bruno-garcia: What's the intent of capturing the session here? I thought we were persisting the session to disk when we start and end it.

Is it for the use case of incrementing error? In that case would it make more sense to couple the change to the session to flushing it? Since this event could have been a normal `captureMessage` for example.

marandaneto: @bruno-garcia the idea is to be NoOp if there's nothing relevant to the session.

We need to flush to the disk every change on the session (eg status, errorCount)

bruno-garcia: Right. But the changes to the session are done through the Hub, right?
Can we keep all the captureSession calls in there? It'll be easier to reason about this

marandaneto: I'll move this to other PR, It'd be nice to talk about it.

PR #320, opened 2020-03-24
eventId on envelope's header should be nullable

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
eventId on envelope's header should be nullable.

because an envelope only requires an eventId if there's an event and related items in it.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/SentryClient.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryClient.java
@@ -209,7 +209,7 @@ public final class SentryClient implements ISentryClient {
   }
 
   @Override
-  public @NotNull SentryId captureEnvelope(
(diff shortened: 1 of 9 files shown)


sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -209,7 +209,7 @@ public void captureSession(final @NotNull Session session, final @Nullable Objec
   }
 
   @Override
-  public @NotNull SentryId captureEnvelope(
+  public @Nullable SentryId captureEnvelope(
```
bruno-garcia: LGTM but I wonder if this makes more sense as `void` instead.

PR #322, opened 2020-03-25
ndk events apply scoped data

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
ndk events apply scoped data.

right now ndk events never apply scoped data, but it should.
fix #291

CI, tests

## :pencil: Checklist

- [X] All tests passing


sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
```diff
@@ -14,14 +15,17 @@
   private final String rootPath;
   private final IEnvelopeSender envelopeSender;
   private @NotNull final ILogger logger;
+  private final long timeout;
 
   // The preferred overload (Taking File instead of String) is only available from API 29
   @SuppressWarnings("deprecation")
-  EnvelopeFileObserver(String path, IEnvelopeSender envelopeSender, @NotNull ILogger logger) {
+  EnvelopeFileObserver(
+      String path, IEnvelopeSender envelopeSender, @NotNull ILogger logger, final long timeout) {
```
bruno-garcia:
```suggestion
      String path, IEnvelopeSender envelopeSender, @NotNull ILogger logger, final long flushTimeout) {
```

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
```diff
@@ -14,14 +15,17 @@
   private final String rootPath;
   private final IEnvelopeSender envelopeSender;
   private @NotNull final ILogger logger;
+  private final long timeout;
```
bruno-garcia:
```suggestion
  private final long flushTimeout;
```

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
```diff
@@ -14,14 +15,17 @@
...
   // The preferred overload (Taking File instead of String) is only available from API 29
   @SuppressWarnings("deprecation")
-  EnvelopeFileObserver(String path, IEnvelopeSender envelopeSender, @NotNull ILogger logger) {
+  EnvelopeFileObserver(
+      String path, IEnvelopeSender envelopeSender, @NotNull ILogger logger, final long timeout) {
     super(path);
     this.rootPath = Objects.requireNonNull(path, "File path is required.");
     this.envelopeSender = Objects.requireNonNull(envelopeSender, "Envelope sender is required.");
     this.logger = Objects.requireNonNull(logger, "Logger is required.");
+    this.timeout = timeout;
```
bruno-garcia:
```suggestion
    this.flushTimeout = flushTimeout;
```

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
```diff
@@ -39,6 +43,7 @@ public void onEvent(int eventType, @Nullable String relativePath) {
 
     // TODO: Only some event types should be pass through?
 
-    envelopeSender.processEnvelopeFile(this.rootPath + File.separator + relativePath);
+    final CachedEnvelopeHint hint = new CachedEnvelopeHint(timeout, logger);
```
bruno-garcia: Wasn't the lack of this hint that made events captured by this class get the scope applied?

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
```diff
@@ -33,32 +32,29 @@
...
-  private final long timeout;
 
   public EnvelopeSender(
       final @NotNull IHub hub,
       final @NotNull IEnvelopeReader envelopeReader,
       final @NotNull ISerializer serializer,
       final @NotNull ILogger logger,
       final long timeout) {
-    super(logger);
+    super(logger, timeout);
```
bruno-garcia: Could you please rename this timeout to `flushTimeout`.

sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
```diff
@@ -78,4 +79,8 @@ private void processNonCachedEvent(SentryEvent event) {
       event.setThreads(sentryThreadFactory.getCurrentThreads());
     }
   }
+
+  private boolean shouldApplyScopeData(final @Nullable Object hint) {
+    return (!(hint instanceof Cached) || (hint instanceof ApplyScopeData));
```
bruno-garcia: Why isn't simply not `Cached` events get scoped applied? Do we have cases where it's cached but shouldn't apply. Is cached and should apply. Isn't cached but should apply and vice versa?

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -65,14 +66,18 @@ public SentryClient(final @NotNull SentryOptions options, @Nullable Connection c
     random = options.getSampleRate() == null ? null : new Random();
   }
 
+  private boolean shouldApplyScopeData(final @Nullable Object hint) {
+    return (!(hint instanceof Cached) || (hint instanceof ApplyScopeData));
```
bruno-garcia: This method being duplicated is a bit worrying since the rule of what should "applyScopeData" could change in the future and we won't realize there are two methods guarding the hint checks

sentry-core/src/main/java/io/sentry/core/hints/Resettable.java
```diff
@@ -0,0 +1,7 @@
+package io.sentry.core.hints;
+
+/** marker interface that resets/restarts a Flushable operation. */
+public interface Resettable {
```
bruno-garcia: Does this need to be public? It's not really `reset a flushable operation` as it's reset fields of the hint so I can use the instance again. It was meant only to be used in a private class as an optimization to no `new` up a new instance for each file.

Perhaps better to delete this hint and create a new hint then to reduce complexity.

sentry-core/src/main/java/io/sentry/core/hints/CachedEnvelopeHint.java
```diff
@@ -0,0 +1,65 @@
...
+import io.sentry.core.ILogger;
+import io.sentry.core.util.Objects;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+@ApiStatus.Internal
+public final class CachedEnvelopeHint
+    implements Cached, Retryable, SubmissionResult, Flushable, Resettable, ApplyScopeData {
```
bruno-garcia: Hints before were two private classes implementing 3 interfaces.
I believe it got a lot more complex right now. Public classes, this one implementing 7 interfaces?

Is this whole refactor needed? If NDK events didn't get Scope applied can we find a simpler way? Like didn't the `Cached` hint alone control that?
What did `Cached` control? Maybe we can replace it with something explicit as u added: `ApplyScopeData` if that was the sole job of `Cached`?

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
marandaneto: `Cached` means it's being read from disk, but still doesn't mean we should or should not apply scope's data.
App. is alive and it should apply scope's data, that's why the combination of being Cached + ApplyScopeData.
this Hint could be an internal class of EnvelopeFileObserver

sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
marandaneto: this is how it was and this is actually the problem.
a captureMessage from NDK is a `Cached` hint, but still should apply scope's data, App is alive, so I composed `ApplyScopeData`

a captureException from NDK is the same case above.

now let's say there's a hard crash and App has died.
a captureException has a Cached hint and it should not apply scope's data, App has died and restarted.

sentry-core/src/main/java/io/sentry/core/SentryClient.java
marandaneto: agreed, I thought to move to a static helper or something, I've written unit tests for both anyway, still, can work on it.

sentry-core/src/main/java/io/sentry/core/hints/CachedEnvelopeHint.java
marandaneto: they are not private because they are organized in different packages, but there's a `@ApiStatus.Internal` annotation.

as this class doesn't belong to the class anymore, `interfaces` were needed to call respective methods like `waitFlush`.

Only the `Cached` hint was not possible to fix the issue if not changing way more things.

`Cached` is also used for storing files on disk again or not.

sentry-core/src/main/java/io/sentry/core/hints/Resettable.java
marandaneto: yeah bad written, my bad.
all the hints are under the `hints` package and as you know, java is like this.

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
bruno-garcia: What does `Cached` alone control?

sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
bruno-garcia: Got it. I just don't understand what's the use-case for Cached alone then.
To avoid caching the file again I guess?

sentry-core/src/main/java/io/sentry/core/hints/Resettable.java
bruno-garcia: Yeah makes sense. I guess we can manage to get rid of this one then at least

sentry-core/src/main/java/io/sentry/core/hints/CachedEnvelopeHint.java
bruno-garcia: > Cached is also used for storing files on disk again or not.

Right, I see.
So seems like we need all of them except `Resettable`, right?

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
marandaneto: `DefaultAndroidEventProcessor` needs it to not overwrite data (eg proguard Id, if App has restarted, we don't want to take the new ID) - similar to scope's data.

`MainEventProcessor` and `SentryClient` as you already know, apply or not scope's data.

`AsyncConnection` needs it to set a `NoOp` for `IEventCache` and `IEnvelopeCache`, so it's not needed to cache it again.

sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
marandaneto: the answer above mentions that

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
bruno-garcia: OK so we'll keep both: 
*  `Cached` don't store this on disk, it's already from disk
*  `ApplyScope`: apply scope 

that makes sense, thanks for clarifying

PR #323, opened 2020-03-25
NPE check so handled flag won't blow up an exception

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

NPE check so handled flag won't blow up an exception

as handled is not a primitive type, we need to NPE check.

## :pencil: Checklist

- [X] All tests passing


bruno-garcia:
good catch

PR #331, opened 2020-03-31
logging envelopes path when possible instead of nullable id

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
logging envelopes path when possible instead of nullable id

envelopes that don't have an `id` would log as `null` which is not useful at all.

running it.

## :pencil: Checklist

- [X] All tests passing


sentry-core/src/test/java/io/sentry/core/cache/SessionCacheTest.kt
```diff
@@ -177,8 +177,6 @@ class SessionCacheTest {
     fun `when session start and current file already exist, close session and start a new one`() {
         val cache = fixture.getSUT()
 
-        val file = File(fixture.options.sessionsPath!!)
```
marandaneto: not used, fixing linting

PR #337, opened 2020-04-01
add auto session tracking flag

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
add auto session tracking flag

we have a single flag to control 2 behavior.
1 is to enable session tracking at all.
2 is to enable the automatic session tracking integration.

it is a use case where one wants to start and end a session manually and to not do it automatically, this flag allows that.

unit test

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-sample/src/main/AndroidManifest.xml
+++ b/sentry-sample/src/main/AndroidManifest.xml
@@ -58,8 +58,8 @@
 <!--    how to enable session tracking-->
     <meta-data android:name="io.sentry.session-tracking.enable" android:value="true" />
 
-<!--    how to disable auto session tracking-->
-    <meta-data android:name="io.sentry.session-tracking.auto.enable" android:value="false" />
+<!--    how to enable auto session tracking integration-->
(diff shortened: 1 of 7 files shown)


sentry-sample/src/main/AndroidManifest.xml
```diff
@@ -58,6 +58,9 @@
 <!--    how to enable session tracking-->
     <meta-data android:name="io.sentry.session-tracking.enable" android:value="true" />
 
+<!--    how to enable auto session tracking integration-->
```
bruno-garcia, line 61: I'm not sure I follow what would be the use case for `io.sentry.session-tracking.enable`

bruno-garcia: In other words: Doesn't the new field replace the old one?

marandaneto:
in favor of #338

PR #340, opened 2020-04-04
Make SentryEvent.getThrowable() public.

metlos:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
`$TITLE`

Custom event processors might want to access it the same way the main event processor. It also makes testing easier/possible (to make sure that integrations set it properly).

NA

## :pencil: Checklist
- [x] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/SentryEvent.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryEvent.java
@@ -64,7 +64,7 @@ public final class SentryEvent implements IUnknownPropertiesConsumer {
     return (Date) timestamp.clone();
   }
 
-  Throwable getThrowable() {
+  public Throwable getThrowable() {
     return throwable;
   }
 


bruno-garcia:
It was made internal on purpose, because once we get this public, it's public API. We'd like to review the use cases before doing that and think this through since `SentryEvent` is supposed to hold the protocol and we're using it to hold extra data while processing the event.

It's useful (more so than just `SentryException`) since a user can have custom code to handle their own exception types. You could read the extra fields of your customer exception and augment the event.

What would be the use cases you already have in mind @metlos?

bruno-garcia:
@marandaneto: @dilbernd brought this up again. 

We either need to reconsider how we do things here (i.e: process the exception inside `captureException` like other SDKs) or make this method public.

I'd rather make this public. What do you folks thing?

dilbernd:
I don’t need it for prod code, I was just porting test code from @metlos into the new structure.

I’ll go a different route then. It was package private, I’ll just create a proxy in the test sources in the same package and give that a public getter.

If I have an actual direct need for it I’ll revisit is later but let’s put it back again for now.

bruno-garcia:
Sounds good. We'll close this PR then.
Thanks

PR #342, opened 2020-04-06
draft about distinct id implementation

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing


sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
```diff
@@ -0,0 +1,90 @@
+package io.sentry.android.core;
+
+import android.content.SharedPreferences;
+import io.sentry.core.IUserCache;
+import io.sentry.core.protocol.User;
+import io.sentry.core.util.Objects;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+// TODO: to think, this could be a SentryOptionsCache that could be expanded, does it make sense?
```
marandaneto, line 10: to discuss

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
```diff
@@ -0,0 +1,90 @@
...
+
+/** the Android User Cache class. It caches the user data in the Android SharedPreferences */
+public final class AndroidUserCache implements IUserCache {
+
+  // TODO: only these 3 fields make sense to me, otherwise caching a raw json would be better, but I
+  // don't want to do that.
```
marandaneto, line 16: to discuss

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
```diff
@@ -0,0 +1,90 @@
...
+      final @NotNull SharedPreferences sharedPreferences,
+      final @Nullable String sentryDeviceId) {
+    this.options = Objects.requireNonNull(options, "SentryAndroidOptions is required.");
+    this.sharedPreferences =
+        Objects.requireNonNull(sharedPreferences, "SharedPreferences is required.");
+    this.sentryDeviceId = sentryDeviceId;
+  }
+
+  //  TODO: I believe this makes sense only for global hub SDKs, but if not,
+  //   how do we cache multiple users (because we have multiple scopes)?
```
marandaneto, line 36: to discuss

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
```diff
@@ -0,0 +1,90 @@
...
+// TODO: to think, this could be a SentryOptionsCache that could be expanded, does it make sense?
+
+/** the Android User Cache class. It caches the user data in the Android SharedPreferences */
+public final class AndroidUserCache implements IUserCache {
+
+  // TODO: only these 3 fields make sense to me, otherwise caching a raw json would be better, but I
+  // don't want to do that.
+  private static final String EMAIL = "email";
+  private static final String ID = "id";
+  private static final String USERNAME = "user_name";
```
marandaneto, line 19: to discuss, for sessions, we only need these 3 fields

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
marandaneto, line 36: to discuss

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
```diff
@@ -0,0 +1,90 @@
...
+        user.setUsername(userName);
+      } else {
+        user.setId(sentryDeviceId);
+      }
+      return user;
+    } else {
+      final SharedPreferences.Editor edit = sharedPreferences.edit();
+      // is it a good idea to do it here? this guarantees that it cleans up on restart if the flag
+      // has been swapped
+      cleanUserFields(edit);
```
marandaneto, line 81: to discuss

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -402,9 +413,30 @@ SessionPair startSession() {
...
+        }
+      }
+
+      // fallback to sentryDeviceId
+      if (distinctId == null) {
+        // if I add a transient field (not serializable) to the User object, I can get rid of the
+        // distinctId on options
+        // I cannot simply set on user.Id cus its gonna be a breaking change
+        distinctId = options.getDistinctId();
+      }
```
marandaneto, line 436: to discuss

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -402,9 +413,30 @@ SessionPair startSession() {
       }
       previousSession = session;
 
-      session =
-          new Session(
-              options.getDistinctId(), user, options.getEnvironment(), options.getRelease());
+      String distinctId = null;
+
+      final User finalUser = getUser();
+      //      synchronized (userLock) {
```
marandaneto, line 419: it's racey if we don't sync it. but as session is already locked, only a copy of User would be fine.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -68,9 +73,24 @@ static void init(
     options.addIntegration(new AnrIntegration());
     options.addIntegration(new SessionTrackingIntegration());
 
-    readDefaultOptionValues(options, context);
+    final SharedPreferences sharedPreferences =
+        context.getSharedPreferences(BuildConfig.LIBRARY_PACKAGE_NAME, Context.MODE_PRIVATE);
+
+    String sentryDeviceId;
+    // cache it if not set
+    if (!sharedPreferences.contains(SENTRY_DEVICE_ID)) {
```
bruno-garcia, line 81: Is the place to have this code on some package private static class? If so we could just call it from here.

I believe this `contains` call could be replace with the `sharedPreferences.getString(SENTRY_DEVICE_ID, null) != null` to avoid two I/O calls.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
bruno-garcia: Makes sense to me to keep this focused on user caching, specially the abstraction

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
```diff
@@ -0,0 +1,90 @@
...
+
+  //  TODO: I believe this makes sense only for global hub SDKs, but if not,
+  //   how do we cache multiple users (because we have multiple scopes)?
+
+  @Override
+  public void setUser(final @Nullable User user) {
+    if (options.isCacheUserForSessions()) {
+      final SharedPreferences.Editor edit = sharedPreferences.edit();
+      if (user != null) {
+        edit.putString(ID, user.getId())
```
bruno-garcia, line 43: Could we settle for only caching the `userId` instead? Or at least a single value (when a user is set, we evaluate the which value based on an order (id>username>email)? Would be worth seeing what the issues page already considers when accounting for "user" of a crash.

If we're expecting to store the actual User, we'd need to consider the protocol can change (versioning/new fields added), [user define data](https://docs.sentry.io/development/sdk-dev/event-payloads/user/#attributes) and makes it trickier:

> All other keys are stored as extra information but not specifically processed by Sentry.

If we're storing just a single value to be used for session tracking only, it would minimize these edge cases.
Everything else you said stands true:

`setUser(...)` makes the SDK evaluate a new value for session `distrinctId`. `setUser(null)` deletes that cached value.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -16,6 +16,11 @@
   /** Enable or disable ANR on Debug mode Default is disabled Used by AnrIntegration */
   private boolean anrReportInDebug = false;
 
+  /**
+   * enable the usage of Settings.Secure.ANDROID_ID to keep retro compatible with older SDK versions
+   */
+  private boolean enableSecureAndroidId;
```
bruno-garcia, line 22: Could this option be named in such as way as to indicate a fallback (as opposed to `enable` which sounds like a feature).
Perhaps `useLegacyUserId` or something else?

Also, should this be guarded by `sendDefaultPii`? Meaning we'd need both this flag on + `sendDefaultPii`?

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -102,16 +105,24 @@ public void setTransaction(@Nullable String transaction) {
    * @return the user
    */
   public @Nullable User getUser() {
-    return user;
+    synchronized (userLock) {
+      if (user == null) {
```
bruno-garcia, line 109: A couple of potential improvements:

* double-check lock
* account for when no user is ever set. Now it'll take the lock and check the cache on each use.

sentry-core/src/main/java/io/sentry/core/Scope.java
bruno-garcia: Technically user would need to be readonly but yeah

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -402,9 +413,30 @@ SessionPair startSession() {
       }
       previousSession = session;
 
-      session =
-          new Session(
-              options.getDistinctId(), user, options.getEnvironment(), options.getRelease());
+      String distinctId = null;
+
+      final User finalUser = getUser();
```
bruno-garcia, line 418: Does this mean that now even if the user isn't cached we'll be user the user from the scope?
Seems like this brings us back to a potential double-user count per device issue.

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -402,9 +413,30 @@ SessionPair startSession() {
...
+        } else if (finalUser.getUsername() != null && !finalUser.getUsername().isEmpty()) {
+          distinctId = finalUser.getUsername();
+        }
+      }
+
+      // fallback to sentryDeviceId
+      if (distinctId == null) {
+        // if I add a transient field (not serializable) to the User object, I can get rid of the
+        // distinctId on options
+        // I cannot simply set on user.Id cus its gonna be a breaking change
```
bruno-garcia, line 434: This comment made me wonder if we should just release 3.0.0 with sessions and by default report the installation id as both `distinctId` and `user.id`. With that fallback to secureId you added already. Would simplify things?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
marandaneto: `SharedPreferences` does in-memory caching sync and it flushes to the disk async., so unless its value has changed, it's not an issue.
I'd separate it in a method or class, yes.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
marandaneto: yep, a single value would be ok, makes sense.
the order would be `id` > `email` > `username` if following the release health specs.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
marandaneto: agreed, naming here can be improved, `useLegacyAndroidId` maybe? the behavior is very dependent on it.
so initially I've added `sendDefaultPii` and later removed. as it's the only field to be guarded, it sounds redundant, one would need to swap 2 flags for a single behavior change.

sentry-core/src/main/java/io/sentry/core/Scope.java
marandaneto: I've noticed the multiple checks of `getUser` call.
Ideally, we'd only check it again if the value has been changed.
Definitely improvements to be done.

sentry-core/src/main/java/io/sentry/core/Scope.java
marandaneto: > Does this mean that now even if the user isn't cached we'll be user the user from the scope?

mm guess I don't get it, what do you mean?

> Seems like this brings us back to a potential double-user count per device issue.
 
this is true only on the first run of the App., if `cacheUserForSessions` is enabled and value has been set, it's gonna reuse the same distinctId, unless setUser has been called again with different data.

sentry-core/src/main/java/io/sentry/core/Scope.java
marandaneto: if we do a major bump, I'd remove `secureAndroidId` once for all.
I'd set `installationId` by default onto `Scope.user.id` on SDK init if `cacheUserForSessions` is disabled and if enabled, the cached `userId`.

this would make `event.user.id` and `session.distinctId` normalized.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
```diff
@@ -0,0 +1,90 @@
...
+import org.jetbrains.annotations.Nullable;
+
+// TODO: to think, this could be a SentryOptionsCache that could be expanded, does it make sense?
+
+/** the Android User Cache class. It caches the user data in the Android SharedPreferences */
+public final class AndroidUserCache implements IUserCache {
+
+  // TODO: only these 3 fields make sense to me, otherwise caching a raw json would be better, but I
+  // don't want to do that.
+  private static final String EMAIL = "email";
```
philipphofmann, line 17: Shouldn't we prefix those constants here? Could it be that the users of our SDK also save email to their shared preferences? What about 'sentry_user_cache_email' or something similar?

marandaneto: its a dedicated SharedPreferences, see:
https://github.com/getsentry/sentry-android/pull/342/files#diff-5b8dbf3e5fc61d592bc69f90d2a3b53cR76-R77

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
```diff
@@ -0,0 +1,90 @@
...
+  //   how do we cache multiple users (because we have multiple scopes)?
+
+  @Override
+  public void setUser(final @Nullable User user) {
+    if (options.isCacheUserForSessions()) {
+      final SharedPreferences.Editor edit = sharedPreferences.edit();
+      if (user != null) {
+        edit.putString(ID, user.getId())
+            .putString(EMAIL, user.getEmail())
+            .putString(USERNAME, user.getUsername())
```
philipphofmann, line 45: For me it is a bit confusing why we pass the whole user to `IUserCache.setUser` and only save id, email and username. Why don't we save the other users properties?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
philipphofmann: I'm not sure if Cache is the right naming here, because we actually persist the data to the `SharedPreferences`. A cache is usually used to improve the performance of something, see CPU caches, browser caches, web server caches. For me it is more like a `UserRepo` or `UserStorage`. What do you think?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
```diff
@@ -0,0 +1,90 @@
...
+        Objects.requireNonNull(sharedPreferences, "SharedPreferences is required.");
+    this.sentryDeviceId = sentryDeviceId;
+  }
+
+  //  TODO: I believe this makes sense only for global hub SDKs, but if not,
+  //   how do we cache multiple users (because we have multiple scopes)?
+
+  @Override
+  public void setUser(final @Nullable User user) {
+    if (options.isCacheUserForSessions()) {
```
philipphofmann, line 40: I think we could separate the logic of whether the user data should be saved or nor from the actual Android code. I would imagine two classes. One called something like `UserCache` that holds `SentryAndroidOptions` and `SentryDeviceId` and this class task to a simple interface like `UserPersistance`, that has an Android implementation an actually saves the user's data to the `SharedPreferences`. Then we have the SDK logic separated from persisting the data to Android. It would be easier to test and easier to extend. What do you think?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
philipphofmann: I think it makes sense, because otherwise you could receive an old user. Imagine the following scenario:
1. `setUser(A)`
2. `cacheUserForSessions = false`
3. `setUser(B)`, which doesn't do anything.
4. `cacheUserForSessions = true`
5. `getUser()` returns `A`

sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
```diff
@@ -45,4 +46,24 @@ static String getVersionCode(final @NotNull PackageInfo packageInfo) {
...
+   */
+  @SuppressWarnings("HardwareIds")
+  static @Nullable String getAndroidId(
+      final @NotNull Context context, final @Nullable String defaultValue) {
+    String androidId =
+        Settings.Secure.getString(context.getContentResolver(), Settings.Secure.ANDROID_ID);
+
+    if (androidId == null
+        || androidId.isEmpty()
+        || androidId.equalsIgnoreCase("9774d56d682e549c")) {
```
philipphofmann, line 64: Could we maybe add a comment about what `9774d56d682e549c` is? I had to Google it ;)

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -402,9 +413,30 @@ SessionPair startSession() {
...
-          new Session(
-              options.getDistinctId(), user, options.getEnvironment(), options.getRelease());
+      String distinctId = null;
+
+      final User finalUser = getUser();
+      //      synchronized (userLock) {
+      if (finalUser != null) {
+        if (finalUser.getId() != null && !finalUser.getId().isEmpty()) {
+          distinctId = finalUser.getId();
+        } else if (finalUser.getEmail() != null && !finalUser.getEmail().isEmpty()) {
```
philipphofmann, line 423: Why do we use the `email` or the `username` as the `distinctId` if the `id` is not set?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
marandaneto: not entirely, I borrowed the idea from https://developer.android.com/reference/android/content/Context.html#getCacheDir()

I see the Caching term as something temporary, being in memory or in the disk.
Also, https://docs.sentry.io/clients/java/integrations/#android (old SDK version)

> Events will be buffered to disk (in the application’s cache directory) by default

I'm fine changing it as well, but it might make it inconsistent with other SDKs.

@bruno-garcia opnions?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
marandaneto: yep, this one is Android-specific, but we can't forget that `sentry-core` is cross-platform.
nothing to do here for now.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
marandaneto: yep, that was exactly the use-case

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
marandaneto: yep, we could do that, just not sure how deep this abstraction should be, do we really see this user caching thing going forward with other SDKs? like java-backend, java-desktop etc... otherwise, it's just more classes for a single block of code, which is an `if`.
@bruno-garcia  opinions?

sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
marandaneto: ups, copy-pasta mistake.
the original code has a comment which leads to the google blogpost
see https://github.com/getsentry/sentry-android/blob/19aaac5ada0d3671501b0959db59437173db96ee/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java#L869

sentry-core/src/main/java/io/sentry/core/Scope.java
marandaneto: because all of them are considered to be a unique identifier, in this order.
if there's no id, check for email, no email as well? username.

marandaneto:
> I added a few comments :)

thanks for the review :)
I've replied all of them o/

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
bruno-garcia: It could be pulled out from the Android code but I believe it would only make sense then in desktop apps, right? Since this User in server apps are usually one per request and doesn't really make sense to cache.
Tbh we're not even sure this feature will even be used, it's just a work around to support `distinctId` for sessions based on user defined `user id`.

The proposal for `distinctId` has an option of not dealing with this at all (using only the generated id) which sounds like a good first version until we can get customer feedback to make a more educated decision.

sentry-core/src/main/java/io/sentry/core/Scope.java
bruno-garcia: This was the approach taken in the session supported added in the python SDK at least.

sentry-core/src/main/java/io/sentry/core/Scope.java
bruno-garcia: That makes sense. Also because a btter fallback is to tell users to call secure.AndroidId and set to the user.id by themselves

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
philipphofmann: I just came up with another scenario:
1. `setUser(A)`
2. cacheUserForSessions = false
3. cacheUserForSessions = true
4. getUser() returns A

In this scenario, even with calling `cleanUserFields` in `setUser` and `getUser`, you get the old user A. `SentryOptions.setCacheUserForSessions` could clear the `UserCache` if set to `false` to solve this problem.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
philipphofmann: Maybe then we should create an extra class for this and name it something like `SessionIdCache` and add this as a comment?

sentry-core/src/main/java/io/sentry/core/Scope.java
philipphofmann: Thanks for clarifying that 👍. Didn't know

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
marandaneto: yeah, I'd say on SDK. init, if the flag is `false`, we should clean this up.
SentryOptions fields are only mutable on SDK init., but not on runtime.
good catch

sentry-android-core/src/main/java/io/sentry/android/core/AndroidUserCache.java
marandaneto: I believe the Caching class should not be aware of what `id` it is, by that I mean (id, email or username).
maybe this class should take a String nominated as `distinctId` and that's all, so the naming would not `matter`.
A session also has an `id` so `SessionIdCache` would not fit, as we are caching the `Session.distinctId` which is a unique identifier of the user or device.
`UserIdCache` would make sense though?! naming is hard hehe

bruno-garcia: > maybe this class should take a String nominated as distinctId

I prefer this too. It says what it does. So when setting the user, if this option is enabled, we select the data from the user object (id, username, email) to become the distinctId and cache that so the caching itself isn't aware of this logic and we don't have to deal with caching the actual user object and possible user-defined fields.

PR #347, opened 2020-04-07
change integration order for cached outbox events

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
change integration order for cached outbox events.

as sentry-native move files around on init. and the outbox file observer now is inited after the NDK integration.
This integration also should be executed afterwards, so hard crashes will be in place and ready to be sent.
I've needed to make DirectoryProcessor public.

running and testing it.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -3,8 +3,10 @@ package io.sentry.android.core;
 import android.content.Context;
 import android.content.pm.PackageInfo;
 import io.sentry.core.EnvelopeReader;
+import io.sentry.core.EnvelopeSender;
 import io.sentry.core.IEnvelopeReader;
 import io.sentry.core.ILogger;
+import io.sentry.core.SendCachedEventFireAndForgetIntegration;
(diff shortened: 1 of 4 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -65,6 +67,33 @@ static void init(
     // because sentry-native move files around and we don't want to watch that.
     options.addIntegration(new NdkIntegration());
     options.addIntegration(EnvelopeFileObserverIntegration.getOutboxFileObserver(envelopeReader));
+
+    // Send cached envelopes from outbox path
+    // this should be executed after NdkIntegration because sentry-native move files on init.
+    // and we'd like to send them right away
+    options.addIntegration(
+        new SendCachedEventFireAndForgetIntegration(
```
bruno-garcia, line 75: One improvement is to pull this code into a its own class and adding here `new Integration()`
This would simplify this method a bit

marandaneto: yep, could be done for the others too then

PR #352, opened 2020-04-14
session should be exited state if state is ok

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
a session should be ended with an exited state.

the session was being ended always with abnormal if errorCount was > 0.
if there's an open session that should not be there and there's no ndk marker file, then it should be abnormal.

unit test

## :pencil: Checklist

- [X] All tests passing


marandaneto:
> Needs a test where status is OK and errors.count > 1?

don't think so, `errorCount` doens't affect the `state` field at all.

PR #353, opened 2020-04-14
envelope has dedicated endpoint

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
envelope has a dedicated endpoint.

@jan-auer can tell more about it.

fixed and ran unit tests.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
+++ b/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
@@ -18,7 +18,10 @@ import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
 import java.net.Proxy;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.util.Date;
@@ -47,7 +50,8 @@ public class HttpTransport implements ITransport {
   private final int connectionTimeout;
   private final int readTimeout;
   private final boolean bypassSecurity;
-  private final @NotNull URL sentryUrl;
+  private final @NotNull URL storeUrl;
+  private final @NotNull URL envelopeUrl;
   private final @NotNull SentryOptions options;
 
   private final @NotNull Map<String, Date> sentryRetryAfterLimit = new ConcurrentHashMap<>();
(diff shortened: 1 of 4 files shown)


sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -81,15 +85,26 @@ public HttpTransport(
     this.readTimeout = readTimeoutMillis;
     this.options = options;
     this.bypassSecurity = bypassSecurity;
-    this.sentryUrl = sentryUrl;
+
+    try {
+      final URI uri = sentryUrl.toURI();
```
marandaneto, line 90: @bruno-garcia ideally I'd get this `URI` in the ctor instead of the current `URL`, but I'd not like to break the public API, I think this would be fine here.
what do you think?

bruno-garcia: yeah this looks good for now but worth adding a task for v3.0.0 to refactor this. Perhaps we need to revisit the `Dsn` type (like add a `toStore` and `toEnvelope` or something completely different).
Such discussion for sure woudl involve the whole SDK team

sentry-core/src/test/java/io/sentry/core/DsnTest.kt
```diff
@@ -11,7 +11,7 @@ class DsnTest {
     fun `dsn parsed with path, sets all properties`() {
         val dsn = Dsn("https://publicKey:secretKey@host/path/id")
 
-        assertEquals("https://host/path/api/id/store/", dsn.sentryUri.toURL().toString())
+        assertEquals("https://host/path/api/id", dsn.sentryUri.toURL().toString())
```
bruno-garcia, line 14: This makes me realize this is now a breaking change since if one relies on "DSN" to make requests, that'll be broken.

marandaneto: only if they have their own `Transport` which is discouraged, but possible anyway.
I'll point this out on the release notes, good catch.

marandaneto:
> We can go ahead with this but there's one comment we ideally would need to discuss

yeah but `toStore` may get deprecated very soon as well and we'd even be able to come back to the original behavior but using `/envelope` lol.
the problem is the `AsyncConnection` and you know that :D

PR #356, opened 2020-04-15
check for androidx availability on runtime

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
check for androidx availability on runtime

if you don't want to use the release health feature, you actually can exclude it.
just to avoid an error if you have enabled it and can't use androidx, we check on runtime via reflection.

running it/build.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SessionTrackingIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SessionTrackingIntegration.java
@@ -31,10 +31,21 @@ public final class SessionTrackingIntegration implements Integration, Closeable 
             options.isEnableSessionTracking());
 
     if (options.isEnableSessionTracking()) {
-      watcher = new LifecycleWatcher(hub, options.getSessionTrackingIntervalMillis());
-      ProcessLifecycleOwner.get().getLifecycle().addObserver(watcher);
+      try {
+        Class.forName("androidx.lifecycle.DefaultLifecycleObserver");
+        Class.forName("androidx.lifecycle.ProcessLifecycleOwner");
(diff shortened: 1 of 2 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SessionTrackingIntegration.java
```diff
@@ -31,10 +31,21 @@ public void register(final @NotNull IHub hub, final @NotNull SentryOptions optio
             options.isEnableSessionTracking());
 
     if (options.isEnableSessionTracking()) {
-      watcher = new LifecycleWatcher(hub, options.getSessionTrackingIntervalMillis());
-      ProcessLifecycleOwner.get().getLifecycle().addObserver(watcher);
-
-      options.getLogger().log(SentryLevel.DEBUG, "SessionTrackingIntegration installed.");
+      try {
```
bruno-garcia, line 34: This is a nice feature but is it worth the cost of using reflection?

marandaneto: it's necessary if release health is enabled and androidx is not available, it's gonna crash.
this could be fixed by the clients on development, of course, but it requires code on their end and it makes the one-liner SDK integration a lie due to their restrictions.
ideally, we'd extract this as a separate maven package, as this would be needed for automatic breadcrumbs too, it's a temporary solution though.
they would work as the sentry ndk module, where reflection is necessary due to being sibling's modules.

bruno-garcia: sounds good

PR #362, opened 2020-04-16
update current session on session start after deleting previous session

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
update current session on session start after deleting the previous session.

when starting a new session, if there's a leftover session, we end it and start a new one, even in this case, we should flush the new session to the current session file.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/cache/SessionCache.java
+++ b/sentry-core/src/main/java/io/sentry/core/cache/SessionCache.java
@@ -153,9 +153,8 @@ public final class SessionCache implements IEnvelopeCache {
         } catch (IOException e) {
           options.getLogger().log(SentryLevel.ERROR, "Error processing session.", e);
         }
-      } else {
-        updateCurrentSession(currentSessionFile, envelope);
       }
+      updateCurrentSession(currentSessionFile, envelope);
     }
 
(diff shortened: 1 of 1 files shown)


sentry-core/src/main/java/io/sentry/core/cache/SessionCache.java
```diff
@@ -153,9 +153,12 @@ public void store(final @NotNull SentryEnvelope envelope, final @Nullable Object
         } catch (IOException e) {
           options.getLogger().log(SentryLevel.ERROR, "Error processing session.", e);
         }
-      } else {
-        updateCurrentSession(currentSessionFile, envelope);
+
+        if (!currentSessionFile.delete()) {
```
marandaneto, line 157: at this point the leftover session and its current session file already became a new envelope file to be sent, so deleting it as the new session will take place.

bruno-garcia, line 157:
```suggestion
        // at this point the leftover session and its current session file already became a new envelope file to be sent
        // so deleting it as the new session will take place.
        if (!currentSessionFile.delete()) {
```

Issue #341, opened 2020-04-06
UnsatisfiedLinkError in SentryNdk


tobi512:
Hi there, 
we recently upgraded Sentry from v1.7.23 to v2.0.2 and so far everything looked good. Now we published the first version via our beta channel in Google Play and received a strange crash related to Sentry's new native crash handling feature. 

**Stacktrace:**
```
java.lang.UnsatisfiedLinkError: 
  at java.lang.Runtime.loadLibrary0 (Runtime.java:1016)
  at java.lang.System.loadLibrary (System.java:1657)
  at io.sentry.android.ndk.SentryNdk.<clinit> (SentryNdk.java:12)
  at java.lang.Class.classForName (Class.java)
  at java.lang.Class.forName (Class.java:453)
  at java.lang.Class.forName (Class.java:378)
  at io.sentry.android.core.NdkIntegration.register (NdkIntegration.java:26)
  at io.sentry.core.Hub.<init> (Hub.java:35)
  at io.sentry.core.Sentry.init (Sentry.java:143)
  at io.sentry.core.Sentry.init (Sentry.java:91)
  at io.sentry.android.core.SentryAndroid.init (SentryAndroid.java:59)
  at io.sentry.android.core.SentryAndroid.init (SentryAndroid.java:44)
```

**Affected device:**
Samsung Galaxy J7 Prime (on7xelte), 2816MB RAM, Android 8.1

Right now, we only have one user suffering from the crash according to the Play Console, but from our experience, the number of users will increase as soon as we publish the update to everyone. We've been testing the app on a couple of devices and it seems to run fine so far (however most of the weird crashes happen out in the wild as you probably know).

Is this a known issue and do you have any idea on how to fix it?
Disabling the NDK crash feature by using `setEnableNdk(false)` might be an option, but the feature is quite nice, so we'd like to keep that enabled.

Regards, 
Tobias

marandaneto:
hey @tobi512 thanks for rising this issue.

actually it's the first `UnsatisfiedLinkError` report and not sure what this is about.
We do have a `try/catch` block when loading the native libs, but it seems to not be enough.
I'll try to investigate a bit deeper on what could cause that.

A few questions to understand a bit more:

Do you have Proguard/R8 enabled?
Do you strip native libs? (separated APK for each CPU architecture)
Do you have multidex enabled?
Do you use App bundles?

if you decide to disable it, for now, you can save the native bits by doing this:
[...]

PR #373, opened 2020-04-22
fix: Merge loadLibrary calls for sentry-native and clean up CMake files

Swatinem:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This loads all the non-essential-system libraries explicitly in order, to hopefully avoid dynamic linker errors such as #341 

## :pencil: Checklist

- [ ] All tests passing

### DIFF
--- a/sentry-sample/src/main/java/io/sentry/sample/NativeSample.java
+++ b/sentry-sample/src/main/java/io/sentry/sample/NativeSample.java
@@ -6,6 +6,8 @@ public class NativeSample {
   public static native void message();
 
   static {
+    System.loadLibrary("log");
+    System.loadLibrary("sentry");
     System.loadLibrary("native-sample");
   }
 }
--- a/sentry-sample/src/main/java/io/sentry/sample/NativeSample.java
+++ b/sentry-sample/src/main/java/io/sentry/sample/NativeSample.java
@@ -6,8 +6,6 @@ public class NativeSample {
   public static native void message();
 
   static {
-    System.loadLibrary("log");
-    System.loadLibrary("sentry");
     System.loadLibrary("native-sample");
   }
 }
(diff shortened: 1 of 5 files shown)


Swatinem:
Don’t get your hopes up to soon. I have no idea if this actually fixes the problem :-D

bruno-garcia:
lol. Well, worth the shot

sentry-sample/src/main/java/io/sentry/sample/NativeSample.java
```diff
@@ -6,6 +6,8 @@
   public static native void message();
 
   static {
+    System.loadLibrary("log");
+    System.loadLibrary("sentry");
```
marandaneto, line 10: @Swatinem why do we need this here?
this should be handled exclusively by `NdkIntegration` and clients should not need extra code.

marandaneto:
> Don’t get your hopes up to soon. I have no idea if this actually fixes the problem :-D

yeah its a try :) thx for that, left a comment btw.

sentry-sample/src/main/java/io/sentry/sample/NativeSample.java
marandaneto: it'd work without this btw, so I think its just a copypasta and can be removed, right?

Swatinem: I guess it can? Is the execution order of these `static` blocks guaranteed in some way?

marandaneto: @Swatinem this is the sample and not part of the SDK -> `sentry-sample` package.
`SentryNdk` is the class responsible for initing sentry-native.
not sure why you've have added these 2 lines here, mind explaining?

Swatinem: well the sample only loads its own library, so I thought I load the transitive dependencies as well. I guess this is irrelevant when sentry-android runs *first* to load them.

marandaneto: yep, exactly

PR #377, opened 2020-04-24
if retry after header has empty categories, apply retry after to all of them

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
if retry after header has empty categories, apply retry after to all of them.

the current implementation was not handling empty categories.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
+++ b/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
@@ -267,7 +267,9 @@ public class HttpTransport implements ITransport {
     // X-Sentry-Rate-Limits looks like: seconds:categories:scope
     // it could have more than one scope so it looks like:
     // quota_limit, quota_limit, quota_limit
+
     // a real example: 50:transaction:key, 2700:default;event;security:organization
+    // 50::key is also a valid case, it means no categories and it should apply to all of them
     final String sentryRateLimitHeader = connection.getHeaderField("X-Sentry-Rate-Limits");
     updateRetryAfterLimits(sentryRateLimitHeader, retryAfterHeader, responseCode);
   }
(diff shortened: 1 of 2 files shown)


sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -292,14 +301,20 @@ private void updateRetryAfterLimits(
...
 
               for (final String catItem : categories) {
                 // we dont care if Date is UTC as we just add the relative seconds
                 sentryRetryAfterLimit.put(
                     catItem, new Date(System.currentTimeMillis() + retryAfterMillis));
               }
+            } else {
+              // if categories are empty, we should apply to all the categories.
+              // we do that using a `default` category as the fallback
+              sentryRetryAfterLimit.clear();
```
marandaneto, line 315: another way would be to loop all the items of sentryRetryAfterLimit and replace them.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -260,7 +261,15 @@ public boolean isRetryAfter(final @NotNull String type) {
...
       final @NotNull HttpURLConnection connection, final int responseCode) {
+    // milliseconds
     final String retryAfterHeader = connection.getHeaderField("Retry-After");
+
+    // X-Sentry-Rate-Limits looks like: seconds:categories:scope
+    // it could have more than one scope so it looks like:
+    // quota_limit, quota_limit, quota_limit
+
+    // a real example: 50:transaction:key, 2700:default;event;security:organization
+    // 50::key is also a valid case, it means no categories and it should apply to all of them
```
bruno-garcia, line 272: Super helpful comment

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
bruno-garcia: This sounds better to me too

PR #381, opened 2020-04-30
fix: rate limiting categories

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
rate limiting categories

We were using envelope item types and it should be its own category.

## :pencil: Checklist

- [X] All tests passing


sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -161,16 +161,49 @@ public HttpTransport(
...
+    return false;
+  }
+
+  /**
+   * Returns a rate limiting category from type
+   *
+   * @param type the type (eg event, session, attachment, ...)
+   * @return the category eg (error, session, attachment)
+   */
+  private @NotNull String getCategoryFromType(final @NotNull String type) {
```
philipphofmann, line 190: I think we have two different mappings. One for `events` and one for `envelopeItems`. Is this still valid @jan-auer? If yes, can we name the method and parameter either `envelopeType` or `eventType` to avoid misinterpretation?

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -161,16 +161,49 @@ public HttpTransport(
...
+    // check all categories
+    if (sentryRetryAfterLimit.containsKey(HTTP_RETRY_ALL_CATEGORIES)) {
+      final Date date = sentryRetryAfterLimit.get(HTTP_RETRY_ALL_CATEGORIES);
+      if (!currentDate.after(date)) {
+        return true;
+      }
+    }
+    // check for specific category
+    if (sentryRetryAfterLimit.containsKey(category)) {
+      final Date date = sentryRetryAfterLimit.get(category);
```
bruno-garcia, line 176: Same here

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -161,16 +161,49 @@ public HttpTransport(
...
   @Override
   public boolean isRetryAfter(final @NotNull String type) {
-    if (sentryRetryAfterLimit.containsKey(type)) {
-      final Date date = sentryRetryAfterLimit.get(type);
+    final String category = getCategoryFromType(type);
+    final Date currentDate = new Date();
+
+    // check all categories
+    if (sentryRetryAfterLimit.containsKey(HTTP_RETRY_ALL_CATEGORIES)) {
+      final Date date = sentryRetryAfterLimit.get(HTTP_RETRY_ALL_CATEGORIES);
```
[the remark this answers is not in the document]
marandaneto: makes sense

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -161,18 +161,52 @@ public HttpTransport(
...
   }
 
+  /**
+   * Returns a rate limiting category from type
+   *
+   * @param type the type (eg event, session, attachment, ...)
+   * @return the category eg (error, session, attachment)
+   */
+  private @NotNull String getCategoryFromType(final @NotNull String type) {
+    // fallback everything else to 'default' category
```
philipphofmann, line 191: `h`: Hm the implementation on Cocoa is now different: https://github.com/getsentry/sentry-cocoa/blob/master/Sources/Sentry/SentryRateLimitCategoryMapper.m
Is this now the proper way now @jan-auer ?

philipphofmann: Okay I see the internal docs state :`default`: Events with an event_type not listed explicitly below.

marandaneto: I've added all the known types and used by either this or SDKs that bundle sentry-android, which can be sentry-react-native or sentry-native, so I believe this should be the correct implementation

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: I guess you are mixing up, event types or envelope types is a way to define the type of something, eg a crash event is the type of `event` and category `error`.
if a crash event is sent by an envelope or not, its still a `event` as a type and `error` as a category.

also, please use our logaf scale

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -161,16 +161,49 @@ public HttpTransport(
...
+
+  /**
+   * Returns a rate limiting category from type
+   *
+   * @param type the type (eg event, session, attachment, ...)
+   * @return the category eg (error, session, attachment)
+   */
+  private @NotNull String getCategoryFromType(final @NotNull String type) {
+    // fallback everything else to 'default' category
+    String category = "default";
```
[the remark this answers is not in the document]
marandaneto: this can be improved, yes.

but a downside of an enum in java is, it cant have a public ctor so it cant be used outside of the package and we do use packages for the organization.
We have `SentryEnvelopeItemType` already.

I'd improve it in another PR if thats ok

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -161,16 +161,49 @@ public HttpTransport(
    */
   @Override
   public boolean isRetryAfter(final @NotNull String type) {
-    if (sentryRetryAfterLimit.containsKey(type)) {
-      final Date date = sentryRetryAfterLimit.get(type);
+    final String category = getCategoryFromType(type);
+    final Date currentDate = new Date();
```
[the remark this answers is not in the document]
marandaneto: yes, agreed, in this case, it's a single comparison and used only once, not sure if it'd pay off.
do we have this in cocoa or its just a room you saw for improvement?

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -161,16 +161,49 @@ public HttpTransport(
...
+
+    // check all categories
+    if (sentryRetryAfterLimit.containsKey(HTTP_RETRY_ALL_CATEGORIES)) {
+      final Date date = sentryRetryAfterLimit.get(HTTP_RETRY_ALL_CATEGORIES);
+      if (!currentDate.after(date)) {
+        return true;
+      }
```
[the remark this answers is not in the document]
marandaneto: this would break the code, it'd return true or false immediately and it'd never check for the specific category.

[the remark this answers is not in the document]
philipphofmann: Whops yes it would sorry.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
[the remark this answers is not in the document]
philipphofmann: We created [`CurrentDateProvider`](https://github.com/getsentry/sentry-cocoa/blob/master/Sources/Sentry/SentryCurrentDate.m), which is static. Be aware that static has the downside that if you run your tests in parallel it might get nasty. So we might change it again and pass a `CurrentDateProvider` as a dependency or pass the date as a method parameter would also work. For this code here it is not a big deal if we send a few events 2 or 3 milliseconds earlier. For other code you might want to make sure, that you don't have a off-by-one error for milliseconds or seconds and that is impossible to test with `Thread.sleep()`.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -327,7 +400,23 @@ private void updateRetryAfterLimits(
...
+    }
+  }
+
+  /**
+   * apply new timestamp for rate limiting only if its longer than the previous one
+   *
+   * @param dataCategory the DataCategory
+   * @param date the Date to be applied
+   */
+  private void applyRetryAfterOnlyIfLonger(
```
philipphofmann, line 413: `l`: Didn't know this part. That's a reason why I like PRs 😜. We have to update this in Cocoa.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -41,6 +43,27 @@
...
+    private final String category;
+
+    DataCategory(final @NotNull String category) {
+      this.category = category;
+    }
+
+    public String getCategory() {
+      return category;
+    }
+  }
```
philipphofmann, line 65: `l`: Nice that we have an enum now for this 👏. I am a fan of nested classes or enums. They should only exist to serve their enclosing classes though. Are we sure that this enum only exists in combination with `HttpTransport`? If yes, I think we could even make it private. If not, we could consider making it an enum outside of `HttpTransport`.

sentry-core/src/test/java/io/sentry/core/util/StringUtilsTest.kt
```diff
@@ -29,4 +29,29 @@ class StringUtilsTest {
...
+
+    @Test
+    fun `capitalize returns itself if null`() {
+        assertNull(StringUtils.capitalize(null))
+    }
+
+    @Test
+    fun `capitalize returns itself if empty`() {
+        assertEquals("", StringUtils.capitalize(""))
+    }
```
philipphofmann, line 56: `l`: Nice tests 👍

sentry-core/src/main/java/io/sentry/core/util/StringUtils.java
```diff
@@ -19,4 +20,18 @@ private StringUtils() {}
     }
     return null;
   }
+
+  /**
+   * Returns a Capitalized String eg session = Session
```
philipphofmann, line 25: `l`: Maybe we could be a little bit more specific here. 
`Changes the first char to upper case and all remaining chars to lower case. eg seSSioN = Session`

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
```diff
@@ -297,6 +297,66 @@ class HttpTransportTest {
         assertTrue(transport.isRetryAfter("event"))
     }
 
+    @Test
+    fun `parse X-Sentry-Rate-Limit and ignore unknown categories`() {
+        val transport = fixture.getSUT()
+
+        whenever(fixture.connection.inputStream).thenThrow(IOException())
+        whenever(fixture.connection.getHeaderField(eq("X-Sentry-Rate-Limits")))
+            .thenReturn("60:default;wtf;error;security:organization")
```
philipphofmann, line 306: `m`: I'm not sure if keeping `wtf` in our code is a good idea. Can we maybe change it to `foobar`?

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
```diff
@@ -297,6 +297,66 @@ class HttpTransportTest {
...
+
+        val event = SentryEvent()
+
+        transport.send(event)
+        Thread.sleep(2000)
```
philipphofmann, line 311: `m`: I don't think we need this `Thread.sleep(2000)` here.

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
```diff
@@ -297,6 +297,66 @@ class HttpTransportTest {
...
+        assertFalse(transport.isRetryAfter("wtf"))
```
philipphofmann, line 312: `m`: Maybe it makes sense to assert that the other categories have a retry after active.
Furthermore, we are going away from `Retry-After` to `Rate-Limits`, maybe it makes sense to rename this method at some point, it's a little bit confusing for me.

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
```diff
@@ -297,6 +297,66 @@ class HttpTransportTest {
...
+        val transport = fixture.getSUT()
+
+        whenever(fixture.connection.inputStream).thenThrow(IOException())
+        whenever(fixture.connection.getHeaderField(eq("X-Sentry-Rate-Limits")))
+            .thenReturn("1::key, 60:default;error;security:organization")
+
+        val event = SentryEvent()
+
+        transport.send(event)
+        Thread.sleep(2000)
```
philipphofmann, line 326: `l`: As already pointed out testing with `Thread.sleep` is not the best idea. Imagine we want to test a maximum timeout of one day. How would we do that? Furthermore, this slows down the test suite a lot if we do it in multiple tests. We don't have to improve this in this PR, but maybe we do it at some point.

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
```diff
@@ -297,6 +297,66 @@ class HttpTransportTest {
...
+        assertTrue(transport.isRetryAfter("event"))
+    }
+
+    @Test
+    fun `When default category is present, apply for error category`() {
```
philipphofmann, line 331: I think the upper test is already testing if `default` applies to `error`. I think we could remove this test.

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
```diff
@@ -297,6 +297,66 @@ class HttpTransportTest {
...
+        assertTrue(transport.isRetryAfter("event"))
+    }
+
+    @Test
+    fun `When category has shorter rate limiting, do not apply new timestamp`() {
+        val transport = fixture.getSUT()
+
+        whenever(fixture.connection.inputStream).thenThrow(IOException())
+        whenever(fixture.connection.getHeaderField(eq("X-Sentry-Rate-Limits")))
+            .thenReturn("60:error:key, 1:error:organization")
```
philipphofmann, line 351: `h`: I think it would be nice to also test if a new timestamp is applied if the category has a longer rate-limiting. Something like this `"2:error:organization 3:error:key, 1:error:organization"` and then test that the rate limit is actually 3 seconds. Here a DateProvider with whom we could move the date 3 seconds would already be handy.

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeItemType.java
```diff
@@ -5,7 +5,14 @@
...
   Session("session"),
-  Event("event");
+  Event("event"), // DataCategory.Error
+  Attachment("attachment"),
+  Transaction("transaction"),
+  Csp("csp"), // DataCategory.Security
+  Hpkp("hpkp"), // DataCategory.Security
+  Expectct("expectct"), // DataCategory.Security
+  Expectstaple("expectstaple"), // DataCategory.Security
+  Unknown("__unknown__"); // DataCategory.Unknown
```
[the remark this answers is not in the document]
marandaneto: yes, `__unknown__` really doesn't exist.
while `DataCategory.Unknown` on the backend it does, I wanted to be sure that they don't overlap

as I have `DataCategory.All` which is `__all__`

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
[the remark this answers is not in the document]
marandaneto: yeah I agree, this should come from a ctor with package-private access modifier.
people implementing their own Transport should not know about it IMO.
I can improve that on another PR along with another `Date`s across the SDK.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: yes, private here makes sense, good catch.

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
marandaneto: lol
https://acronyms.thefreedictionary.com/WTF
the meaning is in the eyes of whom read it :)
will do it

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
marandaneto: good catch, copy pasta error

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
marandaneto: renaming would be a breaking change as Android is released like this since 2.0 GA.
but still it's a `retry after` approach. why would it not make sense?

sentry-core/src/main/java/io/sentry/core/util/StringUtils.java
marandaneto: will do it

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
marandaneto: https://github.com/getsentry/sentry-android/pull/381/files#r419898438

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
marandaneto: indirectly, yes, but I prefer an explicitly test for it.
a change can cause this test to be broken but not the above one. this guarantees that.

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
marandaneto: true, but this test is being done indirectly by all of the other tests.
I don't mind adding an extra one, but we don't have the DateProvider right now, can do next.

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeItemType.java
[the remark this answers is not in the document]
philipphofmann: Okay thanks.

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
philipphofmann: Because it's our custom rate limit approach and I think we just used the retry after header to implement rate-limiting in the first place. But it's not a big deal.

Issue #382, opened 2020-04-30
Provide a way to avoid installation of Sentry's default UncaughtExceptionHandlerIntegration


techyourchance:
It looks like Sentry always installs its own ExceptionHandler as default, even if I disable it in manifest and init in Application's onCreate(). As far as I can tell, there is no option to disable this functionality.

Use case:
I want to implement my own ExceptionHandler which will delegate to Android's default ExceptionHandler in some cases (without going through Sentry's one). However, since Sentry installs itself before I get a chance to grab Android's default, this use case can't be implemented.

What I tried:
I tried to iterate over Integrations list in Options when initializing and remove this integration manually, but got ConcurrentModificationException. Looks like Sentry iterates this list before the initialization is complete.

marandaneto:
hey @techyourchance thanks for raising this issue, we definitely could improve things here, but there's a way of doing it.

1st, you'd need to disable the auto-init mode that uses a ContentProvider.

`<meta-data android:name="io.sentry.auto-init" android:value="false" />`

2nd, you'd need to remove the `UncaughtExceptionHandlerIntegration` from the `integrations` list.

For that, you'd need to init Sentry manually:

```
    SentryAndroid.init(
      getApplicationContext(),
      options -> {
        for (final Iterator<Integration> iterator = options.getIntegrations().iterator(); iterator.hasNext(); ) {
          final Integration integration = iterator.next();
          if (integration instanceof UncaughtExceptionHandlerIntegration) {
            iterator.remove();
          }
        }
      });
```

The `ConcurrentModificationException` is not a bug, but a Java thing.

so this would work if you are removing an item from an ArrayList but using the correct approach, eg using an `iterator` as I did, using a `ListIterator`, Collect to an aux. list and remove it after iterating thru all the items, `Collection.removeIf` is minApi >= 24, etc...

I'm just guessing here that you've tried something like this:

```
        for (final Integration item : options.getIntegrations()) {
          if (item instanceof UncaughtExceptionHandlerIntegration) {
            options.getIntegrations().remove(item);
          }
        }
```

Basically you'd be removing the item that you are iterating and this would cause the issue.

We feel your pain as this would be the first try of everyone, we'll document it properly, thanks.

marandaneto:
PR #384 is open with a proposal

techyourchance:
Thanks for acknowledging this issue.
Quick note: 
[...]

PR #384, opened 2020-05-01
feat: Options for uncaught exception and make SentryOptions list Thread-Safe

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

#382

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/SentryOptions.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryOptions.java
@@ -5,8 +5,8 @@ import io.sentry.core.transport.ITransport;
 import io.sentry.core.transport.ITransportGate;
 import java.io.File;
 import java.net.Proxy;
-import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -22,13 +22,13 @@ public class SentryOptions {
    * Are callbacks that run for every event. They can either return a new event which in most cases
    * means just adding data OR return null in case the event will be dropped and not sent.
    */
-  private final @NotNull List<EventProcessor> eventProcessors = new ArrayList<>();
+  private final @NotNull List<EventProcessor> eventProcessors = new CopyOnWriteArrayList<>();
 
   /**
    * Code that provides middlewares, bindings or hooks into certain frameworks or environments,
    * along with code that inserts those bindings and activates them.
    */
-  private final @NotNull List<Integration> integrations = new ArrayList<>();
+  private final @NotNull List<Integration> integrations = new CopyOnWriteArrayList<>();
 
   /**
    * The DSN tells the SDK where to send the events to. If this value is not provided, the SDK will
@@ -128,13 +128,13 @@ public class SentryOptions {
    * packages. Modules considered not to be part of the app will be hidden from stack traces by
    * default.
    */
-  private final @NotNull List<String> inAppExcludes = new ArrayList<>();
+  private final @NotNull List<String> inAppExcludes = new CopyOnWriteArrayList<>();
 
   /**
    * A list of string prefixes of module names that belong to the app. This option takes precedence
    * over inAppExcludes.
    */
-  private final @NotNull List<String> inAppIncludes = new ArrayList<>();
+  private final @NotNull List<String> inAppIncludes = new CopyOnWriteArrayList<>();
 
   /** The transport is an internal construct of the client that abstracts away the event sending. */
   private @Nullable ITransport transport;
@@ -171,6 +171,11 @@ public class SentryOptions {
 
   /** The server name used in the Sentry messages. */
   private String serverName;
+
+  /*
+  When enabled, Sentry installs UncaughtExceptionHandlerIntegration.
+   */
+  private boolean enableUncaughtExceptionHandler = true;
 
   /**
    * Adds an event processor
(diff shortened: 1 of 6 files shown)


sentry-core/src/main/java/io/sentry/core/SentryOptions.java
```diff
@@ -128,13 +128,13 @@
...
    */
-  private final @NotNull List<String> inAppExcludes = new ArrayList<>();
+  private final @NotNull List<String> inAppExcludes = new CopyOnWriteArrayList<>();
 
   /**
    * A list of string prefixes of module names that belong to the app. This option takes precedence
    * over inAppExcludes.
    */
-  private final @NotNull List<String> inAppIncludes = new ArrayList<>();
+  private final @NotNull List<String> inAppIncludes = new CopyOnWriteArrayList<>();
```
bruno-garcia, line 137: m: We're not really making this class thread-safe just for changing this. This class wasn't designed to be access in multiple threads, and that's OK. The documented way is to create an instance, configure it, then call `Sentry.init` and not spawn a new thread to call `Sentry.init` while you mutate the options in concurrently.

If we decide to make this class thread-safe we'd need to make a pass across and make more changes.

So IMO it's not worth changing this type to `CopyOnWriteArrayList` which will allocate a new instance for each `add`.

sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
```diff
@@ -20,24 +21,24 @@
 public final class UncaughtExceptionHandlerIntegration
     implements Integration, Thread.UncaughtExceptionHandler, Closeable {
   /** Reference to the pre-existing uncaught exception handler. */
-  private Thread.UncaughtExceptionHandler defaultExceptionHandler;
+  private @Nullable Thread.UncaughtExceptionHandler defaultExceptionHandler;
 
-  private IHub hub;
-  private SentryOptions options;
+  private @NotNull IHub hub;
+  private @NotNull SentryOptions options;
```
bruno-garcia, line 27: m: I think this can't be `@NotNull` since the value will be always null once we construct the object.

sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
```diff
@@ -112,6 +126,10 @@ public void close() {
     if (defaultExceptionHandler != null
         && this == threadAdapter.getDefaultUncaughtExceptionHandler()) {
       threadAdapter.setDefaultUncaughtExceptionHandler(defaultExceptionHandler);
+
+      if (options != null) {
```
bruno-garcia, line 130: l: Do we need this null check even with `@NotNull`? It made me scroll up to check whether we accept null in the ctor.

-> I see that we don't take it on the ctor but do we `register` so it could be null if `close` was called before `register`

sentry-core/src/main/java/io/sentry/core/SentryOptions.java
marandaneto: I get your point and agree, but we also don't offer another way of disabling all the default integrations but iterating and removing them.
this change is not meant to make SentryOptions fully thread-safe, but to avoid issues that we already know, iterating a non-thread-safe list and removing it immediately, which might be the 1st try of most developers, either because of not knowing how list work or just guessing its a thread-safe list.
IMO it's worth making this thread-safe list but also fine if not, I Just think this might become a recurrent rant.
cc @philipphofmann ? opinions?

sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
marandaneto: good catch

sentry-core/src/main/java/io/sentry/core/UncaughtExceptionHandlerIntegration.java
marandaneto: we need this check because the comment above is valid, they should be `@Nullable` as all the other integrations

sentry-core/src/main/java/io/sentry/core/SentryOptions.java
bruno-garcia: > way of disabling all the default integrations but iterating and removing them.
I see, they are mutating the iterator instead of creating a new list and setting it back.

It does make sense then to make this change since it's the most obvious way to change things here.

I'm unaware of a legitimate use case of "removing all default integrations". We could though have this list "add only" and control what gets activated though `SentryOptions`. We're sort of doing that already, except we allow Integrations to be removed now even if SentryOptions had that feature turned on.

marandaneto: @bruno-garcia another way would be to provide a `removeIntegration(Class class)` type and we remove by type safely.

bruno-garcia: That's [what I added to the .NET SDK](https://github.com/getsentry/sentry-dotnet/blob/master/src/Sentry/SentryOptionsExtensions.cs#L56-L57):

```csharp
        internal static void RemoveIntegration<TIntegration>(this SentryOptions options) where TIntegration : ISdkIntegration
            => options.Integrations = options.Integrations.RemoveAll(p => p.GetType() == typeof(TIntegration));
```

I'm not a fan since we can have more than once instance taking a strategy, for example we have here the FileObserver bits, there are two instances.

Also, it's still confusing we have the Options flags to turn things on or off and configure them. Then this odd, by type integration thingy.

Personally I'm learning more into just "add integrations at will", configure or turn them off via SentryOptions (which can be bound to configuration files for example and do not requrie code changes)

PR #391, opened 2020-05-06
add a DateTimeProvider for making retry after testable

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
add a DateTimeProvider for making retry after testable

we were using Thread.sleep(x) and this was slowing down the tests and could also become flakey, with this we can assert things properly.
thanks @philipphofmann for the code review #381 good catch.

unit tests.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
+++ b/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
@@ -10,6 +10,7 @@ import io.sentry.core.ISerializer;
 import io.sentry.core.SentryEnvelope;
 import io.sentry.core.SentryEvent;
 import io.sentry.core.SentryOptions;
+import io.sentry.core.util.Objects;
 import io.sentry.core.util.StringUtils;
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
@@ -81,6 +82,8 @@ public class HttpTransport implements ITransport {
 
   private static final int HTTP_RETRY_AFTER_DEFAULT_DELAY_MILLIS = 60000;
 
+  private final @NotNull ICurrentDateProvider currentDateProvider;
+
   /**
    * Constructs a new HTTP transport instance. Notably, the provided {@code requestUpdater} must set
    * the appropriate content encoding header for the {@link io.sentry.core.ISerializer} instance
@@ -101,6 +104,24 @@ public class HttpTransport implements ITransport {
       final int readTimeoutMillis,
       final boolean bypassSecurity,
       final @NotNull URL sentryUrl) {
+    this(
+        options,
+        connectionConfigurator,
+        connectionTimeoutMillis,
+        readTimeoutMillis,
+        bypassSecurity,
+        sentryUrl,
+        new CurrentDateProvider());
+  }
+
+  HttpTransport(
+      final @NotNull SentryOptions options,
+      final @NotNull IConnectionConfigurator connectionConfigurator,
+      final int connectionTimeoutMillis,
+      final int readTimeoutMillis,
+      final boolean bypassSecurity,
+      final @NotNull URL sentryUrl,
+      final @NotNull ICurrentDateProvider currentDateProvider) {
     this.proxy = options.getProxy();
     this.connectionConfigurator = connectionConfigurator;
     this.serializer = options.getSerializer();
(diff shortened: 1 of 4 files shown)


sentry-core/src/main/java/io/sentry/core/transport/ICurrentDateProvider.java
```diff
@@ -0,0 +1,22 @@
...
+   */
+  Date getCurrentDate();
+
+  /**
+   * Returns a new Date based on the given timeMillis
+   *
+   * @param timeMillis the Time in millis
+   * @return the given Date
+   */
+  Date getDate(long timeMillis);
```
philipphofmann, line 21: I think we only need this because we add milliseconds to the current time, like `System.currentTimeMillis() + 1000`. What do you say we remove this method and add `currentTimeMillis` instead? Then we can do `new Date(currentDateProvider.currentTimeMillis() = 1000)`.

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
```diff
@@ -32,6 +33,7 @@ class HttpTransportTest {
         var readTimeout = 500
         var bypassSecurity = false
         val connection = mock<HttpURLConnection>()
+        val currentDateProvider = mock<ICurrentDateProvider>()
```
philipphofmann, line 36: Instead of mocking this all the time, we could implement a `TestCurrentDateProvider` that always returns the same time like `new Date(0)` and use it here. Makes the code cleaner IMO.

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
```diff
@@ -87,6 +90,8 @@ class HttpTransportTest {
         whenever(fixture.connection.inputStream).thenThrow(IOException())
         whenever(fixture.connection.getHeaderField(eq("Retry-After"))).thenReturn("30")
         whenever(fixture.connection.responseCode).thenReturn(429)
+        whenever(fixture.currentDateProvider.currentDate).thenReturn(Date())
```
philipphofmann, line 93: This returns the current date. So each run of the test is going to be with a different date. We could do `Date(0)` or `Date(2020, 12, 1, 8, 00, 00)` then this test would be deterministic.

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
marandaneto: we'd need to have more than 1 fake though, as tests would need to return also values different than `Date(0)` eg if `isRetryAfter` returns `false` when the time has passed.
I think it's not a good idea and also not sure how to test this use case if the return is always the same?

what's your LOGAF here?

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -101,13 +104,33 @@ public HttpTransport(
...
+      final @NotNull URL sentryUrl,
+      final @NotNull ICurrentDateProvider currentDateProvider) {
     this.proxy = options.getProxy();
     this.connectionConfigurator = connectionConfigurator;
     this.serializer = options.getSerializer();
     this.connectionTimeout = connectionTimeoutMillis;
     this.readTimeout = readTimeoutMillis;
     this.options = options;
     this.bypassSecurity = bypassSecurity;
+    this.currentDateProvider =
```
bruno-garcia, line 132: m: Could be on a overload or inlined here where if the date isn't provided, you just new up inline. (or use a singleton since this will likely be used in a few more places).

The reasoning here is: We don't *need* this dependency. This is just for unit testing. In prod code we won't use anything other than the default one.

sentry-core/src/main/java/io/sentry/core/transport/ICurrentDateProvider.java
marandaneto: `+` do you mean? sounds a good idea
what's your LOGAF scale here?

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
marandaneto: true, let me figure this out, don't forget the LOGAF please

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: mm not sure if I got it, the ctor that takes `currentDateProvider` is not public, the public one creates its own `currentDateProvider`, so its not a dependency, can you elaborate?

bruno-garcia: I missed that, its resolved then. thanks

PR #392, opened 2020-05-06
removed Thread.sleep from LifecycleWatcher tests, using awaitility and DateProvider

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
removed Thread.sleep from LifecycleWatcher tests, using awaitility and DateProvider

Thread sleep was slowing down tests and also tests using `currentTimeMillis` were not deterministic.

`@Ignore` test should be working now

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java
@@ -25,7 +25,7 @@ final class LifecycleWatcher implements DefaultLifecycleObserver {
   private final @NotNull IHub hub;
   private final boolean enableSessionTracking;
   private final boolean enableAppLifecycleBreadcrumbs;
-  private final AtomicBoolean runningSession = new AtomicBoolean();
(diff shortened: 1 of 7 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java
```diff
@@ -81,6 +105,7 @@ public void run() {
   private void cancelTask() {
     if (timerTask != null) {
```
bruno-garcia, line 106: Could `cancelTask` concurrently? I guess that'd depend if `onStop` could be called concurrently.
If that's the case (I assume not but), this could be racey now that we set it to `null` below

marandaneto: if they do, it'd be a bug in `androidx.lifecycle:lifecycle-process`, but don't think so, `onStop` is a post background callback, we are safe

PR #397, opened 2020-05-08
added missing getters on Breadcrumb and SentryEvent

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

added missing getters on Breadcrumb and SentryEvent

it was not possible to use `beforeSend` or `beforeBreadcrumb` with those fields.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/SentryEvent.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryEvent.java
@@ -193,15 +193,8 @@ public final class SentryEvent implements IUnknownPropertiesConsumer {
     this.sdk = sdk;
   }
 
-  List<String> getFingerprints() {
+  public List<String> getFingerprints() {
     return fingerprint;
-  }
-
-  public boolean hasFingerprint(final @NotNull String fingerprint) {
-    if (this.fingerprint != null) {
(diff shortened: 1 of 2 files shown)


sentry-core/src/main/java/io/sentry/core/SentryEvent.java
```diff
@@ -196,6 +197,13 @@ public void setSdk(SdkVersion sdk) {
     return fingerprint;
   }
 
+  public boolean hasFingerprint(final @NotNull String fingerprint) {
```
bruno-garcia, line 200: h: Not sure this makes sense since the Fingerprint is the whole list, not a single item in it.
What's the motivation for this API?

marandaneto: If one wants to filter it, it would be possible, the getter is not public, what would you suggest? Returning the whole list?

bruno-garcia: Yeah that'd be the way to go. Just `List<> getFingerprint` looks good

PR #398, opened 2020-05-08
tests default timeout millis to 0

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
tests default timeout millis to 0, as Hub is mock, CountDownLatch was never count down

some tests were taking 15s which is the time to release the CountDownLatch

ran tests and appveyor

## :pencil: Checklist

- [X] All tests passing


marandaneto:
> What about an extension method to `SentryOptions` to get a test instance that comes with flushTimeout 0?

Yep good idea

PR #399, opened 2020-05-09
Update issue templates

marandaneto:


marandaneto:
> Good point. Although maybe a comment suggesting to consider raising on the RN repo?

They already do it directly there, this would be only when they use sentry-android directly in their Java/Kotlin code, or I'd move the issue to sentry-react-native repo.

Issue #361, opened 2020-04-16
Sending cached events periodically


nominalista:
Because some of the events may be not be sent immediately after they are created, it would be great if there was an option to send these events periodically, event if the app wasn't restarted. 

Is there any method in SDK that I can call and try to send cached events? I would use it as a temporary solution along with e.g. `WorkManager`.

marandaneto:
as agreed, I'll mark it as a feature request, thanks @nominalista 
related to #360

nominalista:
Thanks @marandaneto. Just to be sure - are you considering the whole periodic mechanism or only method that would allow to send cached events? If whole mechanism, is such method already existing?

marandaneto:
the mechanism to send periodically is something to be done, this is the feature request.
the method to send cached events are already in place as mentioned in #360 `SendCachedEventFireAndForgetIntegration`
https://github.com/getsentry/sentry-android/blob/master/sentry-core/src/main/java/io/sentry/core/SentryOptions.java#L817

If you use this integration periodically, it'd work, but it might be racey, be aware of that.

nominalista:
Ok, thank you!

PR #405, opened 2020-05-13
fix: send cached events to use a single thread executor

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
send cached events to use a single executor.

running it

## :pencil: Checklist

- [ ] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
+++ b/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
@@ -1,5 +1,6 @@
 package io.sentry.core;
 
+import static io.sentry.core.SentryLevel.DEBUG;
 import static io.sentry.core.SentryLevel.ERROR;
 import static io.sentry.core.cache.SessionCache.PREFIX_CURRENT_SESSION_FILE;
 
@@ -51,7 +52,7 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
     Objects.requireNonNull(file, "File is required.");
 
     if (!isRelevantFileName(file.getName())) {
-      logger.log(SentryLevel.DEBUG, "File '%s' should be ignored.", file.getName());
+      logger.log(SentryLevel.DEBUG, "File '%s' should be ignored.", file.getAbsolutePath());
       return;
     }
 
@@ -68,12 +69,18 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
     } catch (IOException e) {
       logger.log(SentryLevel.ERROR, "Error processing envelope.", e);
     } finally {
-      if ((hint instanceof Retryable) && !((Retryable) hint).isRetry()) {
-        try {
-          file.delete();
-        } catch (RuntimeException e) {
-          logger.log(SentryLevel.ERROR, "Failed to delete.", e);
+      if (hint instanceof Retryable) {
+        if (!((Retryable) hint).isRetry()) {
+          try {
+            if (!file.delete()) {
+              logger.log(SentryLevel.ERROR, "Failed to delete: %s", file.getAbsolutePath());
+            }
+          } catch (RuntimeException e) {
+            logger.log(SentryLevel.ERROR, e, "Failed to delete: %s", file.getAbsolutePath());
+          }
         }
+      } else {
+        logger.log(SentryLevel.DEBUG, "File '%s' is not Retryable.", file.getAbsolutePath());
       }
     }
   }
@@ -130,12 +137,19 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
             }
             hub.captureEvent(event, hint);
             logger.log(SentryLevel.DEBUG, "Item %d is being captured.", items);
-            if ((hint instanceof Flushable) && !((Flushable) hint).waitFlush()) {
+            if (hint instanceof Flushable) {
+              if (!((Flushable) hint).waitFlush()) {
+                logger.log(
+                    SentryLevel.WARNING,
+                    "Timed out waiting for event submission: %s",
+                    event.getEventId());
+                break;
+              }
+            } else {
               logger.log(
-                  SentryLevel.WARNING,
-                  "Timed out waiting for event submission: %s",
-                  event.getEventId());
-              break;
+                  SentryLevel.DEBUG,
+                  "Envelope %s doesn't require a SubmissionResult",
+                  envelope.getHeader().getEventId());
             }
           }
         } catch (Exception e) {
@@ -156,12 +170,17 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
             // TODO: Bundle all session in a single envelope
             hub.captureEnvelope(SentryEnvelope.fromSession(serializer, session), hint);
             logger.log(SentryLevel.DEBUG, "Item %d is being captured.", items);
-            if ((hint instanceof Flushable) && !((Flushable) hint).waitFlush()) {
-              logger.log(
-                  SentryLevel.WARNING,
-                  "Timed out waiting for item submission: %s",
-                  session.getSessionId());
-              break;
+
+            if (hint instanceof Flushable) {
+              if (!((Flushable) hint).waitFlush()) {
+                logger.log(
+                    SentryLevel.WARNING,
+                    "Timed out waiting for item submission: %s",
+                    session.getSessionId());
+                break;
+              }
+            } else {
+              logger.log(DEBUG, "Envelope %s is not Flushable.", envelope.getHeader().getEventId());
             }
           }
         } catch (Exception e) {
@@ -173,14 +192,21 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
             SentryLevel.WARNING, "Item %d of type: %s ignored.", items, item.getHeader().getType());
       }
 
-      if ((hint instanceof SubmissionResult) && !((SubmissionResult) hint).isSuccess()) {
-        // Failed to send an item of the envelope: Stop attempting to send the rest (an attachment
-        // without the event that created it isn't useful)
+      if (hint instanceof SubmissionResult) {
+        if (!((SubmissionResult) hint).isSuccess()) {
+          // Failed to send an item of the envelope: Stop attempting to send the rest (an attachment
+          // without the event that created it isn't useful)
+          logger.log(
+              SentryLevel.WARNING,
+              "Envelope had a failed capture at item %d. No more items will be sent.",
+              items);
+          break;
+        }
+      } else {
         logger.log(
-            SentryLevel.WARNING,
-            "Envelope had a failed capture at item %d. No more items will be sent.",
-            items);
-        break;
+            SentryLevel.DEBUG,
+            "Envelope %s doesn't require a SubmissionResult",
+            envelope.getHeader().getEventId());
       }
     }
   }
--- a/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
+++ b/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
@@ -173,7 +173,7 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
             logger.log(SentryLevel.DEBUG, "Item %d is being captured.", items);
 
             if (hint instanceof Flushable) {
-              logger.log(SentryLevel.DEBUG, "Gonna wait flush %d item.", items);
+              logger.log(SentryLevel.DEBUG, "Going to wait flush %d item.", items);
               if (!((Flushable) hint).waitFlush()) {
                 logger.log(
                     SentryLevel.WARNING,
--- a/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
+++ b/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
@@ -148,7 +148,7 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
             } else {
               logger.log(
                   SentryLevel.DEBUG,
-                  "Envelope %s doesn't require a SubmissionResult",
+                  "Envelope's Hint is not Flushable",
                   envelope.getHeader().getEventId());
             }
           }
@@ -205,7 +205,7 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
       } else {
         logger.log(
             SentryLevel.DEBUG,
-            "Envelope %s doesn't require a SubmissionResult",
+            "Envelope's Hint %s is not SubmissionResult",
             envelope.getHeader().getEventId());
       }
     }
--- a/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
+++ b/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
@@ -65,6 +65,7 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
             file.getAbsolutePath());
       } else {
         processEnvelope(envelope, hint);
+        logger.log(SentryLevel.DEBUG, "File '%s' is done.", file.getAbsolutePath());
       }
     } catch (IOException e) {
       logger.log(SentryLevel.ERROR, "Error processing envelope.", e);
@@ -172,6 +173,7 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
             logger.log(SentryLevel.DEBUG, "Item %d is being captured.", items);
 
             if (hint instanceof Flushable) {
+              logger.log(SentryLevel.DEBUG, "Gonna wait flush %d item.", items);
               if (!((Flushable) hint).waitFlush()) {
                 logger.log(
                     SentryLevel.WARNING,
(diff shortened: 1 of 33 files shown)


bruno-garcia:
We discussed this offline but I wanted to leave one note: Let please leave some sort of timeout. Even if it's 30 seconds. IMHO 30 seconds waiting for a single event to be flushed is a lot but let's just not let the thread block forever without timeout at least

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -162,6 +163,11 @@ public HttpTransport(
         final Writer writer = new BufferedWriter(new OutputStreamWriter(outputStream, UTF_8))) {
       serializer.serialize(event, writer);
 
+      // TODO: remove this, just logging
+      StringWriter wrt = new StringWriter();
+      serializer.serialize(event, wrt);
+      options.getLogger().log(INFO, "event: %s", wrt.toString());
+
```
marandaneto, line 170: only for logging and testing...

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -300,6 +306,14 @@ public boolean isRetryAfter(final @NotNull String itemType) {
         final Writer writer = new BufferedWriter(new OutputStreamWriter(outputStream, UTF_8))) {
       serializer.serialize(envelope, writer);
 
+      // TODO: remove this, just logging
+      options
```
marandaneto, line 310: only for logging and testing...

sentry-sample/src/main/AndroidManifest.xml
```diff
@@ -44,7 +44,7 @@
 <!--    <meta-data android:name="io.sentry.anr.report-debug" android:value="true" />-->
 
 <!--    how to disable ANR-->
-<!--    <meta-data android:name="io.sentry.anr.enable" android:value="false" />-->
+    <meta-data android:name="io.sentry.anr.enable" android:value="false" />
```
marandaneto, line 47: only for testing...

sentry-sample/src/main/AndroidManifest.xml
```diff
@@ -53,7 +53,7 @@
 <!--    <meta-data android:name="io.sentry.ndk.enable" android:value="false" />-->
 
 <!--    how to set a custom release-->
-<!--    <meta-data android:name="io.sentry.release" android:value="io.sentry.sample@1.0.0+10001" />-->
+    <meta-data android:name="io.sentry.release" android:value="io.sentry.yeyey@2.0.0+10001" />
```
marandaneto, line 56: only for testing...

sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
```diff
@@ -90,10 +103,18 @@ protected boolean isRelevantFileName(String fileName) {
...
+            SentryLevel.ERROR,
+            "Failed to delete '%s' " + errorMessageSuffix,
+            file.getAbsolutePath());
+      }
     } catch (Exception e) {
       logger.log(
-          SentryLevel.ERROR, "Failed to delete '%s' " + errorMessageSuffix, file.getName(), e);
+          SentryLevel.ERROR,
+          "Failed to delete '%s' " + errorMessageSuffix,
+          file.getAbsolutePath(),
```
bruno-garcia, line 116: M: The format string is useful to avoid the allocation when logging is disabled so if we can get around passing `%s` instead of `+`, that'd be ideal:


```suggestion
          "Failed to delete '%s' %s",
          file.getAbsolutePath(),
          errorMessageSuffix,
```

sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
```diff
@@ -90,10 +103,18 @@ protected boolean isRelevantFileName(String fileName) {
 
   private void safeDelete(File file, String errorMessageSuffix) {
     try {
-      file.delete();
+      if (!file.delete()) {
+        logger.log(
+            SentryLevel.ERROR,
+            "Failed to delete '%s' " + errorMessageSuffix,
+            file.getAbsolutePath());
```
bruno-garcia, line 110: M: The format string is useful to avoid the allocation when logging is disabled so if we can get around passing `%s` instead of `+`, that'd be ideal:

```suggestion
            "Failed to delete '%s' %s",
            file.getAbsolutePath(),
            errorMessageSuffix);
```

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
```diff
@@ -34,25 +36,21 @@ public final void register(@NotNull IHub hub, @NotNull SentryOptions options) {
...
-      es.shutdown();
+          .getExecutorService()
+          .submit(
+              () -> {
+                try {
+                  sender.send();
+                } catch (Exception e) {
+                  options
+                      .getLogger()
+                      .log(SentryLevel.ERROR, "Failed trying to send cached events.", e);
```
bruno-garcia, line 51: H: This API is error prone because we take `Object...` for the string format but the `Throwable` argument is the first:

```suggestion
                      .log(SentryLevel.ERROR, e, "Failed trying to send cached events.");
```

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
```diff
@@ -34,25 +36,21 @@ public final void register(@NotNull IHub hub, @NotNull SentryOptions options) {
...
-    SendFireAndForget sender = factory.create(hub, options);
+    final SendFireAndForget sender = factory.create(hub, options);
 
     try {
-      ExecutorService es = Executors.newSingleThreadExecutor();
-      es.submit(
-          () -> {
-            try {
-              sender.send();
-              options
```
bruno-garcia, line 45: We lost this log message. Was that intentional?

sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
```diff
@@ -90,10 +103,18 @@ protected boolean isRelevantFileName(String fileName) {
...
+      }
     } catch (Exception e) {
       logger.log(
-          SentryLevel.ERROR, "Failed to delete '%s' " + errorMessageSuffix, file.getName(), e);
+          SentryLevel.ERROR,
+          "Failed to delete '%s' " + errorMessageSuffix,
+          file.getAbsolutePath(),
+          e);
```
bruno-garcia, line 117: H: This should be the first arg, to use this overload:

https://github.com/getsentry/sentry-android/blob/master/sentry-core/src/main/java/io/sentry/core/ILogger.java#L32

sentry-core/src/main/java/io/sentry/core/SendFireAndForgetEnvelopeSender.java
```diff
@@ -0,0 +1,51 @@
...
+  @Override
+  public SendCachedEventFireAndForgetIntegration.SendFireAndForget create(
+      final @NotNull IHub hub, final @NotNull SentryOptions options) {
+    final String dirPath = sendFireAndForgetDirPath.getDirPath();
+    if (dirPath == null) {
+      options
+          .getLogger()
+          .log(
+              SentryLevel.WARNING,
+              "No cache dir path is defined in options, discarding SendCachedEvent.");
```
bruno-garcia, line 29: I'm a bit confused now because it implements `SendCachedEventFireAndForgetIntegration` which has `Event` in the name but this is targeting Envelopes.

```suggestion
              "No cache dir path is defined in options, discarding SendCachedEvelope.");
```

Down below it uses `final File sessionsPath = new File(dirPath);` so maybe it needs some renaming here

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
```diff
@@ -34,25 +36,21 @@ public final void register(@NotNull IHub hub, @NotNull SentryOptions options) {
...
-                  .log(SentryLevel.DEBUG, "Finished processing cached files from %s", cachedDir);
-            } catch (Exception e) {
-              options.getLogger().log(SentryLevel.ERROR, "Failed trying to send cached events.", e);
-            }
-          });
       options
-          .getLogger()
-          .log(SentryLevel.DEBUG, "Scheduled sending cached files from %s", cachedDir);
-      es.shutdown();
+          .getExecutorService()
```
bruno-garcia, line 43: H: Was the goal of adding `executorService` to SentryOptions so we can control the `close`? I'm not sure it belongs here, xan we discuss this a bit in more detail before going ahead with this?

So far the only use of it is to be closed during `Hub.close`. If that's the case, could we simply implement `Closeable` on this type and `shutdown` the executor there? And still have a single executor for this type?

Or do we need to share this executor? We could investigate a `SentryExexcutorService` that we'd pass around and could be used by different bits of the SDK?

sentry-core/src/main/java/io/sentry/core/SentryOptions.java
```diff
@@ -177,6 +179,8 @@
    */
   private boolean enableUncaughtExceptionHandler = true;
 
+  private final @NotNull ExecutorService executorService;
```
bruno-garcia, line 182: Left a comment above but ideally we wouldn't have this here

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
bruno-garcia: What about a `logger.isEnabled(LogLevel.debug)) { ...` ?
This way we only pay for the serialization if logging is enabled for that log level (debug).

marandaneto: I'd not log the whole payload, when this gets out of a Draft, I'll remove this lines, see the TODO above

sentry-core/src/main/java/io/sentry/core/Hub.java
```diff
@@ -249,6 +249,9 @@ public void close() {
             ((Closeable) integration).close();
           }
         }
+        if (!options.getExecutorService().isShutdown()) {
+          options.getExecutorService().shutdown();
+        }
```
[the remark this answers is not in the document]
marandaneto: `shutdown` only stops accepting new tasks, so it won't be a problem, otherwise, we'd need to call `shutdownNow` and wait for X period of time which is thread-safe, I believe this would be part of the `flush` method, but not on this one.

sentry-sample/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
marandaneto: this is Outdated already

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
```diff
@@ -156,12 +171,19 @@ private void processEnvelope(final @NotNull SentryEnvelope envelope, final @Null
...
             logger.log(SentryLevel.DEBUG, "Item %d is being captured.", items);
-            if ((hint instanceof Flushable) && !((Flushable) hint).waitFlush()) {
-              logger.log(
-                  SentryLevel.WARNING,
-                  "Timed out waiting for item submission: %s",
-                  session.getSessionId());
-              break;
+
+            if (hint instanceof Flushable) {
+              logger.log(SentryLevel.DEBUG, "Gonna wait flush %d item.", items);
```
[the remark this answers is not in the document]
marandaneto: thanks

sentry-core/src/main/java/io/sentry/core/SendCachedEvent.java
marandaneto: yep copy-pasta logs, it was like this already, fixed it

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
marandaneto: yes, also noticed that, I already fixed all the wrong args btw, but it'd be nice to fix that

marandaneto: not on this PR though

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
marandaneto: they now live in SendFireAndForgetEnvelopeSender and SendFireAndForgetEventSender

sentry-core/src/main/java/io/sentry/core/SendFireAndForgetEnvelopeSender.java
marandaneto: yep, good catch, copy pasta typo

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
marandaneto: 2 things, 1st is indeed to be able to close and not allow a living thread forever if `close` has been called.

my first try was implementing `Closeable`, but we have 3 integrations using the very same executor, it might be the case that you call close for a single integration and this is going to close the executor for the other 2 integrations too, might not be a real case, but still possible.

I decided to expose an executor that can also be used in `AndroidOptionsInitializer`, as we also need to init. the integration for sending NDK events and then, close the executor on `hub.close()`

We need to share the executor to avoid racey conditions and also to minimize a lot of stuff in parallel on the App. start/restart, it's just optimal to execute one after the other using the same executor, this also makes our `RetryingThreadPoolExecutor` to not get a lot of items at the same time and drop a few if >= queue size.

btw the name of the PR is `send cached events to use a single executor`, so obviously the idea is to share an executor for the integrations that send cached events/envelopes, I've updated this PR yesterday so just telling in case you have missed this.

`SentryExexcutorService` is a great idea, I'd go for something like this too, and this goes to SentryOptions, right? as we need to access in other places.
As this is a Draft PR, it's not fully tested and so on, but it did solve the problems I've found, let's discuss.

sentry-core/src/main/java/io/sentry/core/SentryOptions.java
marandaneto: answered above

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
marandaneto: btw this can also help with #361 in the future

bruno-garcia: > my first try was implementing Closeable, but we have 3 integrations using the very same executor, it might be the case that you call close for a single integration and this is going to close the executor for the other 2 integrations too, might not be a real case, but still possible.

Fair but one SentryOptions could be used on a second SentryClient to be bound to the same Hub (i.e to have a second background queue, also not likely a common use case but possible). And the client also `close` too which should flush its events but not necessarily close all other clients/integrations in the app.

sentry-core/src/main/java/io/sentry/core/SentryExecutorService.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Override
+  public void close(final long timeoutMillis) {
+    synchronized (executorService) {
+      if (!executorService.isShutdown()) {
+        executorService.shutdown();
+        try {
+          executorService.awaitTermination(timeoutMillis, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+          executorService.shutdownNow();
```
bruno-garcia, line 36: Could this throw and we never call `interrupt` below?

sentry-core/src/main/java/io/sentry/core/SendCachedEventFireAndForgetIntegration.java
marandaneto: discussed offline

sentry-core/src/main/java/io/sentry/core/SentryExecutorService.java
marandaneto: no based on the docs https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html

sentry-core/src/test/java/io/sentry/core/SentryClientTest.kt
```diff
@@ -599,4 +636,10 @@ class SentryClientTest {
     }
 
     internal class CustomCachedApplyScopeDataHint : Cached, ApplyScopeData
+
+    internal class DiskFlushNotificationHint : DiskFlushNotification {
+        override fun markFlushed() {
+            TODO("Not yet implemented")
```
bruno-garcia, line 642: L: Is there something to be implemented here or it'll just be empty?

marandaneto: its an internal and test class `SentryClientTest.kt`

marandaneto: but yeah could remove this comment

PR #409, opened 2020-05-14
fix: initing SDK on AttachBaseContext

### LINKED ISSUE #408: Cannot manually init Sentry in Application.attachBaseContext
_Platform:_
- [x]  Android -> API 28
- [ ]  Java -> If yes, which Java (and sourceCompatibility/targetCompatibility) version?
- [ ]  Kotlin -> If yes, which Kotlin (and jvmTarget) version?
- [ ]  NDK -> If yes, which NDK/CMake version?
- [ ]  React-Native -> If yes, which version?

_IDE:_
- [x]  Android Studio -> 3.6.3
- [ ]  IntelliJ -> If yes, which version?
- [ ]  Other -> If yes, which one?

_Build system:_
- [x]  Gradle -> 5.6.4
- [ ]  Buck -> If yes, which version?
- [ ]  Bazel -> If yes, which version?
- [ ]  Other -> If yes, which one?

_Android Gradle Plugin:_
- [x]  Yes -> 3.6.3
- [ ]  No

_Sentry Android Gradle Plugin:_
- [x]  Yes -> 1.7.23
- [ ]  No

_Proguard/R8:_
- [x]  Enabled
- [ ]  Disabled

_sentry-android installed with:_
- [ ] JCenter
- [ ] Bintray
- [x] Maven Central
- [ ] Manually

The version of sentry-android:
**2.1.0**

---
I have the following issue:

If I manually init Sentry with `SentryAndroid.init(context)` from inside `Application#attachBaseContext` I will get an exception.

Root cause: https://github.com/getsentry/sentry-android/blob/46658948a80bebade5ebe07402e38ec6384f1669/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java#L51

`getApplicationContext` will return `null` when called from `attachBaseContext`

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: initing SDK on AttachBaseContext
Similar issue getsentry/sentry-java/pull/714

Fix #408

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -49,12 +49,9 @@ final class AndroidOptionsInitializer {
       final @NotNull ILogger logger) {
     Objects.requireNonNull(context, "The context is required.");
 
-    if (!(context instanceof Application)) {
-      // if context is not an AppContext, let's try to get from getApplicationContext
-      // it returns null if ContextImpl, so let's check for nullability
-      if (context.getApplicationContext() != null) {
-        context = context.getApplicationContext();
-      }
+    // it returns null if ContextImpl, so let's check for nullability
+    if (context.getApplicationContext() != null) {
+      context = context.getApplicationContext();
     }
 
     Objects.requireNonNull(options, "The options object is required.");
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -48,9 +48,13 @@ final class AndroidOptionsInitializer {
       @NotNull Context context,
       final @NotNull ILogger logger) {
     Objects.requireNonNull(context, "The context is required.");
-    context =
-        Objects.requireNonNull(
-            context.getApplicationContext(), "The application context is required.");
+
+    if (!(context instanceof Application)) {
+      context =
+          Objects.requireNonNull(
+              context.getApplicationContext(), "The application context is required.");
+    }
+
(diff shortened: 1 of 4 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -48,9 +48,15 @@ static void init(
       @NotNull Context context,
       final @NotNull ILogger logger) {
     Objects.requireNonNull(context, "The context is required.");
-    context =
-        Objects.requireNonNull(
-            context.getApplicationContext(), "The application context is required.");
+
+    if (!(context instanceof Application)) {
```
ninniuz, line 52: Maybe this is not needed? Wouldn't it be easier to just say "if the application context is not null I will use that, otherwise I will stick with what you gave me"? Don't get me wrong, this is ok. Just wondering about maintainability.

marandaneto: @ninniuz yeah, I thought to use the given Context, and only overwrite if it's actually needed, even though if the context is already an Application, `getApplicationContext` returns the same `Application` instance right.

ninniuz: Yes, that's why I think it is safe for both SDK users and the SDK itself to use `context.getApplicationContext()` if it is not `null` and fallback to the provided one otherwise. Of course you would have to keep the other checks for integrations as you already pointed out.

marandaneto: yep, done, thanks :)

PR #410, opened 2020-05-14
fix: sessions cant be abnormal, but exited if not ended properly

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
sessions cant be abnormal, but exited if not ended properly

abnornal sessions for mobile doesn't make sense.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/SessionAdapter.java
+++ b/sentry-core/src/main/java/io/sentry/core/SessionAdapter.java
@@ -34,13 +34,8 @@ public final class SessionAdapter extends TypeAdapter<Session> {
       writer.name("init").value(value.getInit());
     }
 
-    if (value.getStarted() != null) {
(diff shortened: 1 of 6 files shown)


sentry-core/src/main/java/io/sentry/core/SessionAdapter.java
```diff
@@ -182,6 +175,10 @@ public Session read(JsonReader reader) throws IOException {
     }
     reader.endObject();
 
+    if (status == null || started == null || release == null || release.isEmpty()) {
```
bruno-garcia, line 178: l: Maybe we log here? Would help debugging if this happens

marandaneto: good idea

Issue #406, opened 2020-05-13
High ANR report rate with v2.x


tobi512:
Hi there, 
we're running sentry-android v2.0.3 in production since quite a while now and everything seems fine now after you fixed #341. 
However, we see a lot of ANRs (which is a new feature since v2.x as far as I know, right?) that we don't see in the Play Console. At first we thought that all those reported ANRs are real and tried to fix some (with no success), but right now we're not so sure about this anymore, which is why i'm opening this issue.

Speaking numbers, we got ~18k (yes, 18 thousand) ANR reports in Sentry till now, while the Play Console only shows around 200 if I sum them up by hand (all app versions, all OS versions in the last 60 days) and also displays "100% daily sessions without ANRs" on the dashboard. Interesting about the ANRs from Sentry is that most of them are coming from devices with Android 6.x.

I see that you use ANR-WatchDog with a threshold of 4000ms for detection while Google has 5s as their threshold (according to https://developer.android.com/topic/performance/vitals/anr) but I really don't think that this difference of 1s is responsible for all the ANR reports that we get from Sentry compare to Google. 

The question is now, where are all these reports coming from and why, since we never really see our app hanging during testing nor did we get customer reports about it (with currently over 1 million active installs).

Regards, 
Tobias

marandaneto:
hey @tobi512 thanks for raising this.

Some info before getting to the point:

#388 changes the ANR timeout to 5s by default, released on this version https://github.com/getsentry/sentry-android/releases/tag/2.1.0

you can also set a custom timeout adding this to the Manifest:

// 10s
`<meta-data android:name="io.sentry.anr.timeout-interval-millis" android:value="10000" />-->`

you can also disable the ANR feature as well:
`<meta-data android:name="io.sentry.anr.enable" android:value="false" />`

ANR is enabled by default on `sentry-android-2.x`, you are right. 

We detect ANRs in a different way comparing to Google Vitals.
eg We don't request user consent to send ANR events as Google does.

We detect ANRs on this way https://github.com/SalomonBrys/ANR-WatchDog

That said, we'll investigate if there's any bug/problem with Android 6, but not only, of course.
Stay tuned.

tobi512:
Hi @marandaneto, 
thanks for your quick reply, you guys are doing a great job here!

Customizing the ANR timeout sounds good, we might give that a try or update to the latest version of the SDK. 

> We detect ANRs in a different way comparing to Google Vitals.
[...]

PR #412, opened 2020-05-15
fix: before raising ANR event, we check ProcessErrorStateInfo if available

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: before raising ANR event, we check ProcessErrorStateInfo if available

ANR events are too noisy
#406

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
@@ -92,7 +92,8 @@ final class ANRWatchDog extends Thread {
           continue;
         }
 
-        // we only raise an ANR if the process is in ANR state.
+        // we only raise an ANR event if the process is in ANR state.
+        // if ActivityManager is not available, we'll still be able to send ANRs
         final ActivityManager am =
             (ActivityManager) context.getSystemService(Context.ACTIVITY_SERVICE);
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
@@ -27,7 +27,6 @@ final class ANRWatchDog extends Thread {
   private AtomicLong tick = new AtomicLong(0);
   private AtomicBoolean reported = new AtomicBoolean(false);
 
-  @SuppressWarnings("UnusedVariable")
   private final @NotNull Context context;
 
   @SuppressWarnings("UnnecessaryLambda")
@@ -100,7 +99,7 @@ final class ANRWatchDog extends Thread {
         if (am != null) {
           final List<ActivityManager.ProcessErrorStateInfo> processesInErrorState =
               am.getProcessesInErrorState();
-          // if list is null, there are no process in ANR state.
+          // if list is null, there's no process in ANR state.
           if (processesInErrorState == null) {
             continue;
           }
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
@@ -2,9 +2,15 @@
 // Based on the class above. The API unnecessary here was removed.
 package io.sentry.android.core;
 
+import static android.app.ActivityManager.ProcessErrorStateInfo.NOT_RESPONDING;
+
+import android.app.ActivityManager;
+import android.content.Context;
 import android.os.Debug;
 import io.sentry.core.ILogger;
 import io.sentry.core.SentryLevel;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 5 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
```diff
@@ -73,16 +83,39 @@ public void run() {
...
               SentryLevel.DEBUG,
               "An ANR was detected but ignored because the debugger is connected.");
-          reported = true;
+          reported.set(true);
           continue;
         }
 
+        // we only raise an ANR event if the process is in ANR state.
+        // if ActivityManager is not available, we'll still be able to send ANRs
+        final ActivityManager am =
```
marandaneto, line 97: we could add a flag here to keep the old behavior, so we don't check its process, not sure if it's worth though.
this flag would live in `SentryAndroidOptions` and enabled by default.

sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
```diff
@@ -20,10 +22,17 @@
...
+
+  public AnrIntegration(final @NotNull Context context) {
+    this.context = context;
+  }
+
   /**
    * Responsible for watching UI thread. being static to avoid multiple instances running at the
    * same time.
    */
+  @SuppressLint("StaticFieldLeak") // we have watchDogLock to avoid those leaks
```
marandaneto, line 35: we can make it non-final, but then its hard to guarantee that this integration won't be registered twice.

sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
bruno-garcia: Agree not sure it's worth it. Let's take this approach as the new default and see how it goes first

sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
```diff
@@ -73,16 +83,39 @@ public void run() {
...
+        if (am != null) {
+          final List<ActivityManager.ProcessErrorStateInfo> processesInErrorState =
+              am.getProcessesInErrorState();
+          // if list is null, there's no process in ANR state.
+          if (processesInErrorState == null) {
+            continue;
+          }
+          boolean isAnr = false;
+          for (ActivityManager.ProcessErrorStateInfo item : processesInErrorState) {
+            if (item.condition == NOT_RESPONDING) {
```
bruno-garcia, line 109: So this is a best effort since it could give false positive? Meaning there's some process in ANR state but this isn't the current one? Either way would improve **a lot** already.

sentry-android-core/src/main/java/io/sentry/android/core/AnrIntegration.java
bruno-garcia: We create an instance under a lock and set the instance to null on close so looks good already

sentry-android-core/src/test/java/io/sentry/android/core/ANRWatchDogTest.kt
```diff
@@ -76,4 +81,83 @@ class ANRWatchDogTest {
...
+            assertEquals(expectedState, anr!!.thread.state)
+            assertEquals(stacktrace.className, anr!!.stackTrace[0].className)
+        } finally {
+            sut.interrupt()
+            es.shutdown()
+        }
+    }
+
+    @Test
+    fun `when ANR is detected and ActivityManager has not ANR process, callback is not invoked`() {
```
bruno-garcia, line 127:
```suggestion
    fun `when ANR is detected and ActivityManager has no ANR process, callback is not invoked`() {
```

sentry-android-core/src/test/java/io/sentry/android/core/ANRWatchDogTest.kt
```diff
@@ -76,4 +81,83 @@ class ANRWatchDogTest {
             es.shutdown()
         }
     }
+
+    @Test
+    fun `when ANR is detected and ActivityManager has ANR process, callback not invoked`() {
```
bruno-garcia, line 86: Is this the name instead?

```suggestion
    fun `when ANR is detected and ActivityManager has ANR process, callback is invoked`() {
```

sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
marandaneto: it's only the processes of this app, so that's why it's a list. an App can have multiple processes but its not common at all.

sentry-android-core/src/test/java/io/sentry/android/core/ANRWatchDogTest.kt
marandaneto: damn good catch, I need to pay more attention to my copypasta hehe

PR #417, opened 2020-05-17
enha: make root checker better and minimize false positive

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
make root checker better and minimize false positive.

I still could reproduce false positive on the current root checker

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -21,6 +21,7 @@ import android.provider.Settings;
 import android.util.DisplayMetrics;
 import io.sentry.android.core.util.ConnectivityChecker;
 import io.sentry.android.core.util.DeviceOrientations;
+import io.sentry.android.core.util.RootChecker;
 import io.sentry.core.DateUtils;
 import io.sentry.core.EventProcessor;
 import io.sentry.core.SentryEvent;
@@ -94,7 +95,7 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
       map.put(PROGUARD_UUID, proGuardUuids);
     }
 
-    map.put(ROOTED, isRooted());
+    map.put(ROOTED, RootChecker.isDeviceRooted(context, options.getLogger()));
 
     String androidId = getAndroidId();
     if (androidId != null) {
@@ -777,54 +778,6 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
     }
 
     return defaultVersion;
-  }
-
-  /**
-   * Attempt to discover if this device is currently rooted. From:
-   * https://stackoverflow.com/questions/1101380/determine-if-running-on-a-rooted-device
-   *
-   * @return true if heuristics show the device is probably rooted, otherwise false
-   */
-  private boolean isRooted() {
-    // we could get some inspiration from https://github.com/scottyab/rootbeer
-    if (Build.TAGS != null && Build.TAGS.contains("test-keys")) {
-      return true;
-    }
-
-    String[] probableRootPaths = {
-      "/data/local/bin/su",
-      "/data/local/su",
-      "/data/local/xbin/su",
-      "/sbin/su",
-      "/su/bin",
-      "/su/bin/su",
-      "/system/app/SuperSU",
-      "/system/app/SuperSU.apk",
-      "/system/app/Superuser",
-      "/system/app/Superuser.apk",
-      "/system/bin/failsafe/su",
-      "/system/bin/su",
-      "/system/sd/xbin/su",
-      "/system/xbin/daemonsu",
-      "/system/xbin/su"
-    };
-
-    for (String probableRootPath : probableRootPaths) {
-      try {
-        if (new File(probableRootPath).exists()) {
-          return true;
-        }
-      } catch (Exception e) {
-        options
-            .getLogger()
-            .log(
-                SentryLevel.ERROR,
-                "Exception while attempting to detect whether the device is rooted",
-                e);
-      }
-    }
-
-    return false;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -94,7 +95,7 @@ public DefaultAndroidEventProcessor(
       map.put(PROGUARD_UUID, proGuardUuids);
     }
 
-    map.put(ROOTED, isRooted());
+    map.put(ROOTED, RootChecker.isDeviceRooted(context, options.getLogger()));
```
philipphofmann, line 98: I like that we have this logic now in a separate class 👏

sentry-android-core/src/main/java/io/sentry/android/core/util/RootChecker.java
```diff
@@ -0,0 +1,147 @@
...
+
+  /** the UTF-8 Charset */
+  @SuppressWarnings("CharsetObjectCanBeUsed")
+  private static final Charset UTF_8 = Charset.forName("UTF-8");
+
+  private RootChecker() {}
+
+  /**
+   * Check if the device is rooted or not
+   * https://medium.com/@thehimanshugoel/10-best-security-practices-in-android-applications-that-every-developer-must-know-99c8cd07c0bb
```
philipphofmann, line 27: `h`: The whole class is basically the same as in the article here. I would prefer if we have tests for this. I think we would have to change the code a little, but still, I would prefer it. If we don't it is a Blackbox that we just have to trust and it is hard to change, because you don't know if you break anything.

marandaneto: @philipphofmann ideas on how to test if a device is rooted or not? this is just heuristics and tests won't be deterministic if we run tests on the JVM and not on a real rooted/non-rooted device.

marandaneto: I can't mock a `Runtime.getRuntime()` for example

marandaneto: I can test a few parts of this PR though, not the whole thing.
I'm blocked by #420
I'd need `IBuildInfoProvider` to be able to mock eg `Build.TAGS`

philipphofmann: I think we would need to split the class up in a few components. We could create a class named `BuildTagsProvider`. I wouldn't put this functionality into `IBuildInfoProvider` because BuildTags is more of an Android thing and also this class most likely is only going to stay on Android.

We could delegate `Runtime.getRuntime().exec(su)` to another class. Then we could return a process of our choice and we could make sure that the `BufferedReader` and the error handling is working fine.

sentry-android-core/src/main/java/io/sentry/android/core/util/RootChecker.java
```diff
@@ -0,0 +1,147 @@
...
+        "com.devadvance.rootcloakplus",
+        "com.koushikdutta.superuser",
+        "com.thirdparty.superuser",
+        "eu.chainfire.supersu", // SuperSU
+        "com.noshufou.android.su" // superuser
+      };
+
+      for (final String pkg : packages) {
+        try {
+          pm.getPackageInfo(pkg, 0);
```
philipphofmann, line 139: `m`: Wouldn't `pm.getInstalledPackages(...);` be a better approach, because we avoid catching 6 exceptions.

sentry-android-core/src/main/java/io/sentry/android/core/util/RootChecker.java
philipphofmann: For the `PackageManager` we could create our own function or class that checks whether a package exists and write actual integration tests to make sure if it is working fine.

sentry-android-core/src/main/java/io/sentry/android/core/util/RootChecker.java
marandaneto: @philipphofmann I'd not like that an SDK checks all the installed packages of a device.
Also, this is arguable, what is more expensive? 6 exceptions or looping ~50 installed apps? can't say one or the other

sentry-android-core/src/main/java/io/sentry/android/core/util/RootChecker.java
marandaneto: @philipphofmann for PackageManager only mocking is fine, no need to create extra classes.

sentry-android-core/src/main/java/io/sentry/android/core/util/RootChecker.java
philipphofmann: Checking all installed packages is an argument for me, but looping through the packages I think has to be faster, because you just have one call to the package manager and then you are done. The other approach has 6 calls to the package manager and 6 exceptions, I think this should be slower. I'm fine with both ways.

marandaneto: > Checking all installed packages is an argument for me

@philipphofmann I don't get it, wdym?

philipphofmann: I mean that is an argument for keeping it like it is.

sentry-android-core/src/main/java/io/sentry/android/core/util/RootChecker.java
marandaneto: I did test all the cases, but I'm generally against creating classes for everything just for the sake of readability/testability.
I've created a private package ctor that takes all the necessary params, so I can mock and test them isolated.

PR #420, opened 2020-05-18
fix: ConnectivityChecker should return false if offline and no permission

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: ConnectivityChecker should return false if offline and no permission

ConnectivityChecker was assuming connected even if offline (because of permissions)

## :pencil: Checklist

- [X] All tests passing


sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
```diff
@@ -1,31 +1,161 @@
...
+        whenever(nc.hasTransport(eq(TRANSPORT_ETHERNET))).thenReturn(true)
+        val buildInfo = mock<IBuildInfoProvider>()
+        whenever(buildInfo.sdkInfoVersion).thenReturn(Build.VERSION_CODES.M)
+        assertEquals("ethernet", ConnectivityChecker.getConnectionType(contextMock, mock(), buildInfo))
     }
 
     @Test
-    fun `getConnectionType won't throw exception`() {
-        ConnectivityChecker.getConnectionType(context, mock())
+    fun `When network capabilities is TRANSPORT_CELLULAR, return wifi`() {
```
marandaneto, line 148:
```suggestion
    fun `When network capabilities has TRANSPORT_CELLULAR, return cellular`() {
```

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
```diff
@@ -1,31 +1,161 @@
...
+        whenever(nc.hasTransport(eq(TRANSPORT_WIFI))).thenReturn(true)
+        val buildInfo = mock<IBuildInfoProvider>()
+        whenever(buildInfo.sdkInfoVersion).thenReturn(Build.VERSION_CODES.M)
+        assertEquals("wifi", ConnectivityChecker.getConnectionType(contextMock, mock(), buildInfo))
     }
 
     @Test
-    fun `isConnected won't throw exception`() {
-        ConnectivityChecker.isConnected(context, mock())
+    fun `When network capabilities is TRANSPORT_ETHERNET, return wifi`() {
```
marandaneto, line 133:
```suggestion
    fun `When network capabilities has TRANSPORT_ETHERNET, return ethernet`() {
```

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
```diff
@@ -1,31 +1,161 @@
...
+        whenever(contextMock.getSystemService(any())).thenReturn(cm)
+        val net = mock<Network>()
+        whenever(cm.activeNetwork).thenReturn(net)
+        val buildInfo = mock<IBuildInfoProvider>()
+        whenever(buildInfo.sdkInfoVersion).thenReturn(Build.VERSION_CODES.M)
+        assertNull(ConnectivityChecker.getConnectionType(contextMock, mock(), buildInfo))
+    }
+
+    @Test
+    fun `When network capabilities is TRANSPORT_WIFI, return wifi`() {
```
marandaneto, line 118:
```suggestion
    fun `When network capabilities has TRANSPORT_WIFI, return wifi`() {
```

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
```diff
@@ -1,31 +1,161 @@
...
 class ConnectivityCheckerTest {
 
-    private lateinit var context: Context
+    @Test
+    fun `When network is active and connected, return true for isConnected`() {
+        val contextMock = mock<Context>()
+        val cm = mock<ConnectivityManager>()
+        whenever(contextMock.getSystemService(any())).thenReturn(cm)
+        val an = mock<NetworkInfo>()
+        whenever(cm.activeNetworkInfo).thenReturn(an)
```
philipphofmann, line 32: `m`: Why not extract the mock setup to `@Before`? Same could be done for `val net = mock<Network>()` and `val nc = mock<NetworkCapabilities>()`.

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
```diff
@@ -1,31 +1,161 @@
...
+import kotlin.test.assertTrue
 
-@RunWith(AndroidJUnit4::class)
 class ConnectivityCheckerTest {
 
-    private lateinit var context: Context
+    @Test
+    fun `When network is active and connected, return true for isConnected`() {
+        val contextMock = mock<Context>()
+        val cm = mock<ConnectivityManager>()
```
philipphofmann, line 29: `l`: Why not use long variable names here? Makes the code easier to read IMO.
```suggestion
        val connectivityManager = mock<ConnectivityManager>()
```

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
```diff
@@ -1,31 +1,161 @@
...
+        assertFalse(ConnectivityChecker.isConnected(contextMock, mock())!!)
+    }
+
+    @Test
+    fun `When there's no permission, return true for isConnected`() {
+        val contextMock = mock<Context>()
+        val cm = mock<ConnectivityManager>()
+        whenever(contextMock.getSystemService(any())).thenReturn(cm)
+        whenever(contextMock.checkPermission(any(), any(), any())).thenReturn(PERMISSION_DENIED)
+        assertTrue(ConnectivityChecker.isConnected(contextMock, mock())!!)
```
philipphofmann, line 54: `m`: Wouldn't it be better to return `null` or something else instead? Is it fine to assume we have an internet connection? If we keep it like this, I would definitely add this as a comment to `isConnected`.

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
```diff
@@ -23,12 +24,18 @@ private ConnectivityChecker() {}
...
+      final @NotNull Context context, final @NotNull ILogger logger) {
+    final ConnectivityManager connectivityManager = getConnectivityManager(context, logger);
     if (connectivityManager == null) {
       return null;
     }
-    return isConnected(context, connectivityManager, logger);
+    try {
+      return isConnected(context, connectivityManager, logger);
+    } catch (SecurityException ignored) {
+      // if no permission, we assume its connected.
```
philipphofmann, line 36: `m`: I would also add this comment to the doc of this method.

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
```diff
@@ -1,31 +1,161 @@
...
+        val buildInfo = mock<IBuildInfoProvider>()
+        whenever(buildInfo.sdkInfoVersion).thenReturn(Build.VERSION_CODES.M)
+        assertNull(ConnectivityChecker.getConnectionType(mock(), mock(), buildInfo))
+    }
+
+    @Test
+    fun `When sdkInfoVersion is not min Marshmallow, return null for getConnectionType`() {
+        val buildInfo = mock<IBuildInfoProvider>()
+        whenever(buildInfo.sdkInfoVersion).thenReturn(0)
+        assertNull(ConnectivityChecker.getConnectionType(mock(), mock(), buildInfo))
```
philipphofmann, line 81: `l`: As already pointed out maybe it is better to use an enum for the return result.

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
```diff
@@ -23,12 +24,18 @@ private ConnectivityChecker() {}
    *
    * @return true if the application has internet access
    */
-  public static @Nullable Boolean isConnected(@NotNull Context context, @NotNull ILogger logger) {
-    ConnectivityManager connectivityManager = getConnectivityManager(context, logger);
+  public static @Nullable Boolean isConnected(
```
philipphofmann, line 27: `m`: I am not a huge fan of using `null` as a return value. What about creating an enum for the result here. We have many different outcomes here: connected, not connected, min sdk not high enough, something went wrong. We could combine it too three different outcomes: `connected`, `notConnected` and `unknown`.

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
```diff
@@ -1,31 +1,161 @@
...
+
+    @Test
+    fun `When there's no permission, return null for getConnectionType`() {
+        val contextMock = mock<Context>()
+        val cm = mock<ConnectivityManager>()
+        whenever(contextMock.getSystemService(any())).thenReturn(cm)
+        whenever(contextMock.checkPermission(any(), any(), any())).thenReturn(PERMISSION_DENIED)
+        val buildInfo = mock<IBuildInfoProvider>()
+        whenever(buildInfo.sdkInfoVersion).thenReturn(Build.VERSION_CODES.M)
+        assertNull(ConnectivityChecker.getConnectionType(contextMock, mock(), buildInfo))
```
philipphofmann, line 92: `m`: What about return an empty string or`"unknown"` and/or also defining constants for `"wifi"`, `"ethernet"`, `"cellular"`, and `"unknown"`?

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
```diff
@@ -1,31 +1,161 @@
...
+
+    @Test
+    fun `When network capabilities has TRANSPORT_WIFI, return wifi`() {
+        val contextMock = mock<Context>()
+        val cm = mock<ConnectivityManager>()
+        whenever(contextMock.getSystemService(any())).thenReturn(cm)
+        val net = mock<Network>()
+        whenever(cm.activeNetwork).thenReturn(net)
+        val nc = mock<NetworkCapabilities>()
+        whenever(cm.getNetworkCapabilities(any())).thenReturn(nc)
```
philipphofmann, line 125: `m`: Can we please make use of a `@Before` method for all tests here. The tests are hard to read IMO. A little bit of a splitting up into either arrange, act, assert or given, when, then would be nice.

sentry-android-core/src/main/java/io/sentry/android/core/IBuildInfoProvider.java
```diff
@@ -0,0 +1,12 @@
+package io.sentry.android.core;
+
+/** To make the Build class testable */
+public interface IBuildInfoProvider {
```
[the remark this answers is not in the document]
marandaneto: @philipphofmann `final` yes, `static` no, only with `PowerMock` but it's too magic IMO

sentry-android-core/src/main/java/io/sentry/android/core/BuildInfoProvider.java
```diff
@@ -0,0 +1,21 @@
...
+
+import android.os.Build;
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+public final class BuildInfoProvider implements IBuildInfoProvider {
+
+  private static final BuildInfoProvider instance = new BuildInfoProvider();
+
+  public static IBuildInfoProvider getInstance() {
```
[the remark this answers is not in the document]
marandaneto: this class is ready to be extended eg by #417
`CurrentDateProvider` is also a singleton, just to avoid the expense of creating it multiple times.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -87,6 +93,11 @@ public DefaultAndroidEventProcessor(
     executorService.shutdown();
   }
 
+  public DefaultAndroidEventProcessor(
+      final @NotNull Context context, final @NotNull SentryOptions options) {
```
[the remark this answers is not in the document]
marandaneto: the other ctor has it, so I don't get it.

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
```diff
@@ -37,26 +44,29 @@ private ConnectivityChecker() {}
...
-      @NotNull ConnectivityManager connectivityManager,
-      @NotNull ILogger logger) {
+      final @NotNull Context context,
+      final @NotNull ConnectivityManager connectivityManager,
+      final @NotNull ILogger logger)
+      throws SecurityException {
     if (!Permissions.hasPermission(context, Manifest.permission.ACCESS_NETWORK_STATE)) {
       logger.log(SentryLevel.INFO, "No permission (ACCESS_NETWORK_STATE) to check network status.");
-      return null;
+      throw new SecurityException("No ACCESS_NETWORK_STATE permission");
```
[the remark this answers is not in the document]
marandaneto: good catch

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
```diff
@@ -69,10 +79,13 @@ private static Boolean isConnected(
...
-  @SuppressLint({"ObsoleteSdkInt", "MissingPermission"})
-  public static String getConnectionType(@NotNull Context context, @NotNull ILogger logger) {
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
-      ConnectivityManager connectivityManager = getConnectivityManager(context, logger);
+  @SuppressLint({"ObsoleteSdkInt", "MissingPermission", "NewApi"})
+  public static @Nullable String getConnectionType(
+      final @NotNull Context context,
+      final @NotNull ILogger logger,
+      final @NotNull IBuildInfoProvider buildInfoProvider) {
+    if (buildInfoProvider.getSdkInfoVersion() >= Build.VERSION_CODES.M) {
```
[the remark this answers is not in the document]
marandaneto: `getNetworkCapabilities` requires a `Network` and not a `NetworkInfo`

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
```diff
@@ -81,12 +94,12 @@ public static String getConnectionType(@NotNull Context context, @NotNull ILogge
             SentryLevel.INFO, "No permission (ACCESS_NETWORK_STATE) to check network status.");
         return null;
       }
-      Network activeNetwork = connectivityManager.getActiveNetwork();
+      final Network activeNetwork = connectivityManager.getActiveNetwork();
```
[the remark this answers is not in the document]
marandaneto: it is final already, not sure why you see non final

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
marandaneto: because those mocks are not used in all the tests, using `Before` would slow down tests that are not being used at all

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
marandaneto: no reason, mostly copy pasta, renamed it

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
marandaneto: we still want to be able to send events over the network if no permission is granted to check the network status, the connection can be active or not, we dont know.

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
marandaneto: added as a comment to `@return`, fwiw this is an internal class.

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
marandaneto: `event.contexts.device.connectionType` is a raw `String`, we can use an `enum`, but then we more complexibility with no big gain, it needs to be deserialized anyway.

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
marandaneto: min sdk is only for `getConnectionType` not for `isConnected`.
so its either connected, not connected, or unknown.
I'd prefer to also use `enums` all the type, but all of those types would need to be public.

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
marandaneto: we'll be serializing and sending over the network something useless IMO.
if you read a crash report and see `network type: unknown`, is this any useful?

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
marandaneto: splitting up makes sense, fwiw `act/assert` here are merged as we test a single method and a single assert

marandaneto: as stated before, `Before` won't apply when we don't use all the mocks in all the tests, we'd be paying for something not used

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
marandaneto: done

sentry-android-core/src/main/java/io/sentry/android/core/BuildInfoProvider.java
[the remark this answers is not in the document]
philipphofmann: Hm, I'm not a big fan of singletons as they often make the code harder to test and they hide dependencies. Singleton is a nice pattern, but you should use it sparingly. I would not worry about creating it multiple times. Now it stays in memory the whole time when even if you only use it once.

sentry-android-core/src/main/java/io/sentry/android/core/IBuildInfoProvider.java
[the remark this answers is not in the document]
philipphofmann: We only want to mock `getSdkInfoVersion` right? If `BuildInfoProvider` isn't a singleton then we can also mock it I guess. I wouldn't create the interface because of YAGNI.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
philipphofmann: Actually I would prefer to pass the `BuildInfoProvider` as a parameter. Then we could even delete this constructor and make the other public.

sentry-android-core/src/main/java/io/sentry/android/core/IBuildInfoProvider.java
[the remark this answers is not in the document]
marandaneto: @philipphofmann as stated in other PR, this is gonna be reused, so a new method is coming along, eg `Build.TAGS`

can you show me how to mock this with mockito?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
marandaneto: this only would make sense if `BuildInfoProvider` was not a singleton

sentry-android-core/src/main/java/io/sentry/android/core/BuildInfoProvider.java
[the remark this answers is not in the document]
marandaneto: its not only once, `DefaultAndroidEventProcessor` is triggered for every single event

philipphofmann:
I opened up a [PR](https://github.com/marandaneto/sentry-android/pull/1) on your repo to make use of `beforeTest` in `ConnectivityCheckerTest`. I prefer it because it moves the whole mocking setup to `beforeTest` and the test methods only contain what you are testing. If you like it merge it if not just close my PR.

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
philipphofmann: Still we have lots of duplicated code here and the tests are hard to read IMO. I opened up a PR on your forked repo to show you the difference.

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
philipphofmann: Now we send `network type: null`. With this I might ask myself why is it `null`? Is it working properly or do we not know. Sending `network type: unknown` is clearer IMO because it says we can't tell.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransportGate.java
```diff
@@ -23,9 +23,16 @@ public boolean isConnected() {
...
     // let's assume its connected if there's no permission or something as we can't really know
     // whether is online or not.
-    return connected != null ? connected : true;
+    switch (connectionStatus) {
+      case CONNECTED:
+      case UNKNOWN:
+      case NO_PERMISSION:
+        return true;
+      default:
+        return false;
```
philipphofmann, line 35: For me this is a huge improvement 👏

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
```diff
@@ -36,43 +45,49 @@ private ConnectivityChecker() {}
...
-    }
-    return activeNetwork.isConnected();
-  }
-
-  @SuppressWarnings({"deprecation", "MissingPermission"})
-  private static @Nullable android.net.NetworkInfo getActiveNetworkInfo(
-      @NotNull Context context,
-      @NotNull ConnectivityManager connectivityManager,
-      @NotNull ILogger logger) {
+  private static @NotNull ConnectionStatus isConnected(
```
philipphofmann, line 57: `l`: Now that we have the enum as return type maybe we should rename this to `getConnectionStatus` or something similar?

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
marandaneto: @philipphofmann we are not sending it `null`, we are not sending it at all, `null` values are omitted by default on serialization.
`Gson` needs `serializeNulls()` to serialize null values.

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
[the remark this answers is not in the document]
marandaneto: ahhh you mean Android `M`, I was thinking about the `LOGAF -> M` lol :)
Tomorrow I'll look into it

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
```diff
@@ -16,98 +17,163 @@
...
-        return null;
-      }
-      if (!Permissions.hasPermission(context, Manifest.permission.ACCESS_NETWORK_STATE)) {
-        logger.log(
-            SentryLevel.INFO, "No permission (ACCESS_NETWORK_STATE) to check network status.");
-        return null;
-      }
-      Network activeNetwork = connectivityManager.getActiveNetwork();
+  @SuppressLint({"ObsoleteSdkInt", "MissingPermission", "NewApi"})
+  public static @Nullable String getConnectionStatus(
```
philipphofmann, line 83: `m`: Shouldn't this still be `getConnectionStatus`?

marandaneto: yeah the IDE renamed this one too, damn, good catch

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -71,7 +71,8 @@ static void init(
 
     readDefaultOptionValues(options, context);
 
-    options.addEventProcessor(new DefaultAndroidEventProcessor(context, options));
+    options.addEventProcessor(
+        new DefaultAndroidEventProcessor(context, options, new BuildInfoProvider()));
```
philipphofmann, line 75: ❤️

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
```diff
@@ -16,98 +17,163 @@
...
+        wifi = true;
       }
       if (networkCapabilities.hasTransport(NetworkCapabilities.TRANSPORT_CELLULAR)) {
-        return "cellular";
+        cellular = true;
+      }
+    } else {
+      // ideally using connectivityManager.getAllNetworks(), but its >= Android L only
+
+      // for some reason linting didn't allow a single @SuppressWarnings("deprecation") on method
```
philipphofmann, line 124: Nice comment 🤙

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
```diff
@@ -16,98 +17,163 @@
...
+      final int type = activeNetworkInfo.getType();
+
+      @SuppressWarnings("deprecation")
+      final int TYPE_ETHERNET = ConnectivityManager.TYPE_ETHERNET;
+
+      @SuppressWarnings("deprecation")
+      final int TYPE_WIFI = ConnectivityManager.TYPE_WIFI;
+
+      @SuppressWarnings("deprecation")
+      final int TYPE_MOBILE = ConnectivityManager.TYPE_MOBILE;
```
philipphofmann, line 144: `l`: What about using them in the switch statement directly?

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
```diff
@@ -16,98 +17,163 @@
...
+        case TYPE_WIFI:
+          wifi = true;
+          break;
+        case TYPE_MOBILE:
+          cellular = true;
+          break;
       }
     }
+
+    // TODO: change the protocol to be a list of transports as a device may have the capability of
```
philipphofmann, line 159: This definitely would make sense 👍

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
```diff
@@ -16,98 +17,163 @@
...
-        return "cellular";
+        cellular = true;
+      }
+    } else {
+      // ideally using connectivityManager.getAllNetworks(), but its >= Android L only
+
+      // for some reason linting didn't allow a single @SuppressWarnings("deprecation") on method
+      // signature
+      @SuppressWarnings("deprecation")
+      final android.net.NetworkInfo activeNetworkInfo = connectivityManager.getActiveNetworkInfo();
```
philipphofmann, line 127: Awesome that we now also support `< M` 🥇

sentry-android-core/src/test/java/io/sentry/android/core/DefaultAndroidEventProcessorTest.kt
```diff
@@ -49,30 +50,38 @@ class DefaultAndroidEventProcessorTest {
...
     }
 
     @Test
     fun `when null options is provided, invalid argument is thrown`() {
         val clazz = Class.forName("io.sentry.android.core.DefaultAndroidEventProcessor")
-        val ctor = clazz.getConstructor(Context::class.java, SentryOptions::class.java)
-        val params = arrayOf(mock<Context>(), null)
+        val ctor = clazz.getConstructor(Context::class.java, SentryOptions::class.java, IBuildInfoProvider::class.java)
+        val params = arrayOf(mock<Context>(), null, null)
+        assertFailsWith<IllegalArgumentException> { ctor.newInstance(params) }
```
philipphofmann, line 71: Interesting workaround to test passing null values 😄

sentry-android-core/src/test/java/io/sentry/android/core/AndroidTransportGateTest.kt
```diff
@@ -1,45 +1,43 @@
...
-    @BeforeTest
-    fun `set up`() {
-        context = ApplicationProvider.getApplicationContext()
-        transportGate = AndroidTransportGate(context, mock())
+    private class Fixture {
+        fun getSut(): AndroidTransportGate {
+            return AndroidTransportGate(mock(), mock())
+        }
     }
+    private val fixture = Fixture()
```
philipphofmann, line 17: Nice 👍

sentry-android-core/src/test/java/io/sentry/android/core/ConnectivityCheckerTest.kt
```diff
@@ -1,31 +1,164 @@
...
+
+        assertEquals("cellular", ConnectivityChecker.getConnectionType(contextMock, mock(), buildInfo))
+    }
+
+    @Test
+    fun `When network is TYPE_MOBILE, return cellular`() {
+        whenever(buildInfo.sdkInfoVersion).thenReturn(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+        whenever(networkInfo.type).thenReturn(TYPE_MOBILE)
+
+        assertEquals("cellular", ConnectivityChecker.getConnectionType(contextMock, mock(), buildInfo))
```
philipphofmann, line 162: I like those tests. Clean and easy to understand. 💯

philipphofmann:
65 comments 😄

sentry-android-core/src/main/java/io/sentry/android/core/util/ConnectivityChecker.java
marandaneto: https://github.com/getsentry/sentry-android/pull/420#discussion_r427911518

that was my first try but cant use `@SuppressWarnings("deprecation")` on an enum

PR #422, opened 2020-05-19
ref: retrying thread pool executor is a single thread executor instead of scheduled thread executor

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
ref: retrying thread pool executor is a single thread pool executor instead of a scheduled thread pool executor.

it executes tasks one after the other and we don't have scheduled tasks at all.

running it, tests passed.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -80,6 +80,7 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
     this.context = Objects.requireNonNull(context, "The application context is required.");
     this.options = Objects.requireNonNull(options, "The SentryOptions is required.");
 
+    // TODO: Executor could be ISentryExecutorService
     ExecutorService executorService = Executors.newSingleThreadExecutor();
     // dont ref. to method reference, theres a bug on it
     contextData = executorService.submit(() -> loadContextData());
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -80,7 +80,6 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
     this.context = Objects.requireNonNull(context, "The application context is required.");
     this.options = Objects.requireNonNull(options, "The SentryOptions is required.");
 
-    // TODO: Executor could be ISentryExecutorService
     ExecutorService executorService = Executors.newSingleThreadExecutor();
     // dont ref. to method reference, theres a bug on it
     contextData = executorService.submit(() -> loadContextData());
--- a/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
+++ b/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
@@ -7,6 +7,7 @@ import io.sentry.core.hints.Flushable;
 import io.sentry.core.hints.Retryable;
 import io.sentry.core.hints.SubmissionResult;
 import io.sentry.core.util.CollectionUtils;
+import io.sentry.core.util.LogUtils;
 import io.sentry.core.util.Objects;
 import java.io.BufferedInputStream;
 import java.io.BufferedReader;
@@ -80,10 +81,7 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
           }
         }
       } else {
-        logger.log(
-            SentryLevel.DEBUG,
-            "%s is not Retryable",
-            hint != null ? hint.getClass().getCanonicalName() : "Hint");
+        LogUtils.logIfNotRetryable(logger, hint);
       }
     }
   }
@@ -147,17 +145,15 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
                     "Timed out waiting for event submission: %s",
                     event.getEventId());
 
-                if (hint instanceof Retryable) {
-                  ((Retryable) hint).setRetry(true);
-                }
+                //                TODO: find out about the time out
+                //                if (hint instanceof Retryable) {
+                //                  ((Retryable) hint).setRetry(true);
+                //                }
 
                 break;
               }
             } else {
-              logger.log(
-                  SentryLevel.DEBUG,
-                  "%s is not Flushable",
-                  hint != null ? hint.getClass().getCanonicalName() : "Hint");
+              LogUtils.logIfNotFlushable(logger, hint);
             }
           }
         } catch (Exception e) {
@@ -187,18 +183,16 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
                     "Timed out waiting for item submission: %s",
                     session.getSessionId());
 
-                if (hint instanceof Retryable) {
-                  ((Retryable) hint).setRetry(true);
-                }
+                //                TODO: find out about the time out
+                //                if (hint instanceof Retryable) {
+                //                  ((Retryable) hint).setRetry(true);
+                //                }
 
                 break;
               }
               logger.log(SentryLevel.DEBUG, "Flushed %d item.", items);
             } else {
-              logger.log(
-                  SentryLevel.DEBUG,
-                  "%s is not Flushable",
-                  hint != null ? hint.getClass().getCanonicalName() : "Hint");
+              LogUtils.logIfNotFlushable(logger, hint);
             }
           }
         } catch (Exception e) {
@@ -221,10 +215,7 @@ public final class EnvelopeSender extends DirectoryProcessor implements IEnvelop
           break;
         }
       } else {
-        logger.log(
-            SentryLevel.DEBUG,
-            "%s is not SubmissionResult",
-            hint != null ? hint.getClass().getCanonicalName() : "Hint");
+        LogUtils.logIfNotSubmissionResult(logger, hint);
       }
     }
   }
--- a/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
+++ b/sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
@@ -1,6 +1,5 @@
 package io.sentry.core;
 
-import static io.sentry.core.SentryLevel.DEBUG;
 import static io.sentry.core.SentryLevel.ERROR;
 import static io.sentry.core.cache.SessionCache.PREFIX_CURRENT_SESSION_FILE;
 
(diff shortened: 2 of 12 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -80,6 +80,7 @@ public DefaultAndroidEventProcessor(
     this.context = Objects.requireNonNull(context, "The application context is required.");
     this.options = Objects.requireNonNull(options, "The SentryOptions is required.");
 
+    // TODO: Executor could be ISentryExecutorService
```
marandaneto, line 83: if we want to reuse ISentryExecutorService, getter SentryOptions.getExecutorService should be public

sentry-core/src/main/java/io/sentry/core/cache/SessionCache.java
```diff
@@ -96,6 +96,9 @@ public void store(final @NotNull SentryEnvelope envelope, final @Nullable Object
     }
 
     if (hint instanceof SessionStart) {
+
+      // TODO: should we move this to AppLifecycleIntegration? and do on SDK init? but it's too much
```
marandaneto, line 100: I'd rather not do it or we find a different way to improve things.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
bruno-garcia: I'd rather we don't make any of this public

sentry-core/src/main/java/io/sentry/core/transport/QueuedThreadPoolExecutor.java
```diff
@@ -34,13 +31,20 @@
...
    */
-  public RetryingThreadPoolExecutor(
+  public QueuedThreadPoolExecutor(
       final int corePoolSize,
       final int maxQueueSize,
       final @NotNull ThreadFactory threadFactory,
       final @NotNull RejectedExecutionHandler rejectedExecutionHandler) {
-
-    super(corePoolSize, threadFactory, rejectedExecutionHandler);
+    // similar to newSingleThreadExecutor.newSingleThreadExecutor, but with a max queue size control
```
marandaneto, line 39:
```suggestion
    // similar to Executors.newSingleThreadExecutor, but with a max queue size control
```

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
```diff
@@ -144,13 +146,18 @@ private void processEnvelope(final @NotNull SentryEnvelope envelope, final @Null
                     SentryLevel.WARNING,
                     "Timed out waiting for event submission: %s",
                     event.getEventId());
+
+                if (hint instanceof Retryable) {
```
bruno-garcia, line 150: H: Not sure about this. It could mean we're capturing events twice if we mark the event as retry=true and it actually finishes being sent. Why is this change introduced? We don't seem to have an issue missing events, right?

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
```diff
@@ -179,11 +186,19 @@ private void processEnvelope(final @NotNull SentryEnvelope envelope, final @Null
                     SentryLevel.WARNING,
                     "Timed out waiting for item submission: %s",
                     session.getSessionId());
+
+                if (hint instanceof Retryable) {
```
bruno-garcia, line 190: Should we pull the `if (hint instanceof Flushable) {` logic into some function and reuse it? The fact we need to change it multiple times is a bad sign

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
```diff
@@ -179,11 +186,19 @@ private void processEnvelope(final @NotNull SentryEnvelope envelope, final @Null
...
+
                 break;
               }
               logger.log(SentryLevel.DEBUG, "Flushed %d item.", items);
             } else {
-              logger.log(DEBUG, "Envelope %s is not Flushable.", envelope.getHeader().getEventId());
+              logger.log(
+                  SentryLevel.DEBUG,
+                  "%s is not Flushable",
+                  hint != null ? hint.getClass().getCanonicalName() : "Hint");
```
bruno-garcia, line 201: L: Only if we had extension method eh? But what about a private static method here like `logHintName` but with a decent name?

Or event two methods: `logNotFlushable` and `logNotRetriable`

sentry-core/src/main/java/io/sentry/core/SentryOptions.java
```diff
@@ -920,6 +920,11 @@ public void setBypassSecurity(boolean bypassSecurity) {
 
   /** SentryOptions ctor It adds and set default things */
   public SentryOptions() {
+    // UncaughtExceptionHandlerIntegration should be inited before everything.
+    // if there's an error on the setup, we are able to capture it
+    integrations.add(new UncaughtExceptionHandlerIntegration());
```
bruno-garcia, line 925: worth noting this only adds the instance. `register` will be called later on so adding the instance to the list vs initializing the ExecutionService doesn't really make a difference.

L: Maybe better adding the executtorService initialization above this so it's not confusing for the reader?

sentry-core/src/main/java/io/sentry/core/cache/SessionCache.java
bruno-garcia: fwiw on sentry-cocoa it's done on the thread calling `init` (expected to be the UI thread).
Agree it'd be best to avoid work on the UI thread but we need to make sure we don't race through the current session file which `startSession` would touch upon being called

sentry-core/src/main/java/io/sentry/core/transport/AsyncConnection.java
```diff
@@ -359,8 +368,8 @@ public void run() {
               .getLogger()
               .log(
```
bruno-garcia, line 369: M: More convinced now that it would be nice to have some static log methods we could import and use here, that take `@Nullable Object hint` and log these, we use it a lot

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: I'll remove the comment and close this then, was just trying to reuse the executor

sentry-core/src/main/java/io/sentry/core/cache/SessionCache.java
marandaneto: if I do that before even calling the Lifeycicle callback, no racing.
But sentry-android does more things, like the crash marker from sentry-native with a timestamp, so a hard crash can lead to a few IOs more on the main-thread and at this point, I'd say that we are doing quite a bit.

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
marandaneto: it's a timed out and not a known exception that returns immediately which is 99.9% of the cases (no connection or something), so it really didn't have time enough due to different things, eg, long queue, very slow connection and low-end device, taking more time to serialize/deserialize, etc, I did reproduce this using emulator and very bad configurations purposely

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
marandaneto: makes sense, needs to be public though :P u need to give up of one of them

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
marandaneto: not private though, as the same pattern is being used somewhere else

sentry-core/src/main/java/io/sentry/core/SentryOptions.java
marandaneto: yeah comparing to the ExecutionService, true.
I meant to be before every other integration.

sentry-core/src/main/java/io/sentry/core/transport/AsyncConnection.java
marandaneto: yes, all of them are gonna be public though, they are being used in different packages.

marandaneto:
> Also, it's worth noting now reading `EnvelopeSender` it does look like it needs refactoring.
> Since we're unwrapping the envelope there and special casing things,maybe it's time to move that logic to the transport. In these `Senders` we just call `captureEnvelope` and in there we either call the envelope endpoint, or we unwrap them (code from current `EnvelopeSender`) to call the `/store` endpoint)

yeah, I believe that when we get to the point of getting rid of /store, this refactoring will come naturally, what do you think?

> Main point to discuss is the `Retryable` hint bit. We need to be really conservative here because it can lead to different problems.

replied on the comment.

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
bruno-garcia: It times out waiting for the synchronization mechanism to be called as opposed to an HTTP request timeout. It's very possible that we'll give up waiting here before the HTTP request times out (unless we guarantee that the HTTP request times out before this `waitFlush` method here.

It's also an issue where have some bug in the SDK and now we're keeping things forever because this method is never called.

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
bruno-garcia: I guess we just make it public then and add that `@InternalApi` annotation. or a `io.sentry.internal` package which we can document (or don't even need to) that we'll break without bumping a major version

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
bruno-garcia: True

sentry-core/src/main/java/io/sentry/core/SentryOptions.java
bruno-garcia: true

sentry-core/src/main/java/io/sentry/core/cache/SessionCache.java
bruno-garcia: Sure thing, your call

sentry-core/src/main/java/io/sentry/core/EnvelopeSender.java
marandaneto: I'm not talking about only http request timeout, it's altogether.

marandaneto: I removed those `Retryable` checks from this PR and can come up with a new PR, merging this one without them

PR #423, opened 2020-05-19
enha: end session when user kills the App.

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing


marandaneto:
@philipphofmann are you aware of any side effects of doing it or even different behaviors that won't work?
I know that a few brands/models don't call `onTaskRemoved` but it's not the case for the majority and it's a best-effort solution.

PR #424, opened 2020-05-19
fix: add Abnormal to the Session.State enum as its part of the protocol

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: add Abnormal to the Session.State enum as its part of the protocol

this would make the current version be compatible with older generated sessions data.
but it's not currently being used by the SDK.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Session.java
+++ b/sentry-core/src/main/java/io/sentry/core/Session.java
@@ -14,7 +14,7 @@ public final class Session {
     Ok,
     Exited,
     Crashed,
-    Abnormal // not used by mobile SDKs
+    Abnormal // not currently used in this SDK.
   }
 
   /** started timestamp */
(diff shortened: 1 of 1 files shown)


sentry-core/src/main/java/io/sentry/core/Session.java
```diff
@@ -13,7 +13,8 @@
   public enum State {
     Ok,
     Exited,
-    Crashed
+    Crashed,
+    Abnormal // not used by mobile SDKs
```
bruno-garcia, line 17: L: Abnormal is still used on sentry-cocoa so current comment is a bit misleading. Maybe:

```suggestion
    Abnormal // not currently used in this SDK.
```

PR #433, opened 2020-05-26
feat: make gzip as default content encoding type

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
feat: make gzip as default content encoding type

compress the content of the payload before sending it over the network, so saving some data.

running and tests passed.
serialization layer is not affected and so didn't need extra tests.
the response content is still a raw JSON, no need to change that as we only get a very small response when there's a 4xx error for example.

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
CI error is due to missing NDK on LGTM analyzer.
https://discuss.lgtm.com/t/android-ndk-v21-1-6352462-is-not-available/2910

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
+++ b/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
@@ -310,8 +310,9 @@ public class HttpTransport implements ITransport {
       connection.getInputStream().close();
     } catch (IOException e) {
       logger.log(ERROR, e, "Error while closing the connection.");
+    } finally {
+      connection.disconnect();
     }
-    connection.disconnect();
   }
 
   /**
--- a/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
+++ b/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
@@ -14,7 +14,6 @@ import io.sentry.core.util.Objects;
 import io.sentry.core.util.StringUtils;
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -160,13 +159,12 @@ public class HttpTransport implements ITransport {
 
     int responseCode = -1;
 
-    try (final ByteArrayOutputStream byteOut = new ByteArrayOutputStream();
-        final OutputStream outputStream = connection.getOutputStream();
-        final GZIPOutputStream gzip = new GZIPOutputStream(byteOut);
+    try (final OutputStream outputStream = connection.getOutputStream();
+        final GZIPOutputStream gzip = new GZIPOutputStream(outputStream);
         final Writer writer = new BufferedWriter(new OutputStreamWriter(gzip, UTF_8))) {
       serializer.serialize(event, writer);
 
-      writeAndFlushStreams(connection, byteOut, outputStream, gzip);
+      finishAndFlushStreams(outputStream, gzip);
       options.getLogger().log(DEBUG, "Event sent %s successfully.", event.getEventId());
       return TransportResult.success();
     } catch (IOException e) {
@@ -279,13 +277,12 @@ public class HttpTransport implements ITransport {
 
     int responseCode = -1;
 
-    try (final ByteArrayOutputStream byteOut = new ByteArrayOutputStream();
-        final OutputStream outputStream = connection.getOutputStream();
-        final GZIPOutputStream gzip = new GZIPOutputStream(byteOut);
+    try (final OutputStream outputStream = connection.getOutputStream();
+        final GZIPOutputStream gzip = new GZIPOutputStream(outputStream);
         final Writer writer = new BufferedWriter(new OutputStreamWriter(gzip, UTF_8))) {
       serializer.serialize(envelope, writer);
 
-      writeAndFlushStreams(connection, byteOut, outputStream, gzip);
+      finishAndFlushStreams(outputStream, gzip);
       options.getLogger().log(DEBUG, "Envelope sent successfully.");
       return TransportResult.success();
     } catch (IOException e) {
@@ -302,26 +299,17 @@ public class HttpTransport implements ITransport {
   }
 
   /**
-   * Writes and flushes the streams
+   * Finishes and flushes the streams
    *
-   * @param connection the HttpURLConnection
-   * @param byteOut the ByteArrayOutputStream
    * @param outputStream the OutputStream
    * @param gzip the GZIPOutputStream
    * @throws IOException throws IOException if error while manipulating the streams
    */
-  private void writeAndFlushStreams(
-      final @NotNull HttpURLConnection connection,
-      final @NotNull ByteArrayOutputStream byteOut,
-      final @NotNull OutputStream outputStream,
-      final @NotNull GZIPOutputStream gzip)
+  private void finishAndFlushStreams(
+      final @NotNull OutputStream outputStream, final @NotNull GZIPOutputStream gzip)
       throws IOException {
     gzip.finish();
-    outputStream.write(byteOut.toByteArray());
     outputStream.flush();
-
-    // need to also close the input stream of the connection
-    connection.getInputStream().close();
   }
 
   /**
--- a/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
+++ b/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
@@ -6,6 +6,7 @@ import static io.sentry.core.SentryLevel.INFO;
 import static io.sentry.core.SentryLevel.WARNING;
 
 import com.jakewharton.nopen.annotation.Open;
+import io.sentry.core.ILogger;
 import io.sentry.core.ISerializer;
 import io.sentry.core.SentryEnvelope;
 import io.sentry.core.SentryEvent;
@@ -85,6 +86,8 @@ public class HttpTransport implements ITransport {
 
   private final @NotNull ICurrentDateProvider currentDateProvider;
 
+  private final @NotNull ILogger logger;
+
   /**
    * Constructs a new HTTP transport instance. Notably, the provided {@code requestUpdater} must set
    * the appropriate content encoding header for the {@link io.sentry.core.ISerializer} instance
@@ -132,6 +135,7 @@ public class HttpTransport implements ITransport {
     this.bypassSecurity = bypassSecurity;
     this.currentDateProvider =
         Objects.requireNonNull(currentDateProvider, "CurrentDateProvider is required.");
+    this.logger = Objects.requireNonNull(options.getLogger(), "Logger is required.");
 
     try {
       final URI uri = sentryUrl.toURI();
@@ -164,7 +168,7 @@ public class HttpTransport implements ITransport {
         final Writer writer = new BufferedWriter(new OutputStreamWriter(gzip, UTF_8))) {
       serializer.serialize(event, writer);
 
-      options.getLogger().log(DEBUG, "Event sent %s successfully.", event.getEventId());
+      logger.log(DEBUG, "Event sent %s successfully.", event.getEventId());
       return TransportResult.success();
     } catch (IOException e) {
       final TransportResult errorResult = getResultFromErrorCode(connection, e);
@@ -173,7 +177,7 @@ public class HttpTransport implements ITransport {
     } finally {
       updateRetryAfterLimits(connection, responseCode);
 
-      closeAndDisconnectConnection(connection);
+      closeAndDisconnect(connection);
     }
   }
 
@@ -281,7 +285,7 @@ public class HttpTransport implements ITransport {
         final Writer writer = new BufferedWriter(new OutputStreamWriter(gzip, UTF_8))) {
       serializer.serialize(envelope, writer);
 
-      options.getLogger().log(DEBUG, "Envelope sent successfully.");
+      logger.log(DEBUG, "Envelope sent successfully.");
       return TransportResult.success();
     } catch (IOException e) {
       final TransportResult errorResult = getResultFromErrorCode(connection, e);
@@ -292,7 +296,7 @@ public class HttpTransport implements ITransport {
     } finally {
       updateRetryAfterLimits(connection, responseCode);
 
-      closeAndDisconnectConnection(connection);
+      closeAndDisconnect(connection);
     }
   }
 
@@ -301,11 +305,11 @@ public class HttpTransport implements ITransport {
    *
    * @param connection the HttpURLConnection
    */
-  private void closeAndDisconnectConnection(final @NotNull HttpURLConnection connection) {
+  private void closeAndDisconnect(final @NotNull HttpURLConnection connection) {
     try {
       connection.getInputStream().close();
     } catch (IOException e) {
-      options.getLogger().log(ERROR, e, "Error while closing the connection.");
+      logger.log(ERROR, e, "Error while closing the connection.");
     }
     connection.disconnect();
   }
@@ -325,7 +329,7 @@ public class HttpTransport implements ITransport {
       logErrorInPayload(connection);
       return TransportResult.error(responseCode);
     } catch (IOException responseCodeException) {
-      options.getLogger().log(ERROR, responseCodeException, "Error getting responseCode");
+      logger.log(ERROR, responseCodeException, "Error getting responseCode");
 
       // this should not stop us from continuing.
       return getDefaultErrorResult(ioException);
@@ -339,12 +343,8 @@ public class HttpTransport implements ITransport {
    * @return the TransportResult error
    */
   private @NotNull TransportResult getDefaultErrorResult(final @NotNull Exception exception) {
-    options
-        .getLogger()
-        .log(
-            WARNING,
-            "Failed to obtain error message while analyzing event send failure.",
-            exception);
+    logger.log(
+        WARNING, "Failed to obtain error message while analyzing event send failure.", exception);
     return TransportResult.error();
   }
 
@@ -408,7 +408,7 @@ public class HttpTransport implements ITransport {
                 try {
                   dataCategory = DataCategory.valueOf(StringUtils.capitalize(catItem));
                 } catch (IllegalArgumentException e) {
-                  options.getLogger().log(INFO, e, "Unknown category: %s", catItem);
+                  logger.log(INFO, e, "Unknown category: %s", catItem);
                 }
                 // we dont apply rate limiting for unknown categories
                 if (DataCategory.Unknown.equals(dataCategory)) {
@@ -480,7 +480,7 @@ public class HttpTransport implements ITransport {
         errorMessage = "An exception occurred while submitting the event to the Sentry server.";
       }
 
-      options.getLogger().log(DEBUG, errorMessage);
+      logger.log(DEBUG, errorMessage);
     }
   }
 
@@ -507,9 +507,7 @@ public class HttpTransport implements ITransport {
       }
       return sb.toString();
     } catch (IOException e) {
-      options
-          .getLogger()
-          .log(ERROR, e, "Exception while reading the error message from the connection");
+      logger.log(ERROR, e, "Exception while reading the error message from the connection");
     }
     return null;
   }
--- a/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
+++ b/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
@@ -38,7 +38,7 @@ import org.jetbrains.annotations.Nullable;
 
 /**
  * An implementation of the {@link ITransport} interface that sends the events to the Sentry server
- * over HTTP(S) in gzip encoding.
+ * over HTTP(S).
  */
 @Open
 @ApiStatus.NonExtendable // only not final because of testing
--- a/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
+++ b/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
@@ -14,6 +14,7 @@ import io.sentry.core.util.Objects;
 import io.sentry.core.util.StringUtils;
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -30,6 +31,7 @@ import java.nio.charset.Charset;
 import java.util.Date;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.zip.GZIPOutputStream;
 import javax.net.ssl.HttpsURLConnection;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
@@ -37,7 +39,7 @@ import org.jetbrains.annotations.Nullable;
 
 /**
  * An implementation of the {@link ITransport} interface that sends the events to the Sentry server
- * over HTTP(S) in UTF-8 encoding.
+ * over HTTP(S) in gzip encoding.
  */
 @Open
 @ApiStatus.NonExtendable // only not final because of testing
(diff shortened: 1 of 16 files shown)


sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -30,14 +31,15 @@
...
+import java.util.zip.GZIPOutputStream;
 import javax.net.ssl.HttpsURLConnection;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 /**
  * An implementation of the {@link ITransport} interface that sends the events to the Sentry server
- * over HTTP(S) in UTF-8 encoding.
+ * over HTTP(S) in gzip encoding.
```
rhcarvalho, line 42: gzip and UTF-8 are orthogonal concepts.
The SDK is sending a UTF-8-encoded JSON representation of an event compressed with gzip 😄 

```suggestion
 * over HTTP(S).
```

Is it relevant to talk about UTF-8/JSON/gzip in the javadoc?! I suspect not.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -158,38 +160,19 @@ public HttpTransport(
 
     int responseCode = -1;
 
-    try (final OutputStream outputStream = connection.getOutputStream();
-        final Writer writer = new BufferedWriter(new OutputStreamWriter(outputStream, UTF_8))) {
+    try (final ByteArrayOutputStream byteOut = new ByteArrayOutputStream();
+        final OutputStream outputStream = connection.getOutputStream();
+        final GZIPOutputStream gzip = new GZIPOutputStream(byteOut);
+        final Writer writer = new BufferedWriter(new OutputStreamWriter(gzip, UTF_8))) {
```
rhcarvalho, line 166: I'm not familiar with the stack used in the Java SDK, so perhaps the best I can contribute is to ask whether it is possible / you considered plugging in streams directly instead of using a new byte array buffer for every event?

I suspect it would create a lot of garbage. Maybe not too relevant if sending occasional errors, but can be relevant for code running more often like transactions and maybe sessions (not very familiar with how sessions work though).

In Go, it would be a good use case for [`sync.Pool`](https://golang.org/pkg/sync/#Pool), which is a way to amortize allocation overhead. There must be something similar in Java.

So there is really two independent considerations here:

1. Streaming without in-memory buffering
2. Object reuse / pools to amortize allocation overhead

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -272,7 +255,7 @@ public boolean isRetryAfter(final @NotNull String itemType) {
     connection.setRequestMethod("POST");
     connection.setDoOutput(true);
 
-    connection.setRequestProperty("Content-Encoding", "UTF-8");
+    connection.setRequestProperty("Content-Encoding", "gzip");
```
rhcarvalho, line 258: 👍 

"UTF-8" is not a standard value for `Content-Encoding`
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding

And JSON is "by default" encoded in UTF-8 https://tools.ietf.org/html/rfc8259

Normally the character encoding (charset) is part of the `Content-Type` https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type

marandaneto: that's a very good point, it's been like this since ever though.

rhcarvalho: It didn't hurt, but also didn't do anything useful :)

marandaneto: yeah I think this came from sentry-java

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
bruno-garcia: [in .NET we have an option to use buffered or not](https://github.com/getsentry/sentry-dotnet/blob/04d756b024a2ee00bb5d2748630ba704349b4f88/src/Sentry/Internal/Http/DefaultSentryHttpClientFactory.cs#L76-L87). This came in after some environments that couldn't use the non-buffered approach IIRC.

No byte array is used directly here so that we can swap out with a pooled version, or am I missing something?

`ByteArrayOutputStream` can be using object pooling under the hood though.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -296,47 +279,95 @@ public boolean isRetryAfter(final @NotNull String itemType) {
...
+      final @NotNull ByteArrayOutputStream byteOut,
+      final @NotNull OutputStream outputStream,
+      final @NotNull GZIPOutputStream gzip)
+      throws IOException {
+    gzip.finish();
+    outputStream.write(byteOut.toByteArray());
+    outputStream.flush();
+
+    // need to also close the input stream of the connection
+    connection.getInputStream().close();
```
bruno-garcia, line 324: H: Should we take the `InputStream` here instead of the connection?

if it throws above, do we leak resources here? DO we need this in a `finally` block?

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: I'll have a look if its possible to make a bit more optimal, unfortunately, we are very limited on JDK 6 usage on many things due to the Android runtime, otherwise, we'd need to have transport for Android and one for sentry-java.

Object pooling is not a thing for Java and there are a few things out there but we can't really use it. eg https://commons.apache.org/proper/commons-pool/ or Streams, Slices, etc..

gues I can also use `BufferedOutputStream` instead of `ByteArrayOutputStream` directly.
Or even changing the serialization layer to write directly to this stream, the later would require a bigger change.

let's what what I can find, thanks for the review btw.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: that's true, yeah let's clean this up.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: this method is called inside a https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html
`OutputStream` implements `Closeable` so it's not an issue, but it does make sense to take the `InputStream` and I'll do it.
I've used the IDE to extract this to a method so it went "as before"

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: ok so no `ByteArrayOutputStream` anymore, it writes directly in the stream of the connection.
`GZIPOutputStream` already has its own buffering and I'll keep its default values.

The session works this way: a session starts when the user opens the App and ends when the user goes to the background and remains idle for a period, so no worries here, it's by far not as much as transactions.

Maybe in a later stage, we might need to refactor this a little bit to be more optimal, but we'll be ended up with different transports for Mobile and Backend, as we can't be cheap and effective at the same time (eg right now we open/close connections for each request), this is ~good for mobile, but might not be good for the backend.

This is a lot better than before, thanks for the feedback, if no more concerns, I'd like to resolve this issue.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -159,37 +160,17 @@ public HttpTransport(
     int responseCode = -1;
 
     try (final OutputStream outputStream = connection.getOutputStream();
-        final Writer writer = new BufferedWriter(new OutputStreamWriter(outputStream, UTF_8))) {
+        final GZIPOutputStream gzip = new GZIPOutputStream(outputStream);
+        final Writer writer = new BufferedWriter(new OutputStreamWriter(gzip, UTF_8))) {
```
rhcarvalho, line 164: TIL, I didn't know this "try-with-resources" https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html 👍 

```java
try (...) { ... }
```

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -159,37 +160,17 @@ public HttpTransport(
...
       serializer.serialize(event, writer);
 
-      // need to also close the input stream of the connection
-      connection.getInputStream().close();
+      finishAndFlushStreams(outputStream, gzip);
```
rhcarvalho, line 167: If I understood "try-with-resources", it will call `outputStream.close()` when the try block is over, and `close` calls `flush` according to https://docs.oracle.com/javase/8/docs/api/index.html?java/io/OutputStreamWriter.html.

Do we need to flush it in `finishAndFlushStreams`?!

Similarly, `GZIPOutputStream` inherits `close` from https://docs.oracle.com/javase/8/docs/api/index.html?java/io/OutputStreamWriter.html, which is documented to have the same effect as `finish` + closing the underlying stream (which is again our `outputStream`).

marandaneto: @rhcarvalho you're right, but https://developer.android.com/reference/android/os/StrictMode for some reason complains that there's something open and it should be closed, also a few issues on SO and people said to ignore it, but I prefer to `flush` it again for the sake of having a clean log for our customers.

marandaneto: Screenshot 2020-05-27 at 12 21 18

rhcarvalho: Hmm the log message talks about a _response body_, and what we are finishing and flushing here are streams related to the **request** body, no?!

rhcarvalho: Where is the response body being closed?

marandaneto: @rhcarvalho this warning was noticed before this PR and the response stream was always closed:
https://github.com/getsentry/sentry-android/blob/master/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java#L166

I changed again, let's see if I can experience the log again or not, it's intermittent, but when debugging you can check the status of the streams and they are all closed at least.

rhcarvalho: 👍  Will let Bruno have his final look.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -30,14 +30,15 @@
...
+import java.util.zip.GZIPOutputStream;
 import javax.net.ssl.HttpsURLConnection;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 /**
  * An implementation of the {@link ITransport} interface that sends the events to the Sentry server
- * over HTTP(S) in UTF-8 encoding.
+ * over HTTP(S).
```
philipphofmann, line 41: `l`: is it worth mentioning here that we use gzip?

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -297,46 +277,83 @@ public boolean isRetryAfter(final @NotNull String itemType) {
     int responseCode = -1;
 
     try (final OutputStream outputStream = connection.getOutputStream();
-        final Writer writer = new BufferedWriter(new OutputStreamWriter(outputStream, UTF_8))) {
+        final GZIPOutputStream gzip = new GZIPOutputStream(outputStream);
+        final Writer writer = new BufferedWriter(new OutputStreamWriter(gzip, UTF_8))) {
```
philipphofmann, line 281: `m`: The whole function here almost looks the same as `send(SentryEvent)`. Can we maybe refactor this so we don't have it duplicated?

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -297,46 +277,83 @@ public boolean isRetryAfter(final @NotNull String itemType) {
...
+      closeAndDisconnectConnection(connection);
+    }
+  }
+
+  /**
+   * Closes the Response stream and disconnect the connection
+   *
+   * @param connection the HttpURLConnection
+   */
+  private void closeAndDisconnectConnection(final @NotNull HttpURLConnection connection) {
```
philipphofmann, line 304: `l`: Maybe we can remove the connection in the name and rename it to `closeAndDisconnect`, because calling the function looks like this `closeAndDisconnectConnection(connection)` and there connection is twice. That's some unnecessary noise for me.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -297,46 +277,83 @@ public boolean isRetryAfter(final @NotNull String itemType) {
...
+    try {
+      connection.getInputStream().close();
+    } catch (IOException e) {
+      options.getLogger().log(ERROR, e, "Error while closing the connection.");
+    }
+    connection.disconnect();
```
philipphofmann, line 310: `m`: I think currently it doesn't make a big difference, because we already call this in `finally` and we don't have any return statements in this function, but wouldn't it be safer to wrap this in `finally`?

philipphofmann: `h`: I think it is not best practice to disconnect the connection for every single envelope and event that we send. It is better to keep the connection alive for some time. For each new connection, a handshake needs to be made, a TCP connection needs to be set up, etc. Read more about this [here](https://docs.oracle.com/javase/6/docs/technotes/guides/net/http-keepalive.html). Maybe we can keep the connection open for some time, but I guess this is out of scope in this PR. I just wanted to mention it. 
Maybe we also have good reasons not to do so that I am not aware of.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -297,46 +277,83 @@ public boolean isRetryAfter(final @NotNull String itemType) {
...
       logErrorInPayload(connection);
       return TransportResult.error(responseCode);
-    } catch (Exception e) {
-      options
-          .getLogger()
-          .log(WARNING, "Failed to obtain error message while analyzing event send failure.", e);
-      return TransportResult.error(-1);
-    } finally {
-      updateRetryAfterLimits(connection, responseCode);
+    } catch (IOException responseCodeException) {
```
philipphofmann, line 327: Catching only specific exceptions. I like it 😃

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -297,46 +277,83 @@ public boolean isRetryAfter(final @NotNull String itemType) {
...
 
+  /**
+   * Logs a default error message and return the TransportResult error with -1 responseCode
+   *
+   * @param exception the Exception
+   * @return the TransportResult error
+   */
+  private @NotNull TransportResult getDefaultErrorResult(final @NotNull Exception exception) {
+    options
+        .getLogger()
```
philipphofmann, line 343: `l`: We call ` options.getLogger()` quite often. Maybe it makes sense to call it once in the constructor and keep the logger as a property?

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -297,46 +277,83 @@ public boolean isRetryAfter(final @NotNull String itemType) {
...
+      return getDefaultErrorResult(ioException);
     }
   }
 
+  /**
+   * Logs a default error message and return the TransportResult error with -1 responseCode
+   *
+   * @param exception the Exception
+   * @return the TransportResult error
+   */
```
philipphofmann, line 340: `l`: It' really nice that we have all methods commented, but if they are private and do simple things, I don't think they are needed, because the comment doesn’t add any value IMO. [Detekt](https://arturbosch.github.io/detekt/comments.html#commentoverprivatefunction) has even a rule against this.

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
```diff
@@ -86,7 +86,7 @@ class HttpTransportTest {
     fun `uses Retry-After header if X-Sentry-Rate-Limit is not set when sending an event`() {
         val transport = fixture.getSUT()
 
-        whenever(fixture.connection.inputStream).thenThrow(IOException())
+        whenever(fixture.serializer.serialize(any<SentryEvent>(), any())).thenThrow(IOException())
```
philipphofmann, line 89: It's a good sign for your tests when you change so much in the actual class and so little in your tests 👍

philipphofmann: `l`: We need to change the exact same code in roughly 10 places in this test. That is a sign for me, that we should extract this functionality into a function for example.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: it was `gzip`, but @rhcarvalho preferred to not mention at all.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: @philipphofmann not really, because we use `try-with-resources` and to be able to use it, we can't use a method that returns all the types and are free'd afterwards

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: yes, makes sense

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: 1st comment: this method is called in a `finally` block, but we have 2 calls, send(envelope) and send(event), so extracted to a method, DRY.

marandaneto: 2nd comment, I generally agree, but indeed is out of this scope, worth mentioning that keeping a connection open for some time also has downsides.
this will be worked on when sentry-java lands to this repo.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: this would make sense for closed source, but I'm highly against for open-source.
comments don't add an extra size to the final bytecode and it helps a lot for contributors.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
philipphofmann: If the method returns the `writer` the `try-with-resources` is calling `close()` on the `writer` and the `writer` calls close on the `gzip` and `gzip` calls close on the `outputStream`. I think this should work fine.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
philipphofmann: > 1st comment: this method is called in a `finally` block, but we have 2 calls, send(envelope) and send(event), so extracted to a method, DRY.

Yep, but how do we know that everybody is calling this in a `finally` block in the future? Adding it makes it harder to misuse the method. That's [poka yoke](https://en.wikipedia.org/wiki/Poka-yoke).

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
philipphofmann: I disagree, but that's fine, because I don't have a strict opinion on this. Maybe we can discuss this once f2f. Not worth discussing it right now. 😄

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
marandaneto: I extracted to a method, I do understand the duplication is not good, but it helps to read and understand the test when it's all there in the same block of code, otherwise to understand a test you need to navigate thru all the methods, but indeed, in this case, it's worth it.
I tried to use `inline/reified` but mockito is complaining, I'll check this out later.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: it doesn't, they only guarantee to be closed if they are inside of the try-with-resources.
if creating `BufferedWriter` throws, only the `Writer` will be freed and the previous `OutputStream` and `GZIPOutputStream` won't

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: @philipphofmann it's a private method that closes the response stream and disconnect the connection, calling it on `finally` is just "a good approach", if you prefer to call it in all the catch blocks, it'll behave exactly the same, but then you repeat even more code.
So I don't get your point here, do you prefer to duplicate this code in both finally blocks?

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: sure, let's discuss offline.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
philipphofmann: We could wrap the creation of the `BufferedWriter` in the extracted method in a try catch and then we would be good. Nevermind it is not so important to me. 
What is important to me is that `public @NotNull TransportResult send(final @NotNull SentryEnvelope envelope) throws IOException` and `public @NotNull TransportResult send(final @NotNull SentryEvent event) throws IOException` are almost identical. The only difference is the log message and that sending the envelope has an extra `catch (Exception e)`.

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
philipphofmann: That would be my suggestion.
```suggestion
    finally {
      connection.disconnect();
    }
```
As already said it doesn't make a difference right now. We can keep it like it this. `YAGNI`

marandaneto: @philipphofmann ok got it, I thought you were mentioning the other 2 finally blocks of send(event/envelope)

marandaneto: yep it does make sense, will do it, thx

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
marandaneto: @philipphofmann the extra `catch` is due to the nature of serializing an envelope which is not only GSON's job but also us parsing it correctly, it may throw Exception due to invalid/missing data or something like.
if you don't use `try-with-resources` and wrap to normal try/catch altogether, you'll have way more code to close and end resources altogether, it was like this before, and try with resources is making sure that none of them leak even if it throws, the downside is a bit of duplicated code, but is probably gonna be worked on in the future as the `/store` endpoint will be deprecated and one of this method is gonna be removed anyway, so I'd not focus on refactoring this if that's ok.

philipphofmann: It is fine for me to keep it like it is.

sentry-core/src/test/java/io/sentry/core/transport/HttpTransportTest.kt
philipphofmann: If you name the method properly you shouldn't need to read it's content to understand what it is doing.

PR #437, opened 2020-05-29
fix: update last session timestamp

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: update last session timestamp

when app goes to foreground and get lastUpdatedSession, it should do the same on when app goes to background, so app won't stop and start sessions every time after the interval period.

running it

## :pencil: Checklist

- [X] All tests passing


bruno-garcia:
Good catch!

PR #445, opened 2020-06-03
fix: do not lookup sentry-debug-meta but instead load it directly

### LINKED ISSUE #442: App takes much longer to load when App is built with sentry sdk
_Platform:_
- [ ]  Android -> On All Android devices we have. e.g. Galaxy S6 Edge (version 6.0.1), Galaxy Tab S3(version 7.0)
- [ ]  Java -> JavaVersion.VERSION_1_8
- [ ]  NDK -> android-ndk-r20(Did not try with other NDKs), cmake version 3.12.3

_IDE:_
- [ ]  Android Studio -> 3.5 (issue is seen with cmd gradle build as well)

_Build system:_
- [ ]  Gradle -> 5.5

_Android Gradle Plugin:_
- [ ]  Yes -> If yes, which version?
- [ ]  No

_Sentry Android Gradle Plugin:_
- [ ]  Yes -> If yes, which version?
- [ ]  No

_Proguard/R8:_    both cases
- [ X]  Enabled
- [ X]  Disabled

_sentry-android installed with:_
- [ X] JCenter
- [ ] Bintray
- [ ] Maven Central
- [ ] Manually

The version of sentry-android:
2.1.2

---
I have the following issue:
When android app is built with sentry sdk it takes around 10 seconds to load the app. 

**from Logcat :**
1. Without sentry sdk => ActivityManager: **Displayed** com.machinezone.gow/com.mz.jix.MainActivity: **+492ms**
2. with Sentry SDK => I/ActivityManager: Displayed com.machinezone.gow/com.mz.jix.MainActivity: **+9s696ms**

**Displayed** => This value represents the amount of time elapsed between launching the process and finishing drawing the corresponding activity on the screen

The issue is specially worse with Asset build on Android.

**Steps to reproduce:**
Just do an Asset build on android with sentry SDK and launch the app. After that observe approx 10 seconds of black screen before Application finally starts. Sentry SDK inclusion in built is causing significant app startup delay

**Actual result:**
App startup is significantly delayed.

**Expected result:**
App startup time should be same as app without Sentry SDK

marandaneto:

- [X] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: do not lookup sentry-debug-meta but instead loads it directly

probably fixes #442
one may have thousands of asset files and this will slow down the startup time.
instead of looking up if the sentry-debug-meta exists, we try to load it directly and take care of `FileNotFoundException`.
the downside is: if one doesn't have asset files, catching the exception might be more expensive than the looking up itself, but worth it.

running it, tests are in place.

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
ideally `sentry-debug-meta.properties` will be a manifest config or SentryOptions, as a file this is suboptimal, but needed because of retro compatibility.

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -37,13 +37,14 @@ import io.sentry.core.protocol.SentryThread;
 import io.sentry.core.protocol.User;
 import io.sentry.core.util.ApplyScopeUtils;
 import io.sentry.core.util.Objects;
+import java.io.BufferedInputStream;
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.HashMap;
@@ -880,34 +881,31 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
   }
 
   private @Nullable String[] getProGuardUuids() {
-    AssetManager assets = context.getAssets();
-    try {
-      String[] files = assets.list("");
+    final AssetManager assets = context.getAssets();
 
-      List<String> listFiles = Arrays.asList(files != null ? files : new String[0]);
-      if (listFiles.contains("sentry-debug-meta.properties")) {
-        try (InputStream is = assets.open("sentry-debug-meta.properties")) {
-          Properties properties = new Properties();
-          properties.load(is);
+    try (final InputStream is =
+        new BufferedInputStream(assets.open("sentry-debug-meta.properties"))) {
+      final Properties properties = new Properties();
+      properties.load(is);
 
-          String uuid = properties.getProperty("io.sentry.ProguardUuids");
-          if (uuid != null && !uuid.isEmpty()) {
-            return uuid.split("\\|", -1);
-          }
-          options
-              .getLogger()
-              .log(SentryLevel.INFO, "io.sentry.ProguardUuids property was not found.");
-        } catch (IOException e) {
-          options.getLogger().log(SentryLevel.ERROR, "Error getting Proguard UUIDs.", e);
-        }
-        options
-            .getLogger()
-            .log(SentryLevel.INFO, "io.sentry.ProguardUuids property was not found.");
-      } else {
-        options.getLogger().log(SentryLevel.INFO, "Proguard UUIDs file not found.");
+      final String uuid = properties.getProperty("io.sentry.ProguardUuids");
+      if (uuid != null && !uuid.isEmpty()) {
+        String[] proguardUuids = uuid.split("\\|", -1);
+        options.getLogger().log(SentryLevel.DEBUG, "ProguardUuids found: %s", proguardUuids);
+        return proguardUuids;
       }
+      options
+          .getLogger()
+          .log(
+              SentryLevel.INFO, "io.sentry.ProguardUuids property was not found or it is invalid.");
+    } catch (FileNotFoundException e) {
+      options.getLogger().log(SentryLevel.INFO, "sentry-debug-meta.properties file was not found.");
     } catch (IOException e) {
-      options.getLogger().log(SentryLevel.ERROR, "Error listing Proguard files.", e);
+      options.getLogger().log(SentryLevel.ERROR, "Error getting Proguard UUIDs.", e);
+    } catch (RuntimeException e) {
+      options
+          .getLogger()
+          .log(SentryLevel.ERROR, "sentry-debug-meta.properties file is malformed.", e);
     }
 
     return null;
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -24,6 +24,7 @@ import io.sentry.android.core.util.DeviceOrientations;
 import io.sentry.android.core.util.RootChecker;
 import io.sentry.core.DateUtils;
 import io.sentry.core.EventProcessor;
+import io.sentry.core.ILogger;
 import io.sentry.core.SentryEvent;
 import io.sentry.core.SentryLevel;
 import io.sentry.core.SentryOptions;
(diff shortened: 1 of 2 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -880,34 +881,35 @@ private void setAppPackageInfo(final @NotNull App app, final @NotNull PackageInf
...
+    try (final InputStream is =
+        new BufferedInputStream(assets.open("sentry-debug-meta.properties"))) {
+      final Properties properties = new Properties();
+      properties.load(is);
+
+      final String uuid = properties.getProperty("io.sentry.ProguardUuids");
+      if (uuid != null && !uuid.isEmpty()) {
+        final String[] proguardUuids = uuid.split("\\|", -1);
+
+        // it should be only 1 proguard uuid, but the API accepts an Array so keeping it consistency
```
philipphofmann, line 895: `l`: I think we have a few typos here
```suggestion
        // it should be only 1 proguard uuid, but the API accepts an array so we are keeping it for consistency
```

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -880,34 +881,35 @@ private void setAppPackageInfo(final @NotNull App app, final @NotNull PackageInf
...
+    final AssetManager assets = context.getAssets();
+
+    try (final InputStream is =
+        new BufferedInputStream(assets.open("sentry-debug-meta.properties"))) {
+      final Properties properties = new Properties();
+      properties.load(is);
+
+      final String uuid = properties.getProperty("io.sentry.ProguardUuids");
+      if (uuid != null && !uuid.isEmpty()) {
+        final String[] proguardUuids = uuid.split("\\|", -1);
```
philipphofmann, line 893: `l`: I would rather rename this to
```suggestion
        final String[] proguardUUIDs = uuid.split("\\|", -1);
```
because UUID is an abbreviation.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -880,34 +881,35 @@ private void setAppPackageInfo(final @NotNull App app, final @NotNull PackageInf
...
+
+        // it should be only 1 proguard uuid, but the API accepts an Array so keeping it consistency
+        for (final String item : proguardUuids) {
+          options.getLogger().log(SentryLevel.DEBUG, "ProguardUuids found: %s", item);
```
philipphofmann, line 897: `l`: I think it is only one UUID.
```suggestion
          options.getLogger().log(SentryLevel.DEBUG, "ProguardUuid found: %s", item);
```

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -880,34 +881,35 @@ private void setAppPackageInfo(final @NotNull App app, final @NotNull PackageInf
...
       }
+      options
+          .getLogger()
+          .log(
+              SentryLevel.INFO, "io.sentry.ProguardUuids property was not found or it is invalid.");
+    } catch (FileNotFoundException e) {
+      options.getLogger().log(SentryLevel.INFO, "sentry-debug-meta.properties file was not found.");
     } catch (IOException e) {
-      options.getLogger().log(SentryLevel.ERROR, "Error listing Proguard files.", e);
+      options.getLogger().log(SentryLevel.ERROR, "Error getting Proguard UUIDs.", e);
```
philipphofmann, line 908: `l`: We have `ProguardUuids` most of the time. So I would change this to the following to keep it consistent.
```suggestion
      options.getLogger().log(SentryLevel.ERROR, "Error getting ProguardUuids.", e);
```

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -880,34 +881,35 @@ private void setAppPackageInfo(final @NotNull App app, final @NotNull PackageInf
...
+    } catch (RuntimeException e) {
+      options
```
philipphofmann, line 910: `l`: We call `options.getLogger()` many many times in this file. Does it make sense to create a property for it?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -880,34 +881,35 @@ private void setAppPackageInfo(final @NotNull App app, final @NotNull PackageInf
...
-            return uuid.split("\\|", -1);
-          }
-          options
-              .getLogger()
-              .log(SentryLevel.INFO, "io.sentry.ProguardUuids property was not found.");
-        } catch (IOException e) {
-          options.getLogger().log(SentryLevel.ERROR, "Error getting Proguard UUIDs.", e);
+    final AssetManager assets = context.getAssets();
+
+    try (final InputStream is =
```
philipphofmann, line 886: `h`: Can we please add the information of motivation and context here as a comment. I think this is very important to mention here why we don't check if the asset exists.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -880,34 +881,35 @@ private void setAppPackageInfo(final @NotNull App app, final @NotNull PackageInf
...
+        new BufferedInputStream(assets.open("sentry-debug-meta.properties"))) {
```
philipphofmann, line 887: `h`: Shouldn't we also call `close` on the assets or put `final AssetManager assets = context.getAssets();` in the try-with-resources block here?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: I've used either Proguard UUIDs or Proguard UUID in the logs

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: I've needed to revert this, `AssetManager` is a single instance per app. and when it gets closed, it's for good.
basically all the context.getResources(...) depends on `AssetManager` and if it's once closed, it's not possible to reuse anymore, this caused crashes when trying to use `Resources` :( TIL.

PR #451, opened 2020-06-05
run sample app and do not crash

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing

### DIFF
--- a/.github/workflows/emulator-sample-install.yml
+++ b/.github/workflows/emulator-sample-install.yml
@@ -1,0 +1,45 @@
+name: Workflow Sample Install macOS
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
+
+env:
+  CMAKE_VERSION: "3.10.2.4988404"
+
+jobs:
+  build:
+    runs-on: macOS-10.15
+
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v2
+
+      # Zulu Community distribution of OpenJDK
+      - name: set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: set up sdkmanager and emulators
+        uses: malinskiy/action-android/install-sdk@release/0.0.7
+      - run: adb devices
+      - run: echo $ANDROID_HOME
+      - run: sdkmanager --install "cmake;${CMAKE_VERSION}"
+
+      - uses: malinskiy/action-android/emulator-run-cmd@release/0.0.7
+        with:
+          cmd: ./gradlew installDebug
+          api: 25
+          tag: default
+          abi: x86
+
+      - name: Save logcat output
+        uses: actions/upload-artifact@master
+        if: failure()
+        with:
+          name: logcat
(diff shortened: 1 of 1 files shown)


.github/workflows/emulator-sample-install.yml
```diff
@@ -0,0 +1,48 @@
...
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
+
+env:
+  CMAKE_VERSION: "3.10.2.4988404"
+  NDK_VERSION: "21.0.6113669"
```
bruno-garcia, line 12:
```suggestion
  NDK_VERSION: "21.2.6472646"
```

github actions has this installed

marandaneto: we have a fixed version on the build file
https://github.com/getsentry/sentry-android/blob/master/sentry-android-ndk/build.gradle.kts#L41

21.2.x is not available on LGTM GH App. yet.

.github/workflows/emulator-sample-install.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: set up sdkmanager and emulators
+        uses: malinskiy/action-android/install-sdk@release/0.0.7
+      - run: sdkmanager platform-tools
+      - run: sdkmanager --install "cmake;${CMAKE_VERSION}"
+      - run: sdkmanager --install "ndk;${NDK_VERSION}"
+
+      - uses: malinskiy/action-android/emulator-run-cmd@release/0.0.7
+        with:
+          cmd: make runConnectedTests
+          api: 25
```
marandaneto, line 52: ideally making a matrix here

sentry-sample/src/androidTest/java/io/sentry/sample/MainActivityTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        onView(withId(R.id.send_message)).perform(click())
+        onView(withId(R.id.capture_exception)).perform(click())
+        onView(withId(R.id.breadcrumb)).perform(click())
+        onView(withId(R.id.native_capture)).perform(click())
+
+//        val retrofit = Retrofit.Builder()
+//            .baseUrl("https://sentry.io/")
+//            .client(OkHttpClient.Builder().addInterceptor { chain ->
+//                // token from https://sentry.io/settings/account/api/auth-tokens/
+//                val request = chain.request().newBuilder().addHeader("Authorization", "Bearer ${TOKEN}}").build()
```
marandaneto, line 43: @bruno-garcia do you know if its possible to generate a token on demand? cus we can't read from a system env as this runs on the emulator.
also, it's not "nice" to leak the token on the code, Reverse engineering is too easy at this point.

bruno-garcia: How would we get the generated token?
I agree it's not ideal to embed the key in the apk. We should be able to provide an apk alone and test it out since the testing code is not part of the apk

PR #456, opened 2020-06-12
fix: timestamp with millis from react-native should be in UTC format

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: timestamp with millis from react-native should be already in UTC format

timestamps were off in react-native events if not UTC.

unit test.

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
release a new sentry-react-native version.

### DIFF
--- a/sentry-android-core/src/test/java/io/sentry/android/core/AndroidSerializerTest.kt
+++ b/sentry-android-core/src/test/java/io/sentry/android/core/AndroidSerializerTest.kt
@@ -122,6 +122,17 @@ class AndroidSerializerTest {
     }
 
     @Test
+    fun `when deserializing mills timestamp with mills precision, it shoule be UTC`() {
+        // Jun 7, 2020 12:38:12 PM UTC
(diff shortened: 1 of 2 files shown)


sentry-android-core/src/test/java/io/sentry/android/core/AndroidSerializerTest.kt
```diff
@@ -121,6 +121,17 @@ class AndroidSerializerTest {
         assertEquals(expected, actual!!.timestamp)
     }
 
+    @Test
+    fun `when deserializing mills timestamp with mills precision, it shoule be UTC`() {
```
bruno-garcia, line 125:
```suggestion
    fun `when deserializing mills timestamp with mills precision, it should be UTC`() {
```

marandaneto: gud catch

PR #458, opened 2020-06-14
wip

bruno-garcia:


bruno-garcia:
Discussed offline and this isnt' needed since we always build the android libraries which would break if sentry-core used API incompatible with the minApiLevel /cc @marandaneto

PR #461, opened 2020-06-16
fix: init native libs if available on sdk init

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: init native libs if available on sdk init

one could init the SDK with an empty DSN and still load a native lib that depends on the libsentry, the libsentry was never initialized, now it is and it'll be NoOp as the Android SDK.

running on API 16 and 28, with and without an empty DSN.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -1,8 +1,11 @@
 package io.sentry.android.core;
 
+import static io.sentry.android.core.NdkIntegration.SENTRY_NDK_CLASS_NAME;
+
 import android.app.Application;
 import android.content.Context;
 import android.content.pm.PackageInfo;
+import android.os.Build;
 import io.sentry.core.EnvelopeReader;
 import io.sentry.core.IEnvelopeReader;
 import io.sentry.core.ILogger;
@@ -47,6 +50,22 @@ final class AndroidOptionsInitializer {
       final @NotNull SentryAndroidOptions options,
       @NotNull Context context,
       final @NotNull ILogger logger) {
+    init(options, context, logger, new BuildInfoProvider());
+  }
+
+  /**
+   * Init method of the Android Options initializer
+   *
+   * @param options the SentryOptions
+   * @param context the Application context
+   * @param logger the ILogger interface
+   * @param buildInfoProvider the IBuildInfoProvider interface
+   */
+  static void init(
+      final @NotNull SentryAndroidOptions options,
+      @NotNull Context context,
+      final @NotNull ILogger logger,
+      final @NotNull IBuildInfoProvider buildInfoProvider) {
     Objects.requireNonNull(context, "The context is required.");
 
     // it returns null if ContextImpl, so let's check for nullability
@@ -67,12 +86,12 @@ final class AndroidOptionsInitializer {
 
     final IEnvelopeReader envelopeReader = new EnvelopeReader();
 
-    installDefaultIntegrations(context, options, envelopeReader);
+    installDefaultIntegrations(context, options, envelopeReader, buildInfoProvider);
 
     readDefaultOptionValues(options, context);
 
     options.addEventProcessor(
-        new DefaultAndroidEventProcessor(context, options, new BuildInfoProvider()));
+        new DefaultAndroidEventProcessor(context, options, buildInfoProvider));
 
     options.setSerializer(new AndroidSerializer(options.getLogger(), envelopeReader));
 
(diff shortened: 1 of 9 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -82,11 +101,14 @@ static void init(
...
       final @NotNull Context context,
       final @NotNull SentryOptions options,
-      final @NotNull IEnvelopeReader envelopeReader) {
+      final @NotNull IEnvelopeReader envelopeReader,
+      final @NotNull IBuildInfoProvider buildInfoProvider) {
 
     // Integrations are registered in the same order. NDK before adding Watch outbox,
     // because sentry-native move files around and we don't want to watch that.
-    options.addIntegration(new NdkIntegration());
+    if (loadNdkIfAvailable(options, buildInfoProvider)) {
```
marandaneto, line 109: @bruno-garcia maybe I'd be good to get the `Class<?>` as a return type and pass to `NdkIntegration` as a ctor param, so we do the reflection only once on the class level, wdyt?

sentry-android-core/src/test/java/io/sentry/android/core/AndroidOptionsInitializerTest.kt
```diff
@@ -181,13 +188,47 @@ class AndroidOptionsInitializerTest {
...
+        }
+        val mockContext = ContextUtilsTest.mockMetaData(metaData = bundle)
+        val logger = mock<ILogger>()
+        val sentryOptions = SentryAndroidOptions().apply {
+            isDebug = true
+        }
+        val buildInfo = mock<IBuildInfoProvider>()
+        whenever(buildInfo.sdkInfoVersion).thenReturn(16)
+
+        // hard to test, lets just check that its not throwing anything
```
marandaneto, line 203: @bruno-garcia to be able to test this properly, I'd need to have an interface that hides the static call `Class.forName`, so I can mock and assert it, do you think it's worth it?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
bruno-garcia: That makes sense. So the return type would be `@Nullable Class<?>`.
It would also improve things imho if we didn't fiddle around with the options `setEnableNdk` value. Just leave it as-is.
When we check if it's true, if we did succeed in getting the class name through this method, we enable it. Otherwise we don't. It would just be nice of this method didn't have side effects. Simply get the class name if it's possible.

sentry-android-core/src/test/java/io/sentry/android/core/AndroidOptionsInitializerTest.kt
bruno-garcia: With your suggested refactor of having `loadNdk..` return the class, we could test this by making sure `NdkIntegration` initialized with `null` for `Class<?>` and with an instance it does?

I'd say it's your call. It's nice to have it testable and be able to assert this properly but on the other hand it bloats the SDK further with 1 more type just for this 1 test case.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
marandaneto: yes, but on the NdkIntegration we do need to set setEnableNdk=false because we rely on that to add or not the sentry-android-ndk as a package on the sdk info

sentry-android-core/src/test/java/io/sentry/android/core/AndroidOptionsInitializerTest.kt
marandaneto: yes this is fully testable/tested now

marandaneto: I've added a new type for the sake of testability, this is a critical part of the sdk, as it bundles other sdks and it'd be good to be sure that this is working properly, although those types are package-private anyway.

sentry-android-core/src/main/java/io/sentry/android/core/NdkIntegration.java
```diff
@@ -46,4 +52,10 @@ public final void register(IHub hub, SentryOptions options) {
       options.setEnableNdk(false);
     }
```
bruno-garcia, line 53: L:
```suggestion
    }
    else {
        options.getLogger().log(DEBUG, "NDK won't be initialized"):
    }
```

marandaneto: this is already implicit by:

`options.getLogger().log(SentryLevel.DEBUG, "NdkIntegration enabled: %s", enabled);`

PR #462, opened 2020-06-16
Make JVM target explicit in sentry-core

dilbernd:
IDEA’s gradle support failed to inherit language/bytecode level from
parent project, causing mockito failures w/ 11 bytecode.

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Explicitly set Java / JVM versions for the `sentry-core` subproject.

Technically it doesn’t fix a Sentry bug but rather papers over a bug in IDEA 2020.1.2 which makes tests fail from the IDE: It doesn’t detect the source & target JVM versions from the parent project. Then it generates JVM 11 bytecode, which our Mockito can’t handle.

Reimported in IDEA 2020.1.2 and saw that the JVM versions were now correct in the settings & the tests were green.

Had some tests that did not seem to deliver any result (green or red) in AS 4.0 but this did not change after reverting this change locally so it’s at least unrelated.

## :pencil: Checklist

- [ ] I added tests to verify changes - :warning: Build config only
- [X] All tests passing

### DIFF
--- a/sentry-core/build.gradle.kts
+++ b/sentry-core/build.gradle.kts
@@ -9,6 +9,15 @@ plugins {
     id(Config.QualityPlugins.gradleVersions)
 }
 
+configure<JavaPluginConvention> {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+    kotlinOptions.jvmTarget = "1.8"
+}
+
 dependencies {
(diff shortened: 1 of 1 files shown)


dilbernd:
Forgot to note under “how tested”: Local `./gradlew clean test` on OpenJDK Runtime Environment (Zulu 8.46.0.19-CA-macosx) (build 1.8.0_252-b14) also ran through.

marandaneto:
ideally, we'd set this on `allprojects` but I remember having issues with Kotlin DSL, it works just fine using Groovy though, let's bring this in and I'll check if they already fixed the compatibility with Groovy Clojures and Kotlin DSL.
thanks for the fix @dilbernd

PR #464, opened 2020-06-18
feat: timber integration

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
feat: timber integration

one may raise events from Timber

running and unit tests

## :pencil: Checklist

- [X] All tests passing

## :crystal_ball: Next steps
revert manifest and self init, but we need to define the sample somewhere


sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * Captures a Sentry Event if the min. level is equal or higher than the min. required level.
+     */
+    override fun log(priority: Int, tag: String?, message: String, throwable: Throwable?) {
+        val sentryEvent = SentryEvent()
+        sentryEvent.level = getSentryLevel(priority)
+
+        // if there's no throwable, should it be a breadcrumb then?
```
marandaneto, line 33: @bruno-garcia wdyt?

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
```diff
@@ -0,0 +1,74 @@
...
+        throwable?.let {
+            sentryEvent.setThrowable(it)
+        }
+
+        // timber concatenates the stacktrace to the exception if the exception is not null
+        // let's take only the 1st line as this is probably the only message written by the user.
+        // or the 1st line of the exception which is probably the same as event.title
+        // eg java.lang.Exception: java.lang.Exception: java.lang.Exception: Some exception.
+        // timber should offer a way of disabling it
```
marandaneto, line 42: https://github.com/JakeWharton/timber/pull/393

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
bruno-garcia: In other SDKs there were 2 log levels you could configure the library with:


Both checks run always, so potentially a Log call will result in both Event and Breadcrumb: 
https://github.com/getsentry/sentry-dotnet/blob/17c5dabac328181c6f700f5afb07d301cc0183d6/src/Sentry.Extensions.Logging/SentryLogger.cs#L57-L99

One to configure min level for event:
https://github.com/getsentry/sentry-dotnet/blob/17c5dabac328181c6f700f5afb07d301cc0183d6/src/Sentry.Extensions.Logging/SentryLogger.cs#L125

One to configure min level for breadcrumb: https://github.com/getsentry/sentry-dotnet/blob/17c5dabac328181c6f700f5afb07d301cc0183d6/src/Sentry.Extensions.Logging/SentryLogger.cs#L144

You can check/send event first, and run the check to add breadcrumb later. The advantage is that a second event will include the previous error in its breadcrumb list.

marandaneto: good idea, just added it, thx

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
```diff
@@ -0,0 +1,74 @@
...
+        val realMessage = message.splitToSequence(delimiters = *arrayOf("\n"))
```
philipphofmann, line 61: `h`: I think if the user calls `Timber.e("message\nhello\nworld")` this code leads to cutting off the `hello\nworld`. I don't know if it is a good idea to implement this feature here like this. I would just take the message as it is. When Timber for some reason changes its implementation we also need to adapt this functionality here.

sentry-android-timber/src/test/java/io/sentry/android/timber/SentryTimberTreeTest.kt
```diff
@@ -0,0 +1,185 @@
...
+        Timber.e("message")
+        verify(fixture.hub).captureEvent(check {
+            assertNull(it.getTag("TimberTag"))
+        })
+    }
+
+    @Test
+    fun `Tree captures an event with given message`() {
+        val sut = fixture.getSut()
+        sut.e("message")
```
philipphofmann, line 163: `m`: I would also add a test with a message containing `\n`

sentry-android-timber/src/test/java/io/sentry/android/timber/SentryTimberTreeTest.kt
```diff
@@ -0,0 +1,185 @@
...
+import io.sentry.core.IHub
+import io.sentry.core.SentryLevel
+import io.sentry.core.getExc
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+import timber.log.Timber
+
+class SentryTimberTreeTest {
```
philipphofmann, line 18: Really nice tests 👏

sentry-android-timber/build.gradle.kts
```diff
@@ -0,0 +1,111 @@
...
+    sign = Config.Deploy.sign
+    artifactId = project.name
+    uploadName = "${project.group}:${project.name}"
+    devId = Config.Sentry.userOrg
+    devName = Config.Sentry.devName
+    devEmail = Config.Sentry.devEmail
+    scmConnection = Config.Sentry.scmConnection
+    scmDevConnection = Config.Sentry.scmDevConnection
+    scmUrl  = Config.Sentry.scmUrl
+}
```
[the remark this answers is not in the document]
marandaneto: this doesn't work well for kotlin dsl and we already have a task for this on asana, so it won't be worked on this PR.
ideally, this would be done with Gradle plugins under buildSrc

sentry-android-timber/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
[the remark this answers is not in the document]
marandaneto: we already have on the top level, but this has been created automatically, will delete it

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberIntegration.kt
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    override fun close() {
+        if (this::tree.isInitialized) {
+            Timber.uproot(tree)
+
+            if (this::logger.isInitialized) {
+                logger.log(SentryLevel.DEBUG, "SentryTimberIntegration removed.")
+            }
+        }
```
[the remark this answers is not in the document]
marandaneto: this would print even if the `tree` was not initialized, so IMO this is wrong.

sentry-android-timber/src/test/java/io/sentry/android/timber/SentryTimberIntegrationTest.kt
```diff
@@ -0,0 +1,55 @@
...
+    fun beforeTest() {
+        Timber.uprootAll()
+    }
+
+    @Test
+    fun `Integrations plants a tree into Timber on register`() {
+        val sut = fixture.getSut()
+        sut.register(fixture.hub, fixture.options)
+
+        assertEquals(1, Timber.treeCount())
```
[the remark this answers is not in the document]
marandaneto: `SentryTimberTree` already asserts that

sentry-android-timber/src/test/java/io/sentry/android/timber/SentryTimberIntegrationTest.kt
```diff
@@ -0,0 +1,55 @@
...
+import com.nhaarman.mockitokotlin2.mock
+import io.sentry.core.IHub
+import io.sentry.core.SentryLevel
+import io.sentry.core.SentryOptions
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import timber.log.Timber
+
+class SentryTimberIntegrationTest {
```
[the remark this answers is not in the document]
marandaneto: mm it's possible, but we don't do this for any other integration I guess, as what really matters is the `Hub` to be called and this is already tested -> `SentryTimberTreeTest`
I'll check that though

sentry-sample/src/main/java/io/sentry/sample/MainActivity.java
```diff
@@ -16,19 +16,17 @@ protected void onCreate(Bundle savedInstanceState) {
...
-    Timber.i("Sentry.isEnabled() = %s", Sentry.isEnabled());
+    Timber.d("Sentry.isEnabled() = %s", Sentry.isEnabled());
 
     binding.crashFromJava.setOnClickListener(
         view -> {
           throw new RuntimeException("Uncaught Exception from Java.");
         });
 
-    binding.sendMessage.setOnClickListener(view -> Sentry.captureMessage("Some message."));
+    binding.sendMessage.setOnClickListener(view -> Timber.i("Some message."));
```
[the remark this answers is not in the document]
marandaneto: See the next steps on the description, this will be reverted.
This is just for testing, the examples would be under getsentry/examples

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
```diff
@@ -65,6 +65,7 @@ public Date getTimestamp() {
     return (Date) timestamp.clone();
   }
 
+  @Nullable
```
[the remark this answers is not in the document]
marandaneto: for better interoperability with Kotlin.
I add those annotations on every field/method I touch as we didn't do for all of them at the beginning

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
marandaneto: I do agree, but it's the only workaround I could find to not log the whole exception stack trace, when Proguard is enabled, it's just useless and I could not find a delimiter that separates a real message to the exception

sentry-android-timber/build.gradle.kts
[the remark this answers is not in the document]
philipphofmann: If you are aware of the issue, I'm fine with it.

sentry-android-timber/src/test/java/io/sentry/android/timber/SentryTimberTreeTest.kt
marandaneto: the test would pretty much fail because of https://github.com/getsentry/sentry-android/pull/464#discussion_r443786703

so we'd need to talk if we want to go with that approach or not.

marandaneto: I mean, it'd log only the 1st line

sentry-android-timber/src/test/java/io/sentry/android/timber/SentryTimberIntegrationTest.kt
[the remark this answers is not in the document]
philipphofmann: Yes, but this assert only checks if there is 1 tree. It could be any tree. 
The current tests don't fail 
* if we change `SentryTimberIntegration.tree` to `Timber.DebugTree.
* if we change the init of the tree to `tree = SentryTimberTree(hub, SentryLevel.ERROR)`
* if we initialize our own hub in `register`

We can keep this test, but I would say we could add one or two to make sure the `SentryTimberTree` is properly planted into Timber and we can do this for example by checking that `Timber.d(...)` calls the `hub`. We could also add one test to make sure the `minLevel` is properly passed down to `SentryTimberTree`.

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
philipphofmann: As already pointed out I would just take the message from Timber and not cut it off. This is how Timber is implemented and if we keep the cutting of users are going to lose their information after the first `\n`.

sentry-android-timber/src/test/java/io/sentry/android/timber/SentryTimberIntegrationTest.kt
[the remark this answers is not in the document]
marandaneto: I can't `verify` which tree is planted as `Timber.plant` is a static call or I'd need an adapter for that, it's not worth IMO.
Adding a test for checking the minLevels and calling hub to be sure it planted "any" tree that calls our Hub.
also did a `instanceof` test, guess it's better now, thanks

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
marandaneto: @philipphofmann this will have the side effect of people thinking that is us not demangling the stack traces properly.
I'd like a 3rd vote here, both options are suboptimal.

HazAT: I would also go with not splitting this and just use the message as it is.

Reasoning: We rather send too much than too little. This could have consequences that things a user wants to have for sure are not in the message.

sentry-android-timber/src/test/java/io/sentry/android/timber/SentryTimberTreeTest.kt
marandaneto: no need for that now as we just use the raw given message by timber

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
```diff
@@ -0,0 +1,102 @@
...
+
+        // if there's no throwable, should it be a breadcrumb then?
+        throwable?.let {
+            sentryEvent.setThrowable(it)
+        }
+        sentryEvent.message = Message().apply {
+            formatted = message
+        }
+
+        tag?.let {
```
bruno-garcia, line 61: L: just realize we could use the event.logger field here instead of a custom tag

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
```diff
@@ -0,0 +1,102 @@
...
+        // we need to define a merging mechanism then
+        sentryEvent.setTag("origin", "SentryTimberIntegration")
+
+        hub.captureEvent(sentryEvent)
+    }
+
+    /**
+     * Adds a breadcrumb if the event has an exception.
+     */
+    private fun addBreadcrumbIfEventHasThrowable(level: SentryLevel, throwable: Throwable?, message: String) {
```
bruno-garcia, line 74: H: should this just be add Breadcrumb? We should add it regardless of having throwable.

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Sentry Timber tree which is responsible to capture events via Timber
+ */
+class SentryTimberTree(private val hub: IHub, private val minEventLevel: SentryLevel,
+                       private val minBreadcrumbLevel: SentryLevel) : Timber.Tree() {
+
+    /**
+     * do not log if it's lower than min. required level.
+     */
+    private fun isLoggable(level: SentryLevel, minLevel: SentryLevel): Boolean = level.ordinal >= minLevel.ordinal
```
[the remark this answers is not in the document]
marandaneto: this is called 2 times, also in addBreadcrumbIfEventHasThrowable

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
marandaneto: well, this breadcrumb is `exception` category and `error` as a type, so it'd make sense to be only if there's an error.
if we log everything as a breadcrumb, I can't choose a category and type as they won't match most of the time, so I'm not really sure if I'd like to log literally everything as a breadcrumb.
If doing that, I'd keep type and category just not set at all, wdyt?

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
marandaneto: the way how it works on Timber is different. see:

> Set a one-time tag for use on the next logging call

So I'd rather keep this tag which is a custom tag set by the client, but also setting event.logger always to Timber.

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
bruno-garcia: Other logging integrations log everything, without any specific category that I can recall. It just goes blank (it's not required). The `level` is more important.

It's very useful to see the last 100 log entries (i.e Info or higher) right before an event.

PR #474, opened 2020-06-30
fix: clone session before sending to the transport

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: clone session before sending to the transport

if a session was mutated before sent and discarded, fields could be wrong as they are mutable.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Session.java
+++ b/sentry-core/src/main/java/io/sentry/core/Session.java
@@ -96,12 +96,12 @@ public final class Session {
     this(
         State.Ok,
         DateUtils.getCurrentDateTime(),
-        null,
+        DateUtils.getCurrentDateTime(),
         0,
         distinctId,
         UUID.randomUUID(),
         true,
-        0L,
+        null,
         null,
         (user != null ? user.getIpAddress() : null),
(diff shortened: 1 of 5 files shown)


sentry-core/src/main/java/io/sentry/core/Session.java
```diff
@@ -96,12 +96,12 @@ public Session(
     this(
         State.Ok,
         DateUtils.getCurrentDateTime(),
-        null,
+        DateUtils.getCurrentDateTime(),
```
marandaneto, line 99: so session started also has a timestamp otherwise the server would assume current time

sentry-core/src/main/java/io/sentry/core/Session.java
```diff
@@ -96,12 +96,12 @@ public Session(
...
         0,
         distinctId,
         UUID.randomUUID(),
         true,
-        0L,
+        null,
```
marandaneto, line 104: no need to send sequence 0 if init=true, server already assumes sequence=0

Issue #476, opened 2020-07-02
coreLibraryDesugaring breaks Sentry


PaulWoitaschek:
Enabling coreLibraryDesugaring is incompatible with sentry 2.2.0

To reproduce this, add coreLibraryDesugaring to your application and now no events will be sent.

Here is an example project that demonstrates the issue: (it's the official sample)
https://github.com/PaulWoitaschek/sentry-android/commit/167cd616825e6ffaf7b1d57bfa81ceef1aa4a641

Just install it to your device in release mode:
`./gradlew sentry-sample:installRelease`

Click on `Capture exception`
And you'll see the stacktrace in your log:
```
2020-07-02 09:33:03.475 16774-16800/? E/Sentry: Event submission failed: 39ec87a940ff4d5fb47a138fc11c0764
    java.lang.ArrayIndexOutOfBoundsException: length=0; index=0
        at com.google.gson.internal.$Gson$Types.getGenericSupertype(Unknown Source:40)
        at com.google.gson.internal.$Gson$Types.getSupertype(Unknown Source:20)
        at com.google.gson.internal.$Gson$Types.getMapKeyAndValueTypes(Unknown Source:20)
        at com.google.gson.internal.bind.MapTypeAdapterFactory.create(Unknown Source:22)
        at com.google.gson.Gson.getDelegateAdapter(Unknown Source:35)
        at com.google.gson.internal.bind.TreeTypeAdapter.delegate(Unknown Source:11)
        at com.google.gson.internal.bind.TreeTypeAdapter.write(Unknown Source:4)
        at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.write(Unknown Source:34)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.write(Unknown Source:28)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.write(Unknown Source:42)
        at io.sentry.android.core.UnknownPropertiesTypeAdapterFactory$UnknownPropertiesTypeAdapter.write(Unknown Source:2)
        at io.sentry.android.core.UnknownPropertiesTypeAdapterFactory$UnknownPropertiesTypeAdapter.write(Unknown Source:2)
        at com.google.gson.Gson.toJson(Unknown Source:34)
        at com.google.gson.Gson.toJson(Unknown Source:8)
        at io.sentry.android.core.AndroidSerializer.serialize(Unknown Source:14)
[...]

PR #478, opened 2020-07-03
fix: contexts serializer avoids reflection and fixes desugaring issue

marandaneto:

- [X ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: contexts serializer avoids reflection and fixes desugaring issue

#476

running it, tests are already in place

## :pencil: Checklist

- [X] All tests passing


marandaneto:
> Good one. Maybe we need this for Extra too?

no, this only happens when it extends the thread-safe version of it.
because of https://developer.android.com/studio/write/java8-support?hl=fi#library-desugaring
they desugar ConcurrentHashMap specifically.

PR #483, opened 2020-07-08
feat: add sdkInfo to envelope header and event

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
feat: add sdkInfo to envelope header and event

so the server knows which SDK has sent that envelope

## :pencil: Checklist

- [X] All tests passing


sentry-core/src/main/java/io/sentry/core/SentryEnvelope.java
```diff
@@ -33,25 +34,47 @@ public SentryEnvelope(
...
     Objects.requireNonNull(item, "SentryEnvelopeItem is required.");
 
-    header = new SentryEnvelopeHeader(eventId);
+    header = new SentryEnvelopeHeader(eventId, sdkInfo);
     final List<SentryEnvelopeItem> items = new ArrayList<>(1);
     items.add(item);
     this.items = items;
   }
 
+  public SentryEnvelope(final @Nullable SentryId eventId, final @NotNull SentryEnvelopeItem item) {
```
bruno-garcia, line 61: M: Can we get rid of this overload to avoid creating envelopes without sdk-info?

sentry-core/src/main/java/io/sentry/core/SentryEnvelope.java
```diff
@@ -33,25 +34,47 @@ public SentryEnvelope(
...
+      final @Nullable SdkInfo sdkInfo)
+      throws IOException {
     Objects.requireNonNull(serializer, "Serializer is required.");
     Objects.requireNonNull(session, "session is required.");
 
-    return new SentryEnvelope(null, SentryEnvelopeItem.fromSession(serializer, session));
+    return new SentryEnvelope(null, sdkInfo, SentryEnvelopeItem.fromSession(serializer, session));
+  }
+
+  public static @NotNull SentryEnvelope fromSession(
```
bruno-garcia, line 76: M: This one would go too (no sdk-info)

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeaderAdapter.java
```diff
@@ -24,9 +25,44 @@ public void write(JsonWriter writer, SentryEnvelopeHeader value) throws IOExcept
...
+        writer.name("version_major").value(sdkInfo.getVersionMajor());
+      }
+      if (sdkInfo.getVersionMinor() != null) {
+        hasInitAttrs = initAttrs(writer, hasInitAttrs);
+        writer.name("version_minor").value(sdkInfo.getVersionMinor());
+      }
+      if (sdkInfo.getVersionPatchlevel() != null) {
+        hasInitAttrs = initAttrs(writer, hasInitAttrs);
+        writer.name("version_patchlevel").value(sdkInfo.getVersionPatchlevel());
+      }
```
bruno-garcia, line 47: What about we only add the version numbers if we have a name?

```suggestion
      if (sdkInfo.getSdkName() != null) {
        hasInitAttrs = initAttrs(writer, hasInitAttrs);
        writer.name("sdk_name").value(sdkInfo.getSdkName());
        if (sdkInfo.getVersionMajor() != null) {
          writer.name("version_major").value(sdkInfo.getVersionMajor());
        }
        if (sdkInfo.getVersionMinor() != null) {
          writer.name("version_minor").value(sdkInfo.getVersionMinor());
        }
        if (sdkInfo.getVersionPatchlevel() != null) {
          writer.name("version_patchlevel").value(sdkInfo.getVersionPatchlevel());
        }
      }
```

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeaderAdapter.java
```diff
@@ -35,20 +71,51 @@ public SentryEnvelopeHeader read(JsonReader reader) throws IOException {
...
+                  sdkInfo.setVersionMajor(reader.nextInt());
+                  break;
+                case "version_minor":
+                  sdkInfo.setVersionMinor(reader.nextInt());
+                  break;
+                case "version_patchlevel":
+                  sdkInfo.setVersionPatchlevel(reader.nextInt());
+                  break;
+                default:
+                  reader.skipValue();
```
bruno-garcia, line 102: M: Ideally we'd keep this in a bag (a Map<string, object> or something) to serialize back so it's future proof.
If sentry-native decides to add more data, we'd be dropping those when deserializing

sentry-core/src/main/java/io/sentry/core/cache/SessionCache.java
```diff
@@ -135,7 +135,11 @@ public void store(final @NotNull SentryEnvelope envelope, final @Nullable Object
             }
 
             session.end(timestamp);
-            final SentryEnvelope fromSession = SentryEnvelope.fromSession(serializer, session);
+            // if the App. has been upgraded and there's a new version of the SDK running, SdkInfo
```
bruno-garcia, line 138: L: What about we store the sdk-info with the session data?

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeaderAdapter.java
marandaneto: makes sense

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeaderAdapter.java
marandaneto: this already works for the SdkInfo object in an event, this is specifically for the envelope header, do we need more than the name and version? I mean it's fine to do it but it looks like not useful, wdyt?

sentry-core/src/main/java/io/sentry/core/cache/SessionCache.java
marandaneto: I'll create a new task for that and not totally related to this one as this is a very corner case.
I think we can tackle that along with the session refactoring+store endpoint task.

PR #484, opened 2020-07-09
feat: log request if response code is not 200

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
feat: log request if response code is not 200

we were not logging the error response on every possibility.

running it, unit tests already exist.

## :pencil: Checklist

- [X] All tests passing

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
+++ b/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
@@ -459,8 +459,7 @@ public class HttpTransport implements ITransport {
    * @param connection the HttpURLConnection
    * @return the error message or null if none
    */
-  private @NotNull String getErrorMessageFromStream(final @NotNull HttpURLConnection connection)
-      throws IOException {
+  private @NotNull String getErrorMessageFromStream(final @NotNull HttpURLConnection connection) {
     try (final InputStream errorStream = connection.getErrorStream();
         final BufferedReader reader =
             new BufferedReader(new InputStreamReader(errorStream, UTF_8))) {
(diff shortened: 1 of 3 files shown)


sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -256,6 +251,7 @@ public boolean isRetryAfter(final @NotNull String itemType) {
 
     connection.setRequestMethod("POST");
     connection.setDoOutput(true);
+    connection.setChunkedStreamingMode(0);
```
marandaneto, line 254: just a little optimization:
https://docs.oracle.com/javase/7/docs/api/java/net/HttpURLConnection.html#setChunkedStreamingMode(int)

sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -278,27 +274,48 @@ public boolean isRetryAfter(final @NotNull String itemType) {
...
-
-    int responseCode = -1;
+    TransportResult result;
 
     try (final OutputStream outputStream = connection.getOutputStream();
         final GZIPOutputStream gzip = new GZIPOutputStream(outputStream);
         final Writer writer = new BufferedWriter(new OutputStreamWriter(gzip, UTF_8))) {
       serializer.serialize(envelope, writer);
-
-      logger.log(DEBUG, "Envelope sent successfully.");
```
marandaneto, line 289: this should not have been here in the first place, as we are using `try-with-resources`, the resources are only available and freed to check its response code in the finally block, as the `try-with-resources` finishes every stream at the end, so the proper way to verify and log would be in the final block.

PR #488, opened 2020-07-20
fix: envelope header has sdkVersion and not sdkInfo

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: envelope header has sdkVersion and not sdkInfo

https://github.com/getsentry/develop/pull/89
docs were wrong about which object to use

## :pencil: Checklist

- [X] All tests passing


sentry-android-core/src/test/java/io/sentry/android/core/AndroidSerializerTest.kt
```diff
@@ -358,16 +358,21 @@ class AndroidSerializerTest {
...
-        assertEquals(2, sdkInfo.versionMinor)
-        assertEquals(3, sdkInfo.versionPatchlevel)
+        assertEquals("test", sdkInfo.name)
+        assertEquals("1.2.3", sdkInfo.version)
+
+        assertNotNull(sdkInfo.integrations)
+        assertEquals("NdkIntegration", sdkInfo.integrations!![0])
+
+        assertNotNull(sdkInfo.packages)
+        assertEquals("maven:sentry-android-core", sdkInfo.packages!![0].name)
```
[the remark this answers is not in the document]
philipphofmann: `l`: What about checking if `sdkInfo.packages` contain `maven:sentry-android-core`. It doesn't necessarily need to be the first item.
```suggestion
        assertTrue(sdkInfo.packages!!.any { sentryPackage -> sentryPackage.name ==  "maven:sentry-android-core"})
```

sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java
```diff
@@ -19,9 +21,18 @@ private SentryNdk() {}
...
+    addPackage(options);
     initSentryNative(options);
   }
+
+  private static void addPackage(@NotNull final SentryOptions options) {
+    final SdkVersion sdkVersion = options.getSdkVersion();
+    if (sdkVersion == null) {
+      return;
+    }
+    sdkVersion.addPackage("maven:sentry-android-ndk", BuildConfig.VERSION_NAME);
```
philipphofmann, line 36: `m`: Can we add a test for this?

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberIntegration.kt
```diff
@@ -36,4 +37,9 @@ class SentryTimberIntegration(
             }
         }
     }
+
+    private fun addPackage(options: SentryOptions) {
+        val sdkVersion = options.sdkVersion ?: return
+        sdkVersion.addPackage("maven:sentry-android-timber", BuildConfig.VERSION_NAME)
```
philipphofmann, line 43: `m`: Can we add a test for this?

sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java
```diff
@@ -19,9 +21,18 @@ private SentryNdk() {}
...
+  }
```
philipphofmann, line 37: `l`: We could only pass in the `sdkVersion`. `addPackage` doesn't need to know about `SentryOptions`. 
```suggestion
  private static void addPackage(final SdkVersion sdkVersion) {
    if (sdkVersion == null) {
      return;
    }
    sdkVersion.addPackage("maven:sentry-android-ndk", BuildConfig.VERSION_NAME);
  }
```

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeaderAdapter.java
```diff
@@ -25,44 +28,64 @@ public void write(JsonWriter writer, SentryEnvelopeHeader value) throws IOExcept
...
-        }
-        if (sdkInfo.getVersionMinor() != null) {
-          hasInitAttrs = initAttrs(writer, hasInitAttrs);
-          writer.name("version_minor").value(sdkInfo.getVersionMinor());
-        }
-        if (sdkInfo.getVersionPatchlevel() != null) {
-          hasInitAttrs = initAttrs(writer, hasInitAttrs);
-          writer.name("version_patchlevel").value(sdkInfo.getVersionPatchlevel());
+        final List<String> integrations = sdkVersion.getIntegrations();
+        if (integrations != null) {
```
philipphofmann, line 40: `l`: We could also check if `integrations` is not empty.
```suggestion
        if (integrations != null && !integrations.isEmpty()) {
```

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeaderAdapter.java
```diff
@@ -25,44 +28,64 @@ public void write(JsonWriter writer, SentryEnvelopeHeader value) throws IOExcept
...
+          writer.name("integrations").beginArray();
+
+          for (final String item : integrations) {
+            writer.value(item);
+          }
+
+          // integrations
```
philipphofmann, line 47: I really like those comments. Makes the code easier to understand 👍

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeaderAdapter.java
```diff
@@ -25,44 +28,64 @@ public void write(JsonWriter writer, SentryEnvelopeHeader value) throws IOExcept
...
-        }
-        if (sdkInfo.getVersionPatchlevel() != null) {
-          hasInitAttrs = initAttrs(writer, hasInitAttrs);
-          writer.name("version_patchlevel").value(sdkInfo.getVersionPatchlevel());
+        final List<String> integrations = sdkVersion.getIntegrations();
+        if (integrations != null) {
+          writer.name("integrations").beginArray();
+
+          for (final String item : integrations) {
+            writer.value(item);
```
philipphofmann, line 44: `l`: I prefer naming the items what they are. Makes the code easier to read IMO, but that's very nitpicking and personal preference. For `packages` below it makes more sense IMO, because when I look at `writer.name("name").value(item.getName());` I need to look 4 lines above to find out what the `item` is.
```suggestion
          for (final String integration : integrations) {
            writer.value(integration);
```

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -32,6 +34,25 @@
...
+    final SdkVersion sdkVersion = new SdkVersion();
+
+    sdkVersion.setName("sentry.java.android");
+    String version = BuildConfig.VERSION_NAME;
+    sdkVersion.setVersion(version);
+
+    // add 2 default packages
+    sdkVersion.addPackage("maven:sentry-android-core", version);
+    // TODO: sentry-core should add itself as the version may mismatch
+    sdkVersion.addPackage("maven:sentry-core", version);
```
bruno-garcia, line 51: Created issue to follow up: https://github.com/getsentry/sentry-java/projects/1#card-42202340

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -176,7 +175,7 @@ private void processNonCachedEvent(final @NotNull SentryEvent event) {
       event.setDebugMeta(getDebugMeta());
     }
     if (event.getSdk() == null) {
-      event.setSdk(getSdkVersion());
+      event.setSdk(options.getSdkVersion());
```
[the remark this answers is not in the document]
marandaneto: yep, it's our service locator as we'd need to get this info from Android and non-Android modules, so either we'd have an interface that queries this all the time, or just set it once and get from options, the later seemed better.

sentry-android-core/src/test/java/io/sentry/android/core/AndroidOptionsInitializerTest.kt
```diff
@@ -323,22 +323,20 @@ class AndroidOptionsInitializerTest {
...
+        val sdkVersion = sentryOptions.sdkVersion!!
 
-        assertEquals("sentry.java.android", sdkInfo.sdkName)
+        assertEquals("sentry.java.android", sdkVersion.name)
 
         // versions change on every release, so lets check only if its not null
-        assertNotNull(sdkInfo.versionMajor)
-        assertNotNull(sdkInfo.versionMinor)
-        assertNotNull(sdkInfo.versionPatchlevel)
+        assertNotNull(sdkVersion.version)
```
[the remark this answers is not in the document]
marandaneto: I've tried it and it didnt generate the BuildConfig for testing, but now it did, oh well. hehe thx

sentry-android-core/src/test/java/io/sentry/android/core/AndroidSerializerTest.kt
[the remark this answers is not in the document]
marandaneto: this test reads from the test resources `envelope_session_sdkversion.txt` and `maven:sentry-core` doesn't exist there.

sentry-android-core/src/test/java/io/sentry/android/core/DefaultAndroidEventProcessorTest.kt
```diff
@@ -165,14 +168,12 @@ class DefaultAndroidEventProcessorTest {
...
         val event = SentryEvent()
         processor.process(event, null)
-        assertNotNull(event.debugMeta.sdkInfo)
-        assertEquals("test", event.debugMeta.sdkInfo.sdkName)
-        assertEquals(1, event.debugMeta.sdkInfo.versionMajor)
-        assertEquals(2, event.debugMeta.sdkInfo.versionMinor)
-        assertEquals(3, event.debugMeta.sdkInfo.versionPatchlevel)
+        assertNotNull(event.sdk)
+        assertEquals("test", event.sdk.name)
+        assertEquals("1.2.3", event.sdk.version)
```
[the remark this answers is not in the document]
marandaneto: `AndroidOptionsInitializerTest.init should set SdkVersion` already test if all the values are set.
`DefaultAndroidEventProcessorTest` only asserts if the object itself is copied, I've changed to `assertSame` instead of asserting its values.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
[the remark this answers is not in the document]
philipphofmann: Now I get the point 👍 . I'm also going to do it like this once we need it in Cocoa.

sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java
marandaneto: gonna add it, I just moved to a different class though

sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java
marandaneto: makes sense

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeaderAdapter.java
marandaneto: `integrations` is always null or non-empty

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeaderAdapter.java
marandaneto: I've renamed it but I am not a big fan of it because the list is the plural of the item and it can get confused to read

sentry-core/src/main/java/io/sentry/core/protocol/SdkVersion.java
```diff
@@ -53,4 +54,12 @@ public void addIntegration(String integration) {
...
     this.unknown = unknown;
   }
+
+  public @Nullable List<SentryPackage> getPackages() {
+    return packages;
+  }
+
+  public @Nullable List<String> getIntegrations() {
+    return integrations;
+  }
```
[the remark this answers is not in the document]
marandaneto: I agree that SdkVersion is not the best name, but SdkInfo is a different object nominated by the unified API, we should not mix them.

I'd not rename SdkVersion now because it'll be a breaking change, we can do this on 3.0 if it's the case.
But totally against of adding those things to SdkInfo as this is a totally different object, we discussed about this on TSC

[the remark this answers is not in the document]
philipphofmann: Ah okay, I get it because of the breaking change.

sentry-android-core/src/test/java/io/sentry/android/core/DefaultAndroidEventProcessorTest.kt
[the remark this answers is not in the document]
philipphofmann: Okay didn't see the other test. My bad.

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeaderAdapter.java
philipphofmann: For now yes, but if we change `SdkVersion` and add a `removeIntegration`, which I doubt we gonna, we could have empty integrations. Anyway, the worst thing that can happen is to have an empty `integrations` in the JSON.

sentry-core/src/main/java/io/sentry/core/SentryEnvelopeHeaderAdapter.java
philipphofmann: As I said personal preference. Just wanted to point it out to learn from each other.

sentry-android-ndk/src/test/java/io/sentry/android/ndk/SentryNdkUtilTest.kt
```diff
@@ -0,0 +1,21 @@
+package io.sentry.android.ndk
+
+import io.sentry.core.SentryOptions
+import io.sentry.core.protocol.SdkVersion
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+class SentryNdkUtilTest {
```
philipphofmann, line 8: `m`: We could also add a test for passing `null` to `addPackage`.

PR #490, opened 2020-07-23
feat: drop store endpoint in favor of envelope

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
feat: drop store endpoint in favor of envelope

relay first

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
+++ b/sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
@@ -9,7 +9,6 @@ import com.jakewharton.nopen.annotation.Open;
 import io.sentry.core.ILogger;
 import io.sentry.core.ISerializer;
 import io.sentry.core.SentryEnvelope;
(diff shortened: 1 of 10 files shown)


sentry-core/src/main/java/io/sentry/core/transport/HttpTransport.java
```diff
@@ -139,42 +137,15 @@ public HttpTransport(
...
-
-      serializer.serialize(event, writer);
-    } catch (IOException e) {
-      logger.log(
-          ERROR, e, "An exception occurred while submitting the event to the Sentry server.");
-    } finally {
-      result =
-          readAndLog(connection, String.format("Event sent %s successfully.", event.getEventId()));
-    }
-    return result;
```
philipphofmann, line 177: So nice when we can get rid of complexity 👏

PR #491, opened 2020-07-23
ref: fix test linting

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
ref: fix test linting

linting was complaining

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

### DIFF
--- a/sentry-core/src/test/java/io/sentry/core/SentryClientTest.kt
+++ b/sentry-core/src/test/java/io/sentry/core/SentryClientTest.kt
@@ -404,19 +404,6 @@ class SentryClientTest {
     }
 
     @Test
-    fun `when transport gate is null, it should init an always on transport gate`() {
-        val sentryOptions: SentryOptions = SentryOptions().apply {
-            dsn = dsnString
-        }
-
-        val connection = mock<AsyncConnection>()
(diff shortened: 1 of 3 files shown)


sentry-core/src/test/java/io/sentry/core/SentryClientTest.kt
```diff
@@ -403,19 +403,6 @@ class SentryClientTest {
...
-            dsn = dsnString
-        }
-
-        val connection = mock<AsyncConnection>()
-        SentryClient(sentryOptions, connection)
-
-        assertNotNull(sentryOptions.transportGate)
-        assertTrue(sentryOptions.transportGate!!.isConnected)
-    }
-
```
philipphofmann, line 418: `m`: Why did we remove this test when we fix linting?

marandaneto: transport gate is never null, its NoOp, and there's a test for this already.
it was warning that `!!` is not necessary, but instead of fixing it, I removed the test which is useless after transportGate becoming NoOp in a different PR

PR #494, opened 2020-07-27
feat: rotate cache folder

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
feat: rotate cache folder

if the cache folder is full, we don't cache them anymore, ideally, we'd delete the oldest files making room for the new ones

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
+++ b/sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
@@ -18,8 +18,8 @@ abstract class CacheStrategy {
 
   protected final @NotNull SentryOptions options;
   protected final @NotNull ISerializer serializer;
-  protected @NotNull File directory;
-  private int maxSize;
+  protected final @NotNull File directory;
+  private final int maxSize;
 
   CacheStrategy(
       final @NotNull SentryOptions options,
--- a/sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
+++ b/sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
@@ -19,7 +19,7 @@ abstract class CacheStrategy {
   protected final @NotNull SentryOptions options;
   protected final @NotNull ISerializer serializer;
   protected @NotNull File directory;
-  protected int maxSize;
+  private int maxSize;
 
   CacheStrategy(
       final @NotNull SentryOptions options,
@@ -27,7 +27,7 @@ abstract class CacheStrategy {
       final int maxSize) {
     Objects.requireNonNull(directoryPath, "Directory is required.");
     this.options = Objects.requireNonNull(options, "SentryOptions is required.");
-        
+
     this.serializer = options.getSerializer();
     this.directory = new File(directoryPath);
 
@@ -35,7 +35,7 @@ abstract class CacheStrategy {
   }
 
   /**
-   * Check if a Dir. is valid and have write and read permission
+   * Check if a dir. is valid and have write and read permission
    *
    * @return true if valid and has permissions or false otherwise
    */
@@ -57,10 +57,9 @@ abstract class CacheStrategy {
    *
    * @param files the Files
    */
-  void sortFilesOldestToNewest(@NotNull File[] files) {
+  private void sortFilesOldestToNewest(@NotNull File[] files) {
     // just sort it if more than 1 file
     if (files.length > 1) {
-      // sort by the oldest to the newest
       Arrays.sort(files, (f1, f2) -> Long.compare(f1.lastModified(), f2.lastModified()));
     }
   }
@@ -84,12 +83,10 @@ abstract class CacheStrategy {
       for (int i = 0; i < totalToBeDeleted; i++) {
         final File file = files[i];
         // sanity check if the file actually exists.
-        if (file.exists()) {
-          if (!file.delete()) {
-            options
-                .getLogger()
-                .log(SentryLevel.WARNING, "File can't be deleted: %s", file.getAbsolutePath());
-          }
+        if (!file.delete()) {
+          options
+              .getLogger()
+              .log(SentryLevel.WARNING, "File can't be deleted: %s", file.getAbsolutePath());
         }
       }
     }
--- a/sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
+++ b/sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
@@ -25,10 +25,10 @@ abstract class CacheStrategy {
       final @NotNull SentryOptions options,
       final @NotNull String directoryPath,
       final int maxSize) {
-    this.options = Objects.requireNonNull(options, "SentryOptions is required.");
-    this.serializer = options.getSerializer();
-
     Objects.requireNonNull(directoryPath, "Directory is required.");
+    this.options = Objects.requireNonNull(options, "SentryOptions is required.");
+        
+    this.serializer = options.getSerializer();
     this.directory = new File(directoryPath);
 
     this.maxSize = maxSize;
--- a/sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
+++ b/sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
@@ -1,0 +1,97 @@
+package io.sentry.core.cache;
+
+import static io.sentry.core.SentryLevel.ERROR;
+
+import io.sentry.core.ISerializer;
+import io.sentry.core.SentryLevel;
+import io.sentry.core.SentryOptions;
+import io.sentry.core.util.Objects;
+import java.io.File;
+import java.nio.charset.Charset;
+import java.util.Arrays;
+import org.jetbrains.annotations.NotNull;
+
+abstract class CacheStrategy {
+
+  @SuppressWarnings("CharsetObjectCanBeUsed")
+  protected static final Charset UTF_8 = Charset.forName("UTF-8");
+
+  protected final @NotNull SentryOptions options;
+  protected final @NotNull ISerializer serializer;
+  protected @NotNull File directory;
+  protected int maxSize;
+
+  CacheStrategy(
+      final @NotNull SentryOptions options,
+      final @NotNull String directoryPath,
+      final int maxSize) {
+    this.options = Objects.requireNonNull(options, "SentryOptions is required.");
+    this.serializer = Objects.requireNonNull(options.getSerializer(), "Serializer is required.");
+
+    Objects.requireNonNull(directoryPath, "Directory is required.");
+    this.directory = new File(directoryPath);
+
+    this.maxSize = maxSize;
+  }
+
+  /**
+   * Check if a Dir. is valid, has write and read permission
+   *
+   * @return true if valid and has permission or false otherwise
+   */
+  protected boolean isDirectoryValid() {
+    if (!directory.isDirectory() || !directory.canWrite() || !directory.canRead()) {
+      options
+          .getLogger()
+          .log(
+              ERROR,
+              "The directory for caching files is inaccessible.: %s",
+              directory.getAbsolutePath());
+      return false;
+    }
+    return true;
+  }
+
+  /**
+   * Sort files from oldest to the newest using the lastModified method
+   *
+   * @param files the Files
+   */
+  void sortFilesOldestToNewest(@NotNull File[] files) {
+    // just sort it if more than 1 file
+    if (files.length > 1) {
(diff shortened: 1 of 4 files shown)


sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
```diff
@@ -0,0 +1,97 @@
...
+
+  CacheStrategy(
+      final @NotNull SentryOptions options,
+      final @NotNull String directoryPath,
+      final int maxSize) {
+    this.options = Objects.requireNonNull(options, "SentryOptions is required.");
+    this.serializer = options.getSerializer();
+
+    Objects.requireNonNull(directoryPath, "Directory is required.");
+    this.directory = new File(directoryPath);
```
bruno-garcia, line 32: L:

```suggestion
    Objects.requireNonNull(directoryPath, "Directory is required.");
    this.options = Objects.requireNonNull(options, "SentryOptions is required.");
        
    this.serializer = options.getSerializer();
    this.directory = new File(directoryPath);
```

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
```diff
@@ -0,0 +1,97 @@
...
+      final int totalToBeDeleted = (length - maxSize) + 1;
+
+      sortFilesOldestToNewest(files);
+
+      // delete files from the top of the Array as its sorted by the oldest to the newest
+      for (int i = 0; i < totalToBeDeleted; i++) {
+        final File file = files[i];
+        // sanity check if the file actually exists.
+        if (file.exists()) {
+          if (!file.delete()) {
```
bruno-garcia, line 88: M: Since this is inherently racey I ask: Does delete throw here if the file doesn't exist?
If it doesn't, maybe we can skip the `exists` altogether.
If it does, we should wrap this on a try catch anyway.

The `exists` is likely an I/O we could skip since we'll be hitting the disk for the `delete` anyway

sentry-core/src/test/java/io/sentry/core/cache/CacheStrategyTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        // this test assumes that the dir. has write/read permission.
+        assertTrue(sut.isDirectoryValid)
+    }
+
+    @Test
+    fun `Sort files from the oldest to the newest`() {
+        val sut = fixture.getSUT()
+
+        val files = createTempFiles()
+        val reverseFiles = files.reversedArray()
```
bruno-garcia, line 45: L: It's unclear that `files` is in the right order to begin with without looking into it.
What about we flip a couple explicitly to be more clear they are out of order:

```suggestion
        val reverseFiles = files.reversedArray()
        val reverseFiles[0] = files[1];
        val reverseFiles[1] = files[0];
```

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
```diff
@@ -0,0 +1,97 @@
...
+import io.sentry.core.ISerializer;
+import io.sentry.core.SentryLevel;
+import io.sentry.core.SentryOptions;
+import io.sentry.core.util.Objects;
+import java.io.File;
+import java.nio.charset.Charset;
+import java.util.Arrays;
+import org.jetbrains.annotations.NotNull;
+
+abstract class CacheStrategy {
```
philipphofmann, line 14: `m`: I think we could split up this class into two classes. For me, we have two different responsibilities in here. We have some base functionality that both `DiskCache` and `SessionCache` need. And we have the sorting and deleting of the files. What about renaming `CacheStrategy` to something like `SentryCache` or `BaseCache`, although I'm not a fan of naming something base, ... and extracting `sortFilesOldestToNewest` and `rotateCacheIfNeeded` to a class called `FileCacheStrategy` or something similar. I'm a huge fan of [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance). So with the extracted class, we wouldn't have to create `CustomCache` in the `CacheStrategyTest` to be able to test the deleting and sorting of the files.

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @SuppressWarnings("CharsetObjectCanBeUsed")
+  protected static final Charset UTF_8 = Charset.forName("UTF-8");
+
+  protected final @NotNull SentryOptions options;
+  protected final @NotNull ISerializer serializer;
+  protected @NotNull File directory;
+  protected int maxSize;
```
philipphofmann, line 22: `m`: I think this can be private.

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
```diff
@@ -0,0 +1,97 @@
...
+import org.jetbrains.annotations.NotNull;
+
+abstract class CacheStrategy {
+
+  @SuppressWarnings("CharsetObjectCanBeUsed")
+  protected static final Charset UTF_8 = Charset.forName("UTF-8");
+
+  protected final @NotNull SentryOptions options;
+  protected final @NotNull ISerializer serializer;
+  protected @NotNull File directory;
```
philipphofmann, line 21: `m`: This is currently read-only for `SessionCache` and `DiskCache`. I would enforce this by making `directory` private and creating a getter for it.

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
```diff
@@ -0,0 +1,97 @@
...
+    this.serializer = options.getSerializer();
+
+    Objects.requireNonNull(directoryPath, "Directory is required.");
+    this.directory = new File(directoryPath);
+
+    this.maxSize = maxSize;
+  }
+
+  /**
+   * Check if a Dir. is valid and have write and read permission
```
philipphofmann, line 38: `l`: I guess it should be `dir.` or `directory`? The capital `D` is confusing for me.

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+    return true;
+  }
+
+  /**
+   * Sort files from oldest to the newest using the lastModified method
+   *
+   * @param files the Files
+   */
+  void sortFilesOldestToNewest(@NotNull File[] files) {
```
philipphofmann, line 60: `m`: Why is this not private? Only `CacheStrategyTest` is calling it. Can't we test it with `rotateCacheIfNeeded`?

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
```diff
@@ -0,0 +1,97 @@
...
+    // just sort it if more than 1 file
+    if (files.length > 1) {
+      // sort by the oldest to the newest
```
philipphofmann, line 63: `l`: I think we could skip this comment here as the method name already tells me what it does.

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
```diff
@@ -0,0 +1,97 @@
...
+      Arrays.sort(files, (f1, f2) -> Long.compare(f1.lastModified(), f2.lastModified()));
+    }
+  }
+
+  /**
+   * Rotates the caching folder if full, deleting the oldest files first
+   *
+   * @param files the Files
+   */
+  protected void rotateCacheIfNeeded(final @NotNull File[] files) {
```
philipphofmann, line 73: `l`: I'm a bit confused by the naming with rotating. Wouldn't `clearCacheIfNeeded` be more descriptive?

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
marandaneto: the problem is to log it, `delete` returns `false` even if the file doesn't exist, it doesn't throw, so we don't know if the file doesn't exist or if it had a problem deleting it.

sentry-core/src/test/java/io/sentry/core/cache/CacheStrategyTest.kt
marandaneto: > It's unclear that files is in the right order to begin with without looking into it

I don't follow it, I created the array in the right order, and I use the extension function to reverse it, what is unclear? the impl. of the `reversedArray` do you mean?

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
marandaneto: not sure if it's worth creating another extra class to achieve this, actually only `we wouldn't have to create CustomCache in the CacheStrategyTest` is not enough argument as this is not added to the classpath, only on tests.

I still would need a class + an interface, ideally `DiskCache` and `SessionCache` will be merged very soon (when events become purely envelopes) and this class might lose its meaning, I mean, it'll be a single class anyway

opinions @bruno-garcia ?

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
marandaneto: I wanted to test this in isolation, because its package-private and can't be used anyway outside of this package, but will change it.

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
marandaneto: true, I renamed the method, now the comment doesn't make sense, will change it

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
marandaneto: `clearCacheIfNeeded` -> clear would look like, cleaning the whole folder

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
bruno-garcia: I'm also a huge fan of composition over inheritence but IMHO this is a good use of inheritance.

Also, this stuff is internal so no need to over engineer this and break it apart. If it grows in the upcoming changes, we can refactor then.

That said, IMHO LGTM

sentry-core/src/test/java/io/sentry/core/cache/CacheStrategyTest.kt
bruno-garcia: `createTempFIles` didn't really tell me what order they came in, so I wrote my comment

It's `L:` nonetheless so feel free to ignore

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
bruno-garcia: I'd really just get rid of it.

In between checking `exists` and calling `delete` the file could have been deleted so it's anyway not 100%

If delete is false then it didn't delete and there's no much more we can do anyway but to log and go on.

sentry-core/src/test/java/io/sentry/core/cache/CacheStrategyTest.kt
marandaneto: true that, i will rename the aux. method then for something more specific

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
philipphofmann: I already stated my opinion and still with your arguments for keeping as it is I'm in favor of splitting it up, but I have no strong opinion on this as it's just a small class. I'm fine with keeping as it is. No need for further discussion IMO.

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
philipphofmann: Thanks for pointing this out. I'm not really familiar with those terms 👏

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
marandaneto: they are `final` so read-only anyway

PR #497, opened 2020-07-28
Call `Sentry#close` on JVM shutdown.

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

- Call `Sentry#close` on JVM shutdown - as a result flush all Sentry events that are in progress.
- Configure `SendCachedEventFireAndForgetIntegration` by default only for Android as for the console/server application cache does not apply

Without this change, when Sentry is used in the console application, the main thread may finish before events are sent to Sentry and as a result no events are sent.

This idea was briefly discussed with @marandaneto and it's a subject for discussion - perhaps there is a better way to achieve the same result without attaching shutdown hook.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -9,6 +9,7 @@ import android.os.Build;
 import io.sentry.core.ILogger;
 import io.sentry.core.SendCachedEventFireAndForgetIntegration;
 import io.sentry.core.SendFireAndForgetEnvelopeSender;
+import io.sentry.core.SendFireAndForgetEventSender;
 import io.sentry.core.SentryLevel;
 import io.sentry.core.SentryOptions;
 import io.sentry.core.util.Objects;
@@ -114,6 +115,14 @@ final class AndroidOptionsInitializer {
       final @NotNull SentryOptions options,
       final @NotNull IBuildInfoProvider buildInfoProvider,
       final @NotNull ILoadClass loadClass) {
+
+    options.addIntegration(
+        new SendCachedEventFireAndForgetIntegration(
+            new SendFireAndForgetEventSender(() -> options.getCacheDirPath())));
+
+    options.addIntegration(
+        new SendCachedEventFireAndForgetIntegration(
+            new SendFireAndForgetEnvelopeSender(() -> options.getSessionsPath())));
 
     // Integrations are registered in the same order. NDK before adding Watch outbox,
     // because sentry-native move files around and we don't want to watch that.
--- a/sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
+++ b/sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
@@ -1,0 +1,10 @@
+package io.sentry.core;
+
+/** Registers hook that closes {@link Sentry Sentry SDK} when main thread shuts down. */
+public final class ShutdownHookIntegration implements Integration {
+
+  @Override
+  public void register(IHub hub, SentryOptions options) {
+    Runtime.getRuntime().addShutdownHook(new Thread(Sentry::close));
+  }
+}
--- a/sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
+++ b/sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
@@ -1,10 +1,21 @@
 package io.sentry.core;
 
-/** Registers hook that closes {@link Sentry Sentry SDK} when main thread shuts down. */
+import io.sentry.core.util.Objects;
+import org.jetbrains.annotations.NotNull;
+
+/** Registers hook that closes {@link Hub} when main thread shuts down. */
 public final class ShutdownHookIntegration implements Integration {
 
+  private final @NotNull Runtime runtime;
+
+  public ShutdownHookIntegration(final @NotNull Runtime runtime) {
+    this.runtime = Objects.requireNonNull(runtime, "Runtime is required");
+  }
+
   @Override
-  public void register(IHub hub, SentryOptions options) {
-    Runtime.getRuntime().addShutdownHook(new Thread(Sentry::close));
+  public void register(@NotNull IHub hub, @NotNull SentryOptions options) {
+    Objects.requireNonNull(hub, "Hub is required");
+
+    runtime.addShutdownHook(new Thread(() -> hub.close()));
   }
 }
--- a/sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
+++ b/sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
@@ -2,14 +2,20 @@ package io.sentry.core;
 
 import io.sentry.core.util.Objects;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.TestOnly;
 
 /** Registers hook that closes {@link Hub} when main thread shuts down. */
 public final class ShutdownHookIntegration implements Integration {
 
   private final @NotNull Runtime runtime;
 
+  @TestOnly
   public ShutdownHookIntegration(final @NotNull Runtime runtime) {
     this.runtime = Objects.requireNonNull(runtime, "Runtime is required");
+  }
+
+  public ShutdownHookIntegration() {
+    this(Runtime.getRuntime());
   }
 
   @Override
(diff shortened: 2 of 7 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -115,6 +116,14 @@ private static void installDefaultIntegrations(
       final @NotNull IBuildInfoProvider buildInfoProvider,
       final @NotNull ILoadClass loadClass) {
 
+    options.addIntegration(
```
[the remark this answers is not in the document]
maciejwalkowiak: 👍  Just so that it doesn't get lost - I think we should expose ideally single method that turns on caching which takes `cachePathDir` as parameter and under the hood also turns on all cache-related integrations.

sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
```diff
@@ -0,0 +1,10 @@
+package io.sentry.core;
+
+/** Registers hook that closes {@link Sentry Sentry SDK} when main thread shuts down. */
+public final class ShutdownHookIntegration implements Integration {
+
+  @Override
+  public void register(IHub hub, SentryOptions options) {
+    Runtime.getRuntime().addShutdownHook(new Thread(Sentry::close));
```
marandaneto, line 8: H: I'd avoid method references, I already found issues with it depending on the tooling that does desugaring/obfuscation.
Eg: https://github.com/getsentry/sentry-android/blob/master/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java#L109-L110

sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
```diff
@@ -0,0 +1,10 @@
+package io.sentry.core;
+
+/** Registers hook that closes {@link Sentry Sentry SDK} when main thread shuts down. */
+public final class ShutdownHookIntegration implements Integration {
+
+  @Override
+  public void register(IHub hub, SentryOptions options) {
```
marandaneto, line 7:
```suggestion
  public void register(final @NotNull IHub hub, final @NotNull SentryOptions options) {
```

sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
marandaneto: Also wondering if it's worth to get Runtime as a ctor (package private) param to be able to mock and test it, a ctor overload could pass the `Runtime.getRuntime()` down to the other ctor.

marandaneto, line 8:
```suggestion
    Runtime.getRuntime().addShutdownHook(new Thread(hub::close));
```

maciejwalkowiak:
Thanks @marandaneto for a review. Fixes are applied. 

Regarding target branch, my understanding is that everything that affects Sentry Core goes to master and only other framework-specific integrations are meant to go to `feat/sentry-java`. In practice - next PR will contain sample console application - it would also go to `master`, but the Logback integration that will follow would go to `feat/sentry-java`.

sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
```diff
@@ -0,0 +1,21 @@
...
+public final class ShutdownHookIntegration implements Integration {
+
+  private final @NotNull Runtime runtime;
+
+  public ShutdownHookIntegration(final @NotNull Runtime runtime) {
+    this.runtime = Objects.requireNonNull(runtime, "Runtime is required");
+  }
+
+  @Override
+  public void register(@NotNull IHub hub, @NotNull SentryOptions options) {
```
marandaneto, line 16:
```suggestion
  public void register(final @NotNull IHub hub, final @NotNull SentryOptions options) {
```

marandaneto: any reason to not apply the `final` modifier?

marandaneto:
> Thanks @marandaneto for a review. Fixes are applied.
> 
> Regarding target branch, my understanding is that everything that affects Sentry Core goes to master and only other framework-specific integrations are meant to go to `feat/sentry-java`. In practice - next PR will contain sample console application - it would also go to `master`, but the Logback integration that will follow would go to `feat/sentry-java`.

I see, that makes sense, I'm just wondering if the `ShutdownHookIntegration` should be also a default integration for Android? does the Dalvik/Art behave similarly? I'd rather make some tests before merging it, the idea is to have `master` also stable so we could publish a new version without any problems, so said that, Have you tested the `ShutdownHook` on Android when there's an UncaughtException? if so, happy to approve and merge it.

sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
maciejwalkowiak: No specific reason, I've seen some methods with `final` and some without. I'll add them.

My personal take on `final` on method parameters is that they clutter code and instead all parameters should be treated as `final` implicitly. That's just an opinion, I will adjust to existing coding standards.

maciejwalkowiak:
I didn't find any documentation stating that shutdown hooks behave differently on Android but I haven't tested it and I also don't think I am the right person to do it as my Android experience is limited (I can though it will just take time). I'll leave it up to you and @bruno-garcia to decide how to proceed.

sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
maciejwalkowiak: 👍

sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
marandaneto: got it, just asking cus it was in my first GH "suggestion".

Agreed that they should be treated as `final` implicitly, but IMO when you have multithread systems, having an extra `final` makes the compiler to help you out if you are not cloning/deep copying an object but instead messing up with the references, we had a bug like this a few months ago, so we've decided to add in all the missing ones.

maciejwalkowiak: Perhaps there is a way to configure static code analysis to make sure it doesn't sneak in?

marandaneto: all the data classes are mutable by design, I've tried once but then it warned about almost everything.
so basically what I am doing is, new things I always try to make everything `final` if it can and if I am touching some classes, I already check the missing `final` and `@NotNull`, `@Nullable` to make better interoperability with Kotlin tests

marandaneto:
> I didn't find any documentation stating that shutdown hooks behave differently on Android but I haven't tested it and I also don't think I am the right person to do it as my Android experience is limited (I can though it will just take time). I'll leave it up to you and @bruno-garcia to decide how to proceed.

let's merge then and I test it, last case we just remove adding this integration by default, the integration itself is already a big gain.
so I don't need to fork your repo, I'll take care of it, thanks!

sentry-core/src/main/java/io/sentry/core/ShutdownHookIntegration.java
```diff
@@ -0,0 +1,21 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/** Registers hook that closes {@link Hub} when main thread shuts down. */
+public final class ShutdownHookIntegration implements Integration {
+
+  private final @NotNull Runtime runtime;
+
+  public ShutdownHookIntegration(final @NotNull Runtime runtime) {
+    this.runtime = Objects.requireNonNull(runtime, "Runtime is required");
+  }
```
marandaneto, line 13:
```suggestion
  ShutdownHookIntegration(final @NotNull Runtime runtime) {
    this.runtime = Objects.requireNonNull(runtime, "Runtime is required");
  }
  
  public ShutdownHookIntegration() {
    this(Runtime.getRuntime());
  }
```

maciejwalkowiak: ... and invoke non-param constructor in `SentryOptions`, right?

marandaneto: yep, this is just for the sake of testing, one only needs `new ShutdownHookIntegration()`

PR #498, opened 2020-07-29
Log stacktraces in SystemOutLogger

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Log stacktraces in SystemOutLogger when exception occurs.

With `debug` mode on, when exception happens it's currently difficult to find out why and where did it happen. With stacktraces printed to the console we can easily find the bug.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/SystemOutLogger.java
+++ b/sentry-core/src/main/java/io/sentry/core/SystemOutLogger.java
@@ -64,7 +64,7 @@ public final class SystemOutLogger implements ILogger {
     }
   }
 
-  private String captureStackTrace(@NotNull Throwable throwable) {
+  private @NotNull String captureStackTrace(final @NotNull Throwable throwable) {
     StringWriter stringWriter = new StringWriter();
     PrintWriter printWriter = new PrintWriter(stringWriter);
     throwable.printStackTrace(printWriter);
--- a/sentry-core/src/main/java/io/sentry/core/SystemOutLogger.java
+++ b/sentry-core/src/main/java/io/sentry/core/SystemOutLogger.java
@@ -30,10 +30,14 @@ public final class SystemOutLogger implements ILogger {
   @SuppressWarnings("AnnotateFormatMethod")
   @Override
   public void log(SentryLevel level, String message, Throwable throwable) {
-    System.out.println(
-        String.format(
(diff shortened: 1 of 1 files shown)


sentry-core/src/main/java/io/sentry/core/SystemOutLogger.java
```diff
@@ -42,6 +48,18 @@ public void log(SentryLevel level, String message, Throwable throwable) {
   @Override
   public void log(SentryLevel level, Throwable throwable, String message, Object... args) {
     System.out.println(
-        String.format("%s: %s \n %s", level, String.format(message, args), throwable.toString()));
+        String.format(
```
bruno-garcia, line 51: Makes me realize the previous code was a bit naive but don't we need to check if we have an instance in the first place?
I'm afraid we might throw if any arg is null here

maciejwalkowiak: True! Good catch. Now I see that `throwable` is marked as `@Nullable` in `DiagnosticLogger#log`. I'll take care of it.

sentry-core/src/main/java/io/sentry/core/SystemOutLogger.java
```diff
@@ -41,7 +51,23 @@ public void log(SentryLevel level, String message, Throwable throwable) {
...
+          String.format(
+              "%s: %s \n %s\n%s",
+              level,
+              String.format(message, args),
+              throwable.toString(),
+              captureStackTrace(throwable)));
+    }
+  }
+
+  private String captureStackTrace(@NotNull Throwable throwable) {
```
marandaneto, line 67:
```suggestion
  private @NotNull String captureStackTrace(final @NotNull Throwable throwable) {
```

sentry-core/src/main/java/io/sentry/core/SystemOutLogger.java
marandaneto: good catch, we've never used this class, that's why not so feature-complete (missing stack traces, etc...), we're always in favor of AndroidLogger for Android.

PR #499, opened 2020-07-29
Add console application sample.

maciejwalkowiak:
##  :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Added console application sample. Code is a direct translation from console sample in `sentry-dotnet` repository excluding features that are not implemented in `sentry-java`.

Gradle contains minimal set of configuration to avoid confusion what's required to use Sentry. The only extra task that is added is one that builds fat jar - it's a subject to discussion if we need it here.

Sample console application is a living documentation and guides users through Java SDK features.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sentry-samples/sentry-samples-console/build.gradle.kts
+++ b/sentry-samples/sentry-samples-console/build.gradle.kts
@@ -1,0 +1,28 @@
+plugins {
+    java
+    id(Config.QualityPlugins.gradleVersions)
+}
+
+configure<JavaPluginConvention> {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+dependencies {
+    implementation(project(":sentry-core"))
+}
+
+val fatJar = task("fatJar", type = Jar::class) {
+    archiveFileName.set("${project.name}-fat.jar")
+    manifest {
+        attributes["Main-Class"] = "io.sentry.samples.console.Main"
+    }
+    from(configurations.runtimeClasspath.get().map { if (it.isDirectory) it else zipTree(it) })
+    with(tasks.jar.get() as CopySpec)
+}
+
+tasks {
+    "build" {
+        dependsOn(fatJar)
+    }
+}
--- a/sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
+++ b/sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
@@ -81,30 +81,9 @@ public class Main {
     // Only data added to the scope on `configureScope` above is included.
     Sentry.captureMessage("Some warning!", SentryLevel.WARNING);
 
-    Exception exception = new RuntimeException("Attempting to send this multiple times");
-
-    // Only the first capture will be sent to Sentry
-    for (int i = 0; i < 3; i++) {
-      // The SDK is able to detect duplicate events:
-      // This is useful, for example, when multiple loggers log the same exception. Or exception is
-      // re-thrown and recaptured.
-      Sentry.captureException(exception);
-    }
-
-    int count = 10;
-    for (int i = 0; i < count; i++) {
-      String msg = "%d of %d items we'll wait to flush to Sentry!";
-      Message message = new Message();
-      message.setMessage(msg);
-      message.setFormatted(String.format(msg, i, count));
-      SentryEvent event = new SentryEvent();
-      event.setMessage(message);
-      Sentry.captureEvent(event, SentryLevel.DEBUG);
-    }
-
-    // Console output will show queue being flushed. Task completes then and timeout is never
-    // reached (you don't need to wait a day :)
-    Sentry.flush(10000);
+    // Sending exception:
+    Exception exception = new RuntimeException("Some error!");
+    Sentry.captureException(exception);
 
     // An event with breadcrumb and user data
     SentryEvent evt = new SentryEvent();
--- a/sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
+++ b/sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
@@ -1,0 +1,132 @@
+package io.sentry.samples.console;
+
+import io.sentry.core.Breadcrumb;
+import io.sentry.core.EventProcessor;
+import io.sentry.core.Sentry;
+import io.sentry.core.SentryEvent;
+import io.sentry.core.SentryLevel;
+import io.sentry.core.protocol.Message;
+import io.sentry.core.protocol.User;
+import java.util.Collections;
+
+public class Main {
+
+  public static void main(String[] args) {
+    Sentry.init(
+        options -> {
+          // NOTE: Replace the test DSN below with YOUR OWN DSN to see the events from this app in
+          // your Sentry project/dashboard
+          options.setDsn("https://f7f320d5c3a54709be7b28e0f2ca7081@sentry.io/1808954");
+
+          // Modifications to event before it goes out. Could replace the event altogether
+          options.setBeforeSend(
+              (event, hint) -> {
+                // Drop an event altogether:
+                if (event.getTag("SomeTag") != null) {
+                  return null;
+                }
+                return event;
+              });
+
+          // Allows inspecting and modifying, returning a new or simply rejecting (returning null)
+          options.setBeforeBreadcrumb(
+              (breadcrumb, hint) -> {
+                // Don't add breadcrumbs with message containing:
+                if (breadcrumb.getMessage() != null
+                    && breadcrumb.getMessage().contains("bad breadcrumb")) {
+                  return null;
+                }
+                return breadcrumb;
+              });
+
+          // Configure the background worker which sends events to sentry:
+          // Wait up to 5 seconds before shutdown while there are events to send.
+          options.setShutdownTimeout(5000);
+
+          // Enable SDK logging with Debug level
+          options.setDebug(true);
+          // To change the verbosity, use:
+          options.setDiagnosticLevel(SentryLevel.LOG);
+
+          // Using a proxy:
+          options.setProxy(null); // new Proxy(Type.HTTP, new InetSocketAddress(8090)));
+
+          // Exclude classes from certain packages from stack traces sent to Sentry:
+          options.addInAppExclude("org.jboss");
+        });
+
+    Sentry.addBreadcrumb(
+        "A 'bad breadcrumb' that will be rejected because of 'BeforeBreadcrumb callback above.'");
+
+    // Data added to the root scope (no PushScope called up to this point)
+    // The modifications done here will affect all events sent and will propagate to child scopes.
+    Sentry.configureScope(
+        scope -> {
+          scope.addEventProcessor(new SomeEventProcessor());
+
+          scope.setExtra("SomeExtraInfo", "Some value for extra info");
+        });
+
+    // Configures a scope which is only valid within the callback
+    Sentry.withScope(
+        scope -> {
+          scope.setLevel(SentryLevel.FATAL);
+          scope.setTransaction("main");
+
+          Sentry.captureMessage("Fatal message!");
+        });
+
+    Sentry.captureMessage("Some warning!", SentryLevel.WARNING);
+
+    Exception exception = new RuntimeException("Attempting to send this multiple times");
+
+    // Only the first capture will be sent to Sentry
+    for (int i = 0; i < 3; i++) {
+      // The SDK is able to detect duplicate events:
+      // This is useful, for example, when multiple loggers log the same exception. Or exception is
+      // re-thrown and recaptured.
+      Sentry.captureException(exception);
+    }
+
+    int count = 10;
+    for (int i = 0; i < count; i++) {
+      String msg = "%d of %d items we'll wait to flush to Sentry!";
+      Message message = new Message();
+      message.setMessage(msg);
+      message.setFormatted(String.format(msg, i, count));
+      SentryEvent event = new SentryEvent();
+      event.setMessage(message);
+      Sentry.captureEvent(event, SentryLevel.DEBUG);
+    }
+
+    // Console output will show queue being flushed. Task completes then and timeout is never
+    // reached (you don't need to wait a day :)
+    Sentry.flush(10000);
+
+    // An event with breadcrumb and user data
+    SentryEvent evt = new SentryEvent();
+    Message msg = new Message();
+    msg.setMessage("Detailed event");
+    evt.setMessage(msg);
+    evt.addBreadcrumb(new Breadcrumb("Breadcrumb directly to the event"));
+    User user = new User();
+    user.setUsername("some@user");
+    evt.setUser(user);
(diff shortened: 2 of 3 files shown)


sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.Collections;
+
+public class Main {
+
+  public static void main(String[] args) {
+    Sentry.init(
+        options -> {
+          // NOTE: Replace the test DSN below with YOUR OWN DSN to see the events from this app in
+          // your Sentry project/dashboard
+          options.setDsn("https://f7f320d5c3a54709be7b28e0f2ca7081@sentry.io/1808954");
```
bruno-garcia, line 19: Lets have this one by default (used in the other sample in the repo)
```suggestion
          options.setDsn("https://f7f320d5c3a54709be7b28e0f2ca7081@sentry.io/1808954");
```

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
```diff
@@ -0,0 +1,132 @@
...
+              });
+
+          // Configure the background worker which sends events to sentry:
+          // Wait up to 5 seconds before shutdown while there are events to send.
+          options.setShutdownTimeout(5000);
+
+          // Enable SDK logging with Debug level
+          options.setDebug(true);
+          // To change the verbosity, use:
+          options.setDiagnosticLevel(SentryLevel.LOG);
```
bruno-garcia, line 49: Maybe we note the default and also change to something else (`LOG` is some baggage to support React Native, IMHO `LOG` isn't a level at all and should be gone) :~

```suggestion
          // By default it's DEBUG.
          options.setDiagnosticLevel(SentryLevel.ERROR);
```

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
```diff
@@ -0,0 +1,132 @@
...
+
+          // Using a proxy:
+          options.setProxy(null); // new Proxy(Type.HTTP, new InetSocketAddress(8090)));
+
+          // Exclude classes from certain packages from stack traces sent to Sentry:
```
bruno-garcia, line 54: It's actually just to hide the frames by default. You can click "show all" and the complete stack trace shows up

Maybe something like:
```suggestion
          // Exclude frames from some packages from being "inApp" so are hidden by default in Sentry UI:
```

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
```diff
@@ -0,0 +1,132 @@
...
+    // Configures a scope which is only valid within the callback
+    Sentry.withScope(
+        scope -> {
+          scope.setLevel(SentryLevel.FATAL);
+          scope.setTransaction("main");
+
+          Sentry.captureMessage("Fatal message!");
+        });
+
+    Sentry.captureMessage("Some warning!", SentryLevel.WARNING);
```
bruno-garcia, line 79:
```suggestion
    // Only data added to the scope on `configureScope` above is included.
    Sentry.captureMessage("Some warning!", SentryLevel.WARNING);
```

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
```diff
@@ -0,0 +1,132 @@
...
+          scope.setExtra("SomeExtraInfo", "Some value for extra info");
+        });
+
+    // Configures a scope which is only valid within the callback
+    Sentry.withScope(
+        scope -> {
+          scope.setLevel(SentryLevel.FATAL);
+          scope.setTransaction("main");
+
+          Sentry.captureMessage("Fatal message!");
```
bruno-garcia, line 76:
```suggestion
          // This message includes the data set to the scope in this block:
          Sentry.captureMessage("Fatal message!");
```

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
```diff
@@ -0,0 +1,132 @@
...
+    Sentry.captureMessage("Some warning!", SentryLevel.WARNING);
+
+    Exception exception = new RuntimeException("Attempting to send this multiple times");
+
+    // Only the first capture will be sent to Sentry
+    for (int i = 0; i < 3; i++) {
+      // The SDK is able to detect duplicate events:
+      // This is useful, for example, when multiple loggers log the same exception. Or exception is
+      // re-thrown and recaptured.
+      Sentry.captureException(exception);
```
bruno-garcia, line 88: Not sure we have this de-duplication here though, need to check

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
```diff
@@ -0,0 +1,132 @@
...
+      message.setMessage(msg);
+      message.setFormatted(String.format(msg, i, count));
+      SentryEvent event = new SentryEvent();
+      event.setMessage(message);
+      Sentry.captureEvent(event, SentryLevel.DEBUG);
+    }
+
+    // Console output will show queue being flushed. Task completes then and timeout is never
+    // reached (you don't need to wait a day :)
+    Sentry.flush(10000);
```
bruno-garcia, line 104: Not sure this will work. It should but we need to work on `flush`

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
```diff
@@ -0,0 +1,132 @@
...
+
+    // An event with breadcrumb and user data
+    SentryEvent evt = new SentryEvent();
+    Message msg = new Message();
+    msg.setMessage("Detailed event");
+    evt.setMessage(msg);
+    evt.addBreadcrumb(new Breadcrumb("Breadcrumb directly to the event"));
```
bruno-garcia, line 111: I'd love for us to have an overload like this:

```suggestion
    evt.addBreadcrumb("Breadcrumb directly to the event");
```

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
bruno-garcia: Oh nevermind, it's the right DSN :)

maciejwalkowiak: They look the same to me..?

bruno-garcia: I need to get my brain checked, sorry

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
maciejwalkowiak: I'll add it in separate PR.

maciejwalkowiak: https://github.com/getsentry/sentry-android/pull/500

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
maciejwalkowiak: You're right - it doesn't. `SentryClient#flush` is literally `// TODO: Flush transport` :-)

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
maciejwalkowiak: You're right, the deduplication does not happen.

marandaneto: yep, it's a TODO

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
marandaneto: PR is approved.
although the Sentry static class has overloads for adding breadcrumbs like this, also a few others overload.
new SentryEvent() and calling adders and setters were never intended to be a public API iirc, so not sure if it's worth the effort of adding methods to this class, most probably users will be doing Sentry.addBreadcrumb(...)

sentry-samples/sentry-samples-console/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+    }
+    from(configurations.runtimeClasspath.get().map { if (it.isDirectory) it else zipTree(it) })
+    with(tasks.jar.get() as CopySpec)
+}
+
+tasks {
+    "build" {
+        dependsOn(fatJar)
+    }
+}
```
marandaneto, line 28: nothing against doing it, but what's about sparing some bits and using a Gradle plugin? this might break when upgrading Gradle versions and usually plugins react fast to it.
eg https://github.com/johnrengelman/shadow

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
```diff
@@ -0,0 +1,114 @@
...
+          options.setShutdownTimeout(5000);
+
+          // Enable SDK logging with Debug level
+          options.setDebug(true);
+          // To change the verbosity, use:
+          // By default it's DEBUG.
+          options.setDiagnosticLevel(SentryLevel.ERROR); //  A good option to have SDK debug log in prod is to use only level ERROR here.
+
+          // Using a proxy:
+          options.setProxy(null); // new Proxy(Type.HTTP, new InetSocketAddress(8090)));
```
marandaneto, line 53: I've never tested this class, to be honest, it's pretty much a copy-paste from sentry-java, it's fine to keep here but recently @thinkocapo could not use it for some reason and I still need to figure out if it's a bug or a proxy setup.

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
marandaneto: yep, I'd rather keep `flush` out of this sample, but change it to `close` which is already implemented?

marandaneto:
> I'd also suggest we merge this into master. What do you think?

it'd be fine to go into master as this has no breaking changes, just a sample, although #500 should be merged into master too and I'm fine with it, let's do it

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
```diff
@@ -0,0 +1,114 @@
...
+import io.sentry.core.SentryLevel;
+import io.sentry.core.protocol.Message;
+import io.sentry.core.protocol.User;
+import java.util.Collections;
+
+public class Main {
+
+  public static void main(String[] args) {
+    Sentry.init(
+        options -> {
```
marandaneto, line 16: I'd also set:

```
options.setRelease("io.sentry.samples.console@3.0.0+1");
options.setSentryClientName("sentry-java/3.0.0");
```

while we don't read this automatically from the build conf/CI etc...so if people see this example that will land in master, they at least know that this is needed right now.

sentry-samples/sentry-samples-console/build.gradle.kts
maciejwalkowiak: I wanted to avoid extra dependency that's why I don't go with plugin. 

If people build standalone CLI apps then it's common to build fat jars. It only depends if we want this example to show the code or also to actually let users run the sample application.

marandaneto: I believe both, but if they run, it'll be thru the IDE anyway, we don't export/allow to download the final jars/apks, it might be a good idea, but for that, we also have our official examples repo.
https://github.com/getsentry/examples

the idea of having samples on this repo is to easier the development, so you can run and debug the SDK and samples easily.

of course also a resource of learning.

that said, I'd go either with the plugin, so we don't need to maintain this piece of code or to not generate a fat jar at all.
thoughts @bruno-garcia ?

maciejwalkowiak:
Once @marandaneto or @bruno-garcia merge #501 I'll add all the polishing to this PR.

sentry-samples/sentry-samples-console/build.gradle.kts
maciejwalkowiak: I am getting convinced that we can just remove fat jar building. It can be always added later if we find there is a need for that.

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
maciejwalkowiak: Since `close` is executed anyway at the JVM termination I am not sure if this wouldn't be confusing. I'll leave it up to you and @bruno-garcia. This is anyway not the final version of the sample as it will get likely updated once more features are added to Java SDK.

marandaneto: that's true, if the integration call closes automatically by default, no need, or, to still sample it, but add as a comment that a default integration will do it, but they still could do it manually. @bruno-garcia ?

maciejwalkowiak: Addressed in #502.

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
maciejwalkowiak: Removed in #502 as I believe we should have only working and tested code in sample. We can add it once we gain more confidence.

maciejwalkowiak:
Closing in favour of #502.

sentry-samples/sentry-samples-console/build.gradle.kts
bruno-garcia: I guess removing then is best, otherwise tbh I'd also avoid the dependency and just leave those lines of code.
If gradle changed and broken I'm sure the plugin would take more time to catch up than us just patching the fix here.

I guess it boils down to how much pain have you had because of dependencies in your life. I avoid them as much as possible

sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
bruno-garcia: I rather ppl didn't know this existed. We'll start receiving all sorts of interesting values on the server if ppl are pushed to add something here.

sentry-samples/sentry-samples-console/build.gradle.kts
marandaneto: @bruno-garcia gradle plugins are not dependencies

PR #500, opened 2020-07-29
Add method to add breadcrumb with string parameter.

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Add method to add breadcrumb with string parameter.

See https://github.com/getsentry/sentry-android/pull/499#discussion_r462614172

There's a unit test.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/SentryEvent.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryEvent.java
@@ -218,6 +218,10 @@ public final class SentryEvent implements IUnknownPropertiesConsumer {
     breadcrumbs.add(breadcrumb);
   }
 
+  public void addBreadcrumb(final @Nullable String message) {
+    this.addBreadcrumb(new Breadcrumb(message));
+  }
+
   Map<String, String> getTags() {
     return tags;
   }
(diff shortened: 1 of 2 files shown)


bruno-garcia:
Oh, what about this also going to `master`?

maciejwalkowiak:
I'm good if it goes to `master`. I took today's discussion of not breaking Android build seriously 🙂  I've just switched base branch to master.

bruno-garcia:
Makes sense to be on the safe side.

This is just an overload so not a breaking change though
For sure Android users can take advantage also, so would be nice.

I believe the main concern are things that would need to be tested on different Android version (yup) to make sure things continue to work. Because, Android.

Things like this overload IMO are safe and will make it simpler later for us to merge things from the java branch and master since a lot of code **not ok** to merge into master will follow.

@marandaneto thoughts?

marandaneto:
> Makes sense to be on the safe side.
> 
> This is just an overload so not a breaking change though
> For sure Android users can take advantage also, so would be nice.
> 
> I believe the main concern are things that would need to be tested on different Android version (yup) to make sure things continue to work. Because, Android.
> 
> Things like this overload IMO are safe and will make it simpler later for us to merge things from the java branch and master since a lot of code **not ok** to merge into master will follow.
> 
> @marandaneto thoughts?

yep, agreed, it's just an overload, things like this are LGTM

maciejwalkowiak:
Closed in favour of #501.

PR #504, opened 2020-07-30
fix(sentry-java): Contexts belong on the Scope

maciejwalkowiak:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds `Contexts` to scope together with deep cloning for known `Cloneable` objects.

Continuation of #240.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
```diff
@@ -1,13 +1,14 @@
 package io.sentry.samples.console;
 
+import java.util.Collections;
```
maciejwalkowiak, line 3: Changes in this file are unrelated to PR. It's just formatting that somehow wasn't taken into account when merging PR with sample console app.

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -278,6 +281,24 @@ public void removeExtra(@NotNull String key) {
     this.extra.remove(key);
   }
 
+  /**
+   * Returns the Scope's contexts
+   *
+   * @return the contexts
+   */
+  public @NotNull Contexts getContexts() {
```
bruno-garcia, line 289: I know I originally added this but I'm not sure it's a good idea to expose the setter/getter.
It could be needed later to have a `clear` method on the scope in which case we'd set a new instance of Contexts but that could all be internal.

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -332,7 +353,7 @@ public void removeExtra(@NotNull String key) {
 
     final Map<String, Object> extraRef = extra;
 
-    Map<String, Object> extraClone = new HashMap<>();
+    Map<String, Object> extraClone = new ConcurrentHashMap<>();
```
bruno-garcia, line 356: This is an important fix!

sentry-core/src/main/java/io/sentry/core/protocol/App.java
```diff
@@ -77,9 +80,29 @@ public void setAppBuild(String appBuild) {
...
+
+  /**
+   * Clones an App aka deep copy
+   *
+   * @return the cloned App
+   * @throws CloneNotSupportedException if object is not cloneable
+   */
+  @Override
+  public @NotNull App clone() throws CloneNotSupportedException {
+    final App clone = (App) super.clone();
```
bruno-garcia, line 102: Is this cloning via reflection?
Should we implementing the assignments by hand instead?

sentry-core/src/main/java/io/sentry/core/protocol/Browser.java
```diff
@@ -28,9 +31,29 @@ public void setVersion(String version) {
...
+   * Clones a Browser aka deep copy
+   *
+   * @return the cloned Browser
+   * @throws CloneNotSupportedException if object is not cloneable
+   */
+  @Override
+  public @NotNull Browser clone() throws CloneNotSupportedException {
+    final Browser clone = (Browser) super.clone();
+
+    clone.unknown = CollectionUtils.shallowCopy(unknown);
```
bruno-garcia, line 55: I just realized that if unknown isn't thread safe this could throw. Maybe we should guard all of these clones?

sentry-core/src/main/java/io/sentry/core/util/CollectionUtils.java
```diff
@@ -25,4 +28,28 @@ public static int size(Iterable<?> data) {
...
+   * Creates a shallow copy of map given by parameter.
+   *
+   * @param map the map to copy
+   * @param <K> the type of map keys
+   * @param <V> the type of map values
+   * @return the shallow copy of map
+   */
+  public static <K, V> @Nullable Map<K, V> shallowCopy(@Nullable Map<K, V> map) {
+    if (map != null) {
+      final Map<K, V> clone = new HashMap<>();
```
bruno-garcia, line 42: If we're cloning into non-thread-safe collections, couldn't we be cloning them later and end up crashing this for them being changed while we clone?

sentry-core/src/main/java/io/sentry/core/protocol/App.java
maciejwalkowiak: It executes shallow clone using native JVM methods - there's no reflection overhead.

sentry-core/src/main/java/io/sentry/core/Scope.java
marandaneto: not really IMO, Java guarantees that a field is thread-safe within a method, the global fields are using thread-safe map/list.
IMO this change is fine but has no effect

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
```diff
@@ -343,9 +346,35 @@ public void setBatteryTemperature(Float batteryTemperature) {
...
+  @Override
+  public @NotNull Device clone() throws CloneNotSupportedException {
+    final Device clone = (Device) super.clone();
+
+    final String[] archsRef = this.archs;
+    clone.archs = archsRef != null ? this.archs.clone() : null;
+
+    final TimeZone timezoneRef = this.timezone;
+    clone.timezone = timezoneRef != null ? (TimeZone) this.timezone.clone() : null;
+
```
marandaneto, line 375: I think you need to deep copy `orientation` here too, like:

```
    final SentryLevel levelRef = level;
    clone.level =
        levelRef != null ? SentryLevel.valueOf(levelRef.name().toUpperCase(Locale.ROOT)) : null;
```

sentry-core/src/main/java/io/sentry/core/util/CollectionUtils.java
marandaneto: I'd do something like:

`final Map<String, Object> unknownRef = unknown;`

check if `unknownRef` is null first and only use `unknownRef`

sentry-core/src/test/java/io/sentry/core/protocol/AppTest.kt
```diff
@@ -0,0 +1,57 @@
...
+        app.deviceAppHash = "device app hash"
+        val unknown = mapOf(Pair("unknown", "unknown"))
+        app.acceptUnknownProperties(unknown)
+
+        val clone = app.clone()
+
+        assertEquals("app build", clone.appBuild)
+        assertEquals("app identifier", clone.appIdentifier)
+        assertEquals("app name", clone.appName)
+        assertEquals(date.time, clone.appStartTime.time)
```
marandaneto, line 51:
```suggestion
        assertEquals(date.time, clone.appStartTime.time)
        assertNotSame(date, clone.appStartTime)
```

marandaneto: never mind, the other test does it

sentry-core/src/test/java/io/sentry/core/protocol/DeviceTest.kt
```diff
@@ -0,0 +1,104 @@
...
+    fun `cloning device wont have the same references`() {
+        val device = Device()
+        device.archs = arrayOf("archs1", "archs2")
+        device.bootTime = Date()
+        device.timezone = TimeZone.getDefault()
+        val unknown = mapOf(Pair("unknown", "unknown"))
+        device.acceptUnknownProperties(unknown)
+
+        val clone = device.clone()
+
```
marandaneto, line 21:
```suggestion
assertNotSame(device.orientation, clone.orientation)
```

sentry-core/src/main/java/io/sentry/core/Breadcrumb.java
```diff
@@ -227,20 +227,7 @@ public void acceptUnknownProperties(@Nullable Map<String, Object> unknown) {
...
-        if (item != null) {
-          unknownClone.put(item.getKey(), item.getValue()); // shallow copy
-        }
-      }
-
-      clone.unknown = unknownClone;
-    } else {
-      clone.unknown = null;
-    }
+    clone.unknown = CollectionUtils.shallowCopy(unknown);
```
marandaneto, line 230: more like a suggestion, what's about using this method and replace all the others shallow copy within clone methods? eg `Breadcrumb.data`

sentry-core/src/main/java/io/sentry/core/Scope.java
marandaneto: just in case it needs to be public, we can add `@ApiStatus.Internal` at least.
but I guess in this case package-private would be enough

marandaneto: now reading again, if we don't make this public, we won't be able to offer statically typed getters and setters like
`Device getDevice()` and `void setDevice(Device device)` on the scope, as only the Contetxts class has these methods.

or we could at least to expose:
setContexts(String key, Object value), also a getter, we'd need to solve that either way.

sentry-core/src/main/java/io/sentry/core/util/CollectionUtils.java
maciejwalkowiak: After thinking it through, `HashMap` constructor does pretty much the same and if we change implementation to `ConcurrentHashMap` we would avoid `ConcurrentModificationException` if map gets modified while other thread iterates over it.

Whole method could be replaced with:

```java
public static <K, V> @Nullable Map<K, V> shallowCopy(@Nullable Map<K, V> map) {
  if (map != null) {
    return new ConcurrentHashMap<>(map);
  } else {
    return null;
  }
}
```

maciejwalkowiak:
Another thing I have in mind if we should implement `Cloneable` interface and use `clone` method at all. If we go instead with our own custom `Cloneable<T>` interface we could let library users to implement it in their custom objects that are put to `Contexts` and we wouldn't be limited to clone only our well known objects. It's also generally [not recommended](https://www.artima.com/intv/bloch.html#part13) to use implement Java `Cloneable` interfaces.

sentry-core/src/main/java/io/sentry/core/protocol/Browser.java
maciejwalkowiak: Take a look please at https://github.com/getsentry/sentry-android/pull/504#discussion_r463582852

sentry-core/src/main/java/io/sentry/core/protocol/App.java
maciejwalkowiak: That being said I have never used `clone` method before and rather went with copy constructors for similar use cases.

sentry-core/src/main/java/io/sentry/core/util/CollectionUtils.java
marandaneto: I think the biggest problem is if `map` is set to null in a different thread, then it'd cause NPE I guess?!
so creating an aux. field reference for `map` might solve the problem.

maciejwalkowiak: The only risk I see is that `map` can be modified while `ConcurrentHashMap` is being constructed. Perhaps `shallowCopy` should only accept `ConcurrentHashMap` as a parameter and all `unknown` and other maps in objects that are meant to put into `Contexts` should be also enforced to be a type of `ConcurrentHashMap`.

marandaneto: making everything of the type `ConcurrentHashMap` has some overhead.
why not just:

```
final Map mapRef = map;
if (mapRef != null) {
    return new ConcurrentHashMap<>(mapRef);
} else {
    return null;
}
```

I believe only this would be enough

maciejwalkowiak: `map` cannot be set to `null` in other thread as it's a copy of the reference of the original reference passed to this function, right?

`ConcurrentHashMap` in the constructor iterates over `map` passed as a parameter, meaning that if during this time this map gets modified and it's a plain `HashMap` I'm afraid `ConcurrentModificationException` may be thrown.

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
maciejwalkowiak: Why is this part `.toUpperCase(Locale.ROOT)` needed?

maciejwalkowiak:
@marandaneto regarding actually using `Scope#contexts` - your and @bruno-garcia guidance is welcome here. I believe we should use it in `SentryClient#applyScope` like:

```
if (event.getContexts().isEmpty()) {
  event.setContexts(scope.getContexts());
}
```
but I might have missed something from the bigger picture.

marandaneto:
> Another thing I have in mind if we should implement `Cloneable` interface and use `clone` method at all. If we go instead with our own custom `Cloneable<T>` interface we could let library users to implement it in their custom objects that are put to `Contexts` and we wouldn't be limited to clone only our well known objects. It's also generally [not recommended](https://www.artima.com/intv/bloch.html#part13) to use implement Java `Cloneable` interfaces.

that sounds a good idea, as I said, the `Cloneable` is just a marker interface anyway, as we're cloning the list/map etc manually.
if you create a `Cloneable<T>`, we'd need to fix all the `clone` methods and copy field by field.

marandaneto:
> @marandaneto regarding actually using `Scope#contexts` - your and @bruno-garcia guidance is welcome here. I believe we should use it in `SentryClient#applyScope` like:
> 
> ```
> if (event.getContexts().isEmpty()) {
>   event.setContexts(scope.getContexts());
> }
> ```
> 
> but I might have missed something from the bigger picture.

yep, this would be moved either to SentryClient or MainEventProcessor, but a note here is, in order to read some device context, we need to have the Android Context and that's why we have this `DefaultAndroidEventProcessor`.
So right now `DefaultAndroidEventProcessor` takes an event and mutates it in there, I believe this should change now for something like DefaultAndroidEventProcessor returns a Map with the things we need from Android and then either SentryClient or MainEventProcessor set to the event respecting the precedence of the event or scope

sentry-core/src/main/java/io/sentry/core/Scope.java
bruno-garcia: My comment should have gone on the `setContexts`, the getter is fine as you said to get the typed context stuff

sentry-core/src/main/java/io/sentry/core/Scope.java
bruno-garcia: Sorry but how does Java guarantee a field is thread safe?

This Map being an instance of `HashMap` and a thread going through it's keys while another is mutating it will throw on the reader. Or will it not?

bruno-garcia:
Please be aware:

```java
if (event.getContexts().isEmpty()) {
  event.setContexts(scope.getContexts());
}
```
The values copied must be cloned. Otherwise while the event is queued for submission, the scope by be changed, affecting that event. The only references they'd share is any user-defined one, which we were not able to clone with our own deep cloning strategy (i.e Reference types in Extra and Contexts)

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -278,6 +281,24 @@ public void removeExtra(@NotNull String key) {
...
+  }
+
+  /**
+   * Sets the Scope's contexts
+   *
+   * @param contexts the contexts
+   */
+  public void setContexts(final @NotNull Contexts contexts) {
+    this.contexts = contexts;
+  }
```
bruno-garcia, line 300:
```suggestion
  public void setContexts(final @NotNull String key, @NotNull Object value) {
    this.contexts.add(key, value);
  }
  
  As per: https://github.com/getsentry/sentry-android/pull/504/files#r463443426
```

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
bruno-garcia: Enum is immutable so that's fine to use the same ref

sentry-core/src/test/java/io/sentry/core/protocol/DeviceTest.kt
maciejwalkowiak: As discussed [here](https://github.com/getsentry/sentry-android/pull/504#discussion_r463621947), these enums are immutable. Resolving.

sentry-core/src/main/java/io/sentry/core/Scope.java
maciejwalkowiak: 👍

sentry-core/src/main/java/io/sentry/core/Breadcrumb.java
maciejwalkowiak: Fixed in https://github.com/maciejwalkowiak/sentry-android/pull/1

sentry-core/src/main/java/io/sentry/core/util/CollectionUtils.java
maciejwalkowiak: Fix for that has been implemented in https://github.com/maciejwalkowiak/sentry-android/pull/1 (separate PR so that we can isolated thread-safety discussions).

maciejwalkowiak:
There's a PR to this PR: https://github.com/maciejwalkowiak/sentry-android/pull/1

sentry-core/src/main/java/io/sentry/core/Scope.java
marandaneto: a field declared within a method is not accessible outside of it, so I don't see how `extraClone` could be modified by different threads.
`extra` yes, it could, but then we have a copy ref to avoid that

sentry-core/src/main/java/io/sentry/core/protocol/Device.java
marandaneto: I remember having tests failing because it was the same ref. and recreating it using `valueOf` fixed the issue.
we had a short time to do this deep copy methods so it might be an oversight too.

maciejwalkowiak:
> > @marandaneto regarding actually using `Scope#contexts` - your and @bruno-garcia guidance is welcome here. I believe we should use it in `SentryClient#applyScope` like:
> > ```
> > if (event.getContexts().isEmpty()) {
> >   event.setContexts(scope.getContexts());
> > }
> > ```
> > 
> > 
> > but I might have missed something from the bigger picture.
> 
> yep, this would be moved either to SentryClient or MainEventProcessor, but a note here is, in order to read some device context, we need to have the Android Context and that's why we have this `DefaultAndroidEventProcessor`.
> So right now `DefaultAndroidEventProcessor` takes an event and mutates it in there, I believe this should change now for something like DefaultAndroidEventProcessor returns a Map with the things we need from Android and then either SentryClient or MainEventProcessor set to the event respecting the precedence of the event or scope

What if we add this to `SentryClient#applyScope` before preprocessors do the work:

```java
for (Map.Entry<String, Object> entry : scope.getContexts().clone().entrySet()) {
  if (!event.getContexts().containsKey(entry.getKey())) {
    event.getContexts().put(entry.getKey(), entry.getValue());
  }
}
```
`DefaultAndroidEventProcessor` will set `Device` and `Operating` system only if these properties are missing. This would mean that the contexts on `SentryEvent` take precedence over the contexts on `Scope` which takes precedence over what is applied in `DefaultAndroidEventProcessor`. What do you think about this logic @marandaneto?

bruno-garcia:
> DefaultAndroidEventProcessor will set Device and Operating system only if these properties are missing. This would mean that the contexts on SentryEvent take precedence over the contexts on Scope which takes precedence over what is applied in DefaultAndroidEventProcessor. What do you think about this logic @marandaneto?

Makes sense to me.

sentry-core/src/main/java/io/sentry/core/Scope.java
bruno-garcia: > a field declared within a method is not accessible outside of it, so I don't see how extraClone could be modified by different threads.
extra yes, it could, but then we have a copy ref to avoid that

It won't be accessed while it's in this scope (as in this method's body and not accessible from the outside.
It will though be acessed by multiple threads once the Hub is global, with this Scope object global:

A few lines below:
`     clone.extra = extraClone; `

Now you can do `addExtra` via the public API and mutate this field.
Once someone wants to clone it next time, it'll iterate over this reference, which is not thread-safe

marandaneto:
> Approving but let's see what @marandaneto has to say

its LGTM for me too, well done @maciejwalkowiak

PR #505, opened 2020-08-03
fix: converting UTC and ISO timestamp when missing Locale/TimeZone do not error

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: converting UTC and ISO timestamp if missing Locale or time zones.

Andromax G36C1H
Smartfren Andromax AD6B1H
Andromax C46B2H

devices have problems with getting US locale or UTC/GTM time zones.

I can't really test it, I don't have those devices or found emulators.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/Session.java
+++ b/sentry-core/src/main/java/io/sentry/core/Session.java
@@ -95,8 +95,8 @@ public final class Session {
       final @NotNull String release) {
     this(
         State.Ok,
-        DateUtils.getCurrentDateTime(),
-        DateUtils.getCurrentDateTime(),
+        DateUtils.getCurrentDateTimeOrNull(),
+        DateUtils.getCurrentDateTimeOrNull(),
         0,
         distinctId,
         UUID.randomUUID(),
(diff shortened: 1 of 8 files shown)


sentry-core/src/main/java/io/sentry/core/Session.java
```diff
@@ -186,11 +186,13 @@ public void end(final @Nullable Date timestamp) {
       if (timestamp != null) {
         this.timestamp = timestamp;
       } else {
-        this.timestamp = DateUtils.getCurrentDateTime();
+        this.timestamp = DateUtils.getCurrentDateTimeOrNull();
       }
 
-      duration = calculateDurationTime(this.timestamp);
-      sequence = getSequenceTimestamp(this.timestamp);
+      if (this.timestamp != null) {
```
bruno-garcia, line 192: Likely worth not having sessions enabled at all for this device since they'll be broken

marandaneto: yep, `SessionAdapter` will drop on deserialization at least.

PR #506, opened 2020-08-03
Set `SdkVersion` in default `SentryOptions` created in sentry-core module

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Sets `SdkVersion` in `SentryOptions` created in sentry-core module from build-time generated class `BuildConfig`.

https://github.com/getsentry/sentry-java/issues/873

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/buildSrc/src/main/java/Config.kt
+++ b/buildSrc/src/main/java/Config.kt
@@ -5,6 +5,8 @@ object Config {
     object BuildPlugins {
         val androidGradle = "com.android.tools.build:gradle:4.0.1"
         val kotlinGradlePlugin = "gradle-plugin"
+        val buildConfig = "com.github.gmazzo.buildconfig"
+        val buildConfigVersion = "2.0.2"
     }
 
     object Android {
--- a/buildSrc/src/main/java/Config.kt
+++ b/buildSrc/src/main/java/Config.kt
@@ -60,8 +60,8 @@ object Config {
     }
 
     object Sentry {
-        val SENTRY_CLIENT_NAME = "sentry.java"
-        val SENTRY_ANDROID_CLIENT_NAME = "$SENTRY_CLIENT_NAME.android"
+        val SENTRY_JAVA_CLIENT_NAME = "sentry.java"
+        val SENTRY_ANDROID_CLIENT_NAME = "$SENTRY_JAVA_CLIENT_NAME.android"
         val group = "io.sentry"
         val description = "SDK for sentry.io"
         val website = "https://sentry.io"
--- a/buildSrc/src/main/java/Config.kt
+++ b/buildSrc/src/main/java/Config.kt
@@ -60,8 +60,8 @@ object Config {
     }
 
     object Sentry {
-        val SENTRY_JAVA_CLIENT_NAME = "sentry.java"
-        val SENTRY_ANDROID_CLIENT_NAME = "$SENTRY_JAVA_CLIENT_NAME.android"
+        val SENTRY_JAVA_SDK_NAME = "sentry.java"
+        val SENTRY_ANDROID_SDK_NAME = "$SENTRY_JAVA_SDK_NAME.android"
         val group = "io.sentry"
         val description = "SDK for sentry.io"
         val website = "https://sentry.io"
--- a/buildSrc/src/main/java/Config.kt
+++ b/buildSrc/src/main/java/Config.kt
@@ -60,7 +60,8 @@ object Config {
     }
 
     object Sentry {
-        val SENTRY_CLIENT_NAME = "sentry.java.android"
+        val SENTRY_CLIENT_NAME = "sentry.java"
+        val SENTRY_ANDROID_CLIENT_NAME = "$SENTRY_CLIENT_NAME.android"
         val group = "io.sentry"
         val description = "SDK for sentry.io"
         val website = "https://sentry.io"
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
@@ -35,7 +35,7 @@ public final class SentryAndroidOptions extends SentryOptions {
   private boolean enableAppComponentBreadcrumbs = true;
 
   public SentryAndroidOptions() {
-    setSentryClientName(BuildConfig.SENTRY_CLIENT_NAME + "/" + BuildConfig.VERSION_NAME);
+    setSentryClientName(BuildConfig.SENTRY_ANDROID_SDK_NAME + "/" + BuildConfig.VERSION_NAME);
     setSdkVersion(createSdkVersion());
   }
 
@@ -46,7 +46,7 @@ public final class SentryAndroidOptions extends SentryOptions {
       sdkVersion = new SdkVersion();
     }
 
-    sdkVersion.setName(BuildConfig.SENTRY_CLIENT_NAME);
+    sdkVersion.setName(BuildConfig.SENTRY_ANDROID_SDK_NAME);
     String version = BuildConfig.VERSION_NAME;
     sdkVersion.setVersion(version);
     sdkVersion.addPackage("maven:sentry-android-core", version);
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
@@ -40,7 +40,11 @@ public final class SentryAndroidOptions extends SentryOptions {
   }
 
   private @NotNull SdkVersion createSdkVersion() {
-    final SdkVersion sdkVersion = getSdkVersion() != null ? getSdkVersion() : new SdkVersion();
+    SdkVersion sdkVersion = getSdkVersion();
+
+    if (sdkVersion == null) {
+      sdkVersion = new SdkVersion();
+    }
 
     sdkVersion.setName(BuildConfig.SENTRY_CLIENT_NAME);
     String version = BuildConfig.VERSION_NAME;
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
@@ -40,16 +40,12 @@ public final class SentryAndroidOptions extends SentryOptions {
   }
 
   private @NotNull SdkVersion createSdkVersion() {
-    final SdkVersion sdkVersion = new SdkVersion();
+    final SdkVersion sdkVersion = getSdkVersion() != null ? getSdkVersion() : new SdkVersion();
 
     sdkVersion.setName(BuildConfig.SENTRY_CLIENT_NAME);
     String version = BuildConfig.VERSION_NAME;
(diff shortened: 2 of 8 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -40,16 +40,12 @@ public SentryAndroidOptions() {
   }
 
   private @NotNull SdkVersion createSdkVersion() {
-    final SdkVersion sdkVersion = new SdkVersion();
+    final SdkVersion sdkVersion = getSdkVersion() != null ? getSdkVersion() : new SdkVersion();
```
bruno-garcia, line 43: Everytime I see a getter called twice I think it might be allocating on each call or otherwise have side effect. Might not be common in Java but it's hard not to wonder (or go look into it).

If it had a `getSdkVersion() ?? new SdkVersion` that would solve, but .. Java.

Is it too much hassle to assing the result of `getSdkVersion()` to test for null and assign?

sentry-core/build.gradle.kts
```diff
@@ -65,6 +66,18 @@ tasks {
     }
 }
 
+buildConfig {
+    useJavaOutput()
+    packageName("io.sentry.core")
```
bruno-garcia, line 71: It's package as in package-manager:

```suggestion
    packageName("maven:io.sentry.sentry-core")
```

bruno-garcia: Would change when we rename to `io.sentry`

sentry-core/build.gradle.kts
```diff
@@ -65,6 +66,18 @@ tasks {
...
+    buildConfigField("String", "SDK_NAME", "\"${project.name}\"")
```
bruno-garcia, line 72: project name will be what here exactly? `sentry-core?
I'm a bit lost but SDK name is `sentry.java`, `sentry.java.android`.. etc. Sentry convention

maciejwalkowiak: It would be `sentry-core`. Ok, I'll change it to `sentry.java`.

sentry-core/build.gradle.kts
maciejwalkowiak: This is a package where the plugin generates a class.

sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidOptionsTest.kt
```diff
@@ -26,12 +26,12 @@ class SentryAndroidOptionsTest {
         assertEquals(BuildConfig.VERSION_NAME, sdkVersion.version)
 
         assertTrue(sdkVersion.packages!!.any {
-            it.name == "maven:sentry-android-core"
+            it.name == "maven:sentry-android-core" &&
```
maciejwalkowiak, line 29: This was a funny bug. These tests were not testing the first condition at all but since Kotlin does allow this syntax the first like was just ignored.

On the side note - have you @bruno-garcia @marandaneto considered using AssertJ for assertions?

bruno-garcia: I'd prefer we don't have than a single assert lib in usage, so if we were to change we'd need to replace everything. Too often new libraries creep in and we end up with a mix of stuff in the code base.

@marandaneto thoughts?

sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidOptionsTest.kt
```diff
@@ -26,12 +26,12 @@ class SentryAndroidOptionsTest {
...
             it.version == BuildConfig.VERSION_NAME
         })
 
         assertTrue(sdkVersion.packages!!.any {
```
bruno-garcia, line 33: Honestly I'd have a single assert per condition here so:
```suggestion
        assertTrue(sdkVersion.packages!!.any { it.name == "" });
        assertTrue(sdkVersion.packages!!.any { it.version == "" });
```

maciejwalkowiak: If we change it, it would be possible to have a one package matching first assertion and another (or the same) package matching second one, right? This test is meant to verify if there is a single package that meets both criteria.

sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidOptionsTest.kt
maciejwalkowiak: I agree that mixing multiple assertions can lead to confusion. On the Java backend side, at least in what is considered a "modern Java backend", AssertJ has become a de-facto standard and it's been a while since I've used JUnit assertions.

sentry-core/build.gradle.kts
marandaneto: maybe the key of this field should be `SENTRY_CLIENT_NAME` so it matches `options.sentryClientName`

sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidOptionsTest.kt
marandaneto: I wanted to use `assertk` which is an inspiration kotlin version of `AssertJ`, but we're trying to reuse/migrate the tests from the old sentry-java to this project, the idea was to gain some time, obviously, it didn't work because the code became quite different

sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidOptionsTest.kt
marandaneto: yes, agree with @maciejwalkowiak

sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidOptionsTest.kt
marandaneto: good catch @maciejwalkowiak , wondering If we've done the same mistake in other places :D

buildSrc/src/main/java/Config.kt
```diff
@@ -58,7 +60,8 @@ object Config {
     }
 
     object Sentry {
-        val SENTRY_CLIENT_NAME = "sentry.java.android"
+        val SENTRY_CLIENT_NAME = "sentry.java"
```
marandaneto, line 63:
```suggestion
        val SENTRY_JAVA_CLIENT_NAME = "sentry.java"
```

sentry-core/build.gradle.kts
maciejwalkowiak: Actually I missed here setting `options.sentryClientName` - at this moment we are just setting `SdkVersion`. I will align it with Android implementation.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
maciejwalkowiak: Fixed.

sentry-core/build.gradle.kts
marandaneto: ah wait wait, I've mistaken the topic maybe too.
sentryClientName != the sdkVersion.packages.

maciejwalkowiak: I think you're right in the end. I'll add this line to `SentryOptions`.

```
setSentryClientName(BuildConfig.SENTRY_CLIENT_NAME + "/" + BuildConfig.VERSION_NAME);
```

buildSrc/src/main/java/Config.kt
maciejwalkowiak: Done.

sentry-core/build.gradle.kts
bruno-garcia: This client name goes into the outgoing HTTP header, right? That to be honest doesn't really matter. We don't use for anything except debuggin (and possibly dropping all events on a load balancer based on this value) but the `sdk name` is important (i.e: sentry.java, sentry.java.android, sentry.java.spring, etc). And packages, are the actual package in the format pkgmanager/pkgname, like maven:io.sentry, maven:io.sentry.android, nuget:Sentry.Serilog

buildSrc/src/main/java/Config.kt
```diff
@@ -58,7 +60,8 @@ object Config {
     }
 
     object Sentry {
-        val SENTRY_CLIENT_NAME = "sentry.java.android"
+        val SENTRY_JAVA_CLIENT_NAME = "sentry.java"
```
bruno-garcia, line 63: Can we name it like this to avoid confusion?
"Client" is ambiguous, "sdk name" we all understand as the following:
```suggestion
        val SENTRY_JAVA_SDK_NAME = "sentry.java"
```

buildSrc/src/main/java/Config.kt
```diff
@@ -58,7 +60,8 @@ object Config {
...
+        val SENTRY_ANDROID_CLIENT_NAME = "$SENTRY_JAVA_CLIENT_NAME.android"
```
bruno-garcia, line 64: Doesn't the string format need to be like this?
```suggestion
        val SENTRY_ANDROID_SDK_NAME = "${SENTRY_JAVA_CLIENT_NAME}.android"
```

sentry-core/src/test/java/io/sentry/core/SentryOptionsTest.kt
```diff
@@ -124,4 +124,28 @@ class SentryOptionsTest {
...
+    @Test
+    fun `init should set SdkVersion`() {
+        val sentryOptions = SentryOptions()
+        assertNotNull(sentryOptions.sdkVersion)
+        val sdkVersion = sentryOptions.sdkVersion!!
+
+        assertEquals(BuildConfig.SENTRY_CLIENT_NAME, sdkVersion.name)
+        assertEquals(BuildConfig.VERSION_NAME, sdkVersion.version)
+
+        assertTrue(sdkVersion.packages!!.any {
```
bruno-garcia, line 137: Can we have a test like this for the android package?
Would be nice to have one fro the `ndk` package too

maciejwalkowiak: We have these tests already:
- NDK - https://github.com/getsentry/sentry-android/blob/master/sentry-android-ndk/src/test/java/io/sentry/android/ndk/SentryNdkUtilTest.kt#L12
- Android - https://github.com/getsentry/sentry-android/blob/master/sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidOptionsTest.kt#L20

buildSrc/src/main/java/Config.kt
maciejwalkowiak: Done

buildSrc/src/main/java/Config.kt
maciejwalkowiak: Done

maciejwalkowiak:
@bruno-garcia take please another look if there is anything pending.

bruno-garcia:
Thanks @maciejwalkowiak !

PR #507, opened 2020-08-04
make EventProcessor.process Nullable

davemt:
Null return is used when the event should be dropped. This aligns with the current handling in `sentry-android` code and also with the Sentry unified API definition of event processors.

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

Alignment with unified API https://develop.sentry.dev/sdk/unified-api/#terminology

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/EventProcessor.java
+++ b/sentry-core/src/main/java/io/sentry/core/EventProcessor.java
@@ -3,5 +3,6 @@ package io.sentry.core;
 import org.jetbrains.annotations.Nullable;
 
 public interface EventProcessor {
+  @Nullable
   SentryEvent process(SentryEvent event, @Nullable Object hint);
 }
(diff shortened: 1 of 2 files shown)


davemt:
The in-project implementations (e.g. `MainEventProcessor`, `DefaultAndroidEventProcessor`) happen to not be able to return null and so can technically be marked `NotNull`, even with the `Nullable` allowance specified on the interface. Would it be better to mark them as `Nullable` to match interface regardless?

bruno-garcia:
> Would it be better to mark them as Nullable to match interface regardless?

You're right, it's best to mark them explicitly to what they do there: `@NotNull`. The interface on the other hand is `@Nullable`.

The code in the client already behaves properly, so it's just a matter of annotation:

https://github.com/getsentry/sentry-android/blob/e8b5549bf2443c70e1a1e8579decbc3a114e426c/sentry-core/src/main/java/io/sentry/core/SentryClient.java#L81-L93

PR #511, opened 2020-08-11
Add Logback integration.

maciejwalkowiak:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Added SentryAppender that converts Logback events into Sentry events and sends them to Sentry.

I added Sample project using Logback, there is also an integration test. On the side I have a POC of Sentry + Logback + Spring Boot integration - https://github.com/maciejwalkowiak/sentry-spring-boot-demo

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -1,5 +1,8 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
 plugins {
     `kotlin-dsl`
+    kotlin("jvm") version "1.3.72"
 }
 
 repositories {
@@ -9,4 +12,15 @@ repositories {
 // Disabling the warning about the use of experimental Kotlin compiler features
 kotlinDslPluginOptions {
     experimentalWarning.set(false)
+}
+dependencies {
+    implementation(kotlin("stdlib-jdk8"))
+}
+val compileKotlin: KotlinCompile by tasks
+compileKotlin.kotlinOptions {
+    jvmTarget = "1.8"
+}
+val compileTestKotlin: KotlinCompile by tasks
+compileTestKotlin.kotlinOptions {
+    jvmTarget = "1.8"
 }
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -1,8 +1,5 @@
-import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
-
 plugins {
     `kotlin-dsl`
-    kotlin("jvm") version "1.3.72"
 }
 
 repositories {
@@ -12,15 +9,4 @@ repositories {
 // Disabling the warning about the use of experimental Kotlin compiler features
 kotlinDslPluginOptions {
     experimentalWarning.set(false)
-}
-dependencies {
-    implementation(kotlin("stdlib-jdk8"))
-}
-val compileKotlin: KotlinCompile by tasks
-compileKotlin.kotlinOptions {
-    jvmTarget = "1.8"
-}
-val compileTestKotlin: KotlinCompile by tasks
-compileTestKotlin.kotlinOptions {
-    jvmTarget = "1.8"
 }
--- a/sentry-core/src/main/java/io/sentry/core/SentryEvent.java
+++ b/sentry-core/src/main/java/io/sentry/core/SentryEvent.java
@@ -67,7 +67,8 @@ public final class SentryEvent implements IUnknownPropertiesConsumer {
   }
 
   @Nullable
-  Throwable getThrowable() {
+  @ApiStatus.Internal
+  public Throwable getThrowable() {
     return throwable;
   }
 
--- a/sentry-logback/build.gradle.kts
+++ b/sentry-logback/build.gradle.kts
@@ -1,0 +1,93 @@
+import com.novoda.gradle.release.PublishExtension
+
+plugins {
+    `java-library`
+    kotlin("jvm")
+    jacoco
+    id(Config.QualityPlugins.errorProne)
+    id(Config.Deploy.novodaBintray)
+    id(Config.QualityPlugins.gradleVersions)
+}
+
+configure<JavaPluginConvention> {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+    kotlinOptions.jvmTarget = JavaVersion.VERSION_1_8.toString()
+}
+
+dependencies {
+    implementation(project(":sentry-core"))
+    // Envelopes require JSON. Until a parse is done without GSON, we'll depend on it explicitly here
+    implementation(Config.Libs.gson)
+    implementation(Config.Libs.logbackCore)
+    implementation(Config.Libs.logbackClassic)
+
+    compileOnly(Config.CompileOnly.nopen)
+    errorprone(Config.CompileOnly.nopenChecker)
+    errorprone(Config.CompileOnly.errorprone)
+    errorproneJavac(Config.CompileOnly.errorProneJavac8)
+    compileOnly(Config.CompileOnly.jetbrainsAnnotations)
+
+    // tests
+    testImplementation(kotlin(Config.kotlinStdLib))
+    testImplementation(Config.TestLibs.kotlinTestJunit)
+    testImplementation(Config.TestLibs.mockitoKotlin)
+    testImplementation(Config.TestLibs.awaitility)
+}
+
+configure<SourceSetContainer> {
+    test {
+        java.srcDir("src/test/java")
+    }
+}
+
+jacoco {
+    toolVersion = Config.QualityPlugins.jacocoVersion
+}
+
+tasks.jacocoTestReport {
+    reports {
+        xml.isEnabled = true
+        html.isEnabled = false
+    }
+}
+
+tasks {
+    jacocoTestCoverageVerification {
+        violationRules {
+            // TODO: Raise the minimum to a sensible value.
+            rule { limit { minimum = BigDecimal.valueOf(0.1) } }
+        }
+    }
+    check {
+        dependsOn(jacocoTestCoverageVerification)
+        dependsOn(jacocoTestReport)
+    }
+}
+
+//TODO: move these blocks to parent gradle file, DRY
+configure<PublishExtension> {
+    userOrg = Config.Sentry.userOrg
+    groupId = project.group.toString()
+    publishVersion = project.version.toString()
+    desc = Config.Sentry.description
+    website = Config.Sentry.website
+    repoName = Config.Sentry.repoName
+    setLicences(Config.Sentry.licence)
+    setLicenceUrls(Config.Sentry.licenceUrl)
+    issueTracker = Config.Sentry.issueTracker
+    repository = Config.Sentry.repository
+    sign = Config.Deploy.sign
+    artifactId = project.name
+    uploadName = "${project.group}:${project.name}"
+    devId = Config.Sentry.userOrg
+    devName = Config.Sentry.devName
+    devEmail = Config.Sentry.devEmail
+    scmConnection = Config.Sentry.scmConnection
+    scmDevConnection = Config.Sentry.scmDevConnection
+    scmUrl  = Config.Sentry.scmUrl
+}
+
--- a/sentry-logback/build.gradle.kts
+++ b/sentry-logback/build.gradle.kts
@@ -20,7 +20,7 @@ tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach 
 }
 
 dependencies {
-    implementation(project(":sentry-core"))
+    api(project(":sentry-core"))
     implementation(Config.Libs.logbackClassic)
 
     compileOnly(Config.CompileOnly.nopen)
@@ -56,8 +56,7 @@ tasks.jacocoTestReport {
 tasks {
     jacocoTestCoverageVerification {
         violationRules {
-            // TODO: Raise the minimum to a sensible value.
-            rule { limit { minimum = BigDecimal.valueOf(0.1) } }
+            rule { limit { minimum = BigDecimal.valueOf(0.6) } }
         }
     }
     check {
--- a/sentry-logback/build.gradle.kts
+++ b/sentry-logback/build.gradle.kts
@@ -7,6 +7,7 @@ plugins {
     id(Config.QualityPlugins.errorProne)
     id(Config.Deploy.novodaBintray)
     id(Config.QualityPlugins.gradleVersions)
+    id(Config.BuildPlugins.buildConfig) version Config.BuildPlugins.buildConfigVersion
 }
 
 configure<JavaPluginConvention> {
@@ -20,9 +21,6 @@ tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach 
 
 dependencies {
     implementation(project(":sentry-core"))
-    // Envelopes require JSON. Until a parse is done without GSON, we'll depend on it explicitly here
-    implementation(Config.Libs.gson)
-    implementation(Config.Libs.logbackCore)
     implementation(Config.Libs.logbackClassic)
 
     compileOnly(Config.CompileOnly.nopen)
@@ -66,6 +64,18 @@ tasks {
         dependsOn(jacocoTestCoverageVerification)
         dependsOn(jacocoTestReport)
     }
+}
+
+buildConfig {
+    useJavaOutput()
+    packageName("io.sentry.logback")
+    buildConfigField("String", "SENTRY_LOGBACK_SDK_NAME", "\"${Config.Sentry.SENTRY_LOGBACK_SDK_NAME}\"")
+    buildConfigField("String", "VERSION_NAME", "\"${project.version}\"")
+}
+
+val generateBuildConfig by tasks
+tasks.withType<JavaCompile>().configureEach {
+    dependsOn(generateBuildConfig)
 }
 
 //TODO: move these blocks to parent gradle file, DRY
--- a/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
+++ b/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
@@ -4,6 +4,7 @@ import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.classic.spi.ThrowableProxy;
 import ch.qos.logback.core.UnsynchronizedAppenderBase;
+import io.sentry.core.DateUtils;
 import io.sentry.core.Sentry;
 import io.sentry.core.SentryEvent;
 import io.sentry.core.SentryLevel;
@@ -77,7 +78,7 @@ public final class SentryAppender extends UnsynchronizedAppenderBase<ILoggingEve
    */
   @SuppressWarnings("JdkObsolete")
   final @NotNull SentryEvent createEvent(@NotNull ILoggingEvent loggingEvent) {
-    final SentryEvent event = new SentryEvent(new Date(loggingEvent.getTimeStamp()));
+    final SentryEvent event = new SentryEvent(DateUtils.getDateTime(new Date(loggingEvent.getTimeStamp())));
     final Message message = new Message();
     message.setMessage(loggingEvent.getMessage());
     message.setFormatted(loggingEvent.getFormattedMessage());
--- a/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
+++ b/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
@@ -7,7 +7,9 @@ import ch.qos.logback.core.UnsynchronizedAppenderBase;
 import io.sentry.core.Sentry;
 import io.sentry.core.SentryEvent;
 import io.sentry.core.SentryLevel;
+import io.sentry.core.SentryOptions;
 import io.sentry.core.protocol.Message;
+import io.sentry.core.protocol.SdkVersion;
 import java.util.*;
 import java.util.stream.Collectors;
 import org.jetbrains.annotations.NotNull;
@@ -43,6 +45,8 @@ public final class SentryAppender extends UnsynchronizedAppenderBase<ILoggingEve
             Optional.ofNullable(debug).ifPresent(options::setDebug);
             Optional.ofNullable(attachThreads).ifPresent(options::setAttachThreads);
             Optional.ofNullable(attachStacktrace).ifPresent(options::setAttachStacktrace);
+            options.setSentryClientName(BuildConfig.SENTRY_LOGBACK_SDK_NAME);
+            options.setSdkVersion(createSdkVersion(options));
           });
     }
     super.start();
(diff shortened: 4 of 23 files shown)


sentry-logback/build.gradle.kts
```diff
@@ -0,0 +1,93 @@
...
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+    kotlinOptions.jvmTarget = JavaVersion.VERSION_1_8.toString()
+}
+
+dependencies {
+    implementation(project(":sentry-core"))
```
maciejwalkowiak, line 22: @marandaneto I am not 100% sure about scope of these dependencies.

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -0,0 +1,164 @@
...
+  private @Nullable Boolean debug;
+  private @Nullable Boolean attachThreads;
+  private @Nullable Boolean attachStacktrace;
+
+  @Override
+  public void start() {
+    if (dsn != null) {
+      Sentry.init(
+          options -> {
+            options.setDsn(dsn);
```
maciejwalkowiak, line 35: @marandaneto @bruno-garcia should we set here custom `sentryClientName` or do we stick to one set by `sentry-core`?

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -0,0 +1,164 @@
...
+  private @Nullable Integer readTimeoutMillis;
+  private @Nullable Double sampleRate;
+  private @Nullable Boolean bypassSecurity;
+  private @Nullable Boolean debug;
+  private @Nullable Boolean attachThreads;
+  private @Nullable Boolean attachStacktrace;
+
+  @Override
+  public void start() {
+    if (dsn != null) {
```
bruno-garcia, line 32: We need some config here to allow a user to opt out for initializing thought the sdk.

It's confusing when the 3 integrations are on and folks don't know what is initing it.

If we can also check if the sdk is already initializing and then warn about it. (Like Sentry.isEnabled) would be best. Last init wins is OK so dispose previous bound hub l/Clientand get it init again ur it's often a mistake.

bruno-garcia:
BTW super excited to see this landing!! Thanks a lot

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
maciejwalkowiak: Passing config here is optional. If user has at the same time Spring Boot starter.and configures Sentry through it he can omit completely passing options through Logback

buildSrc/build.gradle.kts
```diff
@@ -10,3 +13,14 @@ repositories {
 kotlinDslPluginOptions {
     experimentalWarning.set(false)
 }
+dependencies {
+    implementation(kotlin("stdlib-jdk8"))
```
marandaneto, line 17:
```suggestion
    implementation(kotlin(Config.kotlinStdLib, KotlinCompilerVersion.VERSION))
```

buildSrc/build.gradle.kts
```diff
@@ -10,3 +13,14 @@ repositories {
...
+}
+val compileKotlin: KotlinCompile by tasks
+compileKotlin.kotlinOptions {
+    jvmTarget = "1.8"
+}
```
marandaneto, line 22: might be possible like:

```
    kotlinOptions {
        jvmTarget = JavaVersion.VERSION_1_8.toString()
    }
```

buildSrc/build.gradle.kts
```diff
@@ -10,3 +13,14 @@ repositories {
...
+val compileTestKotlin: KotlinCompile by tasks
```
marandaneto, line 23: same as above, but using compileTestKotlin

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
```diff
@@ -67,7 +67,8 @@ public Date getTimestamp() {
   }
 
   @Nullable
-  Throwable getThrowable() {
+  @ApiStatus.Internal
+  public Throwable getThrowable() {
```
marandaneto, line 71: I remember @bruno-garcia being against this, but if it's necessary, I'm fine with it.

sentry-logback/build.gradle.kts
marandaneto: the timber integration has it as an `api` instead of `implementation` because one may configure events and breadcrumbs min level, so the client needs access to sentry-core classes, if in this case, they dont need to access any of the Sentry classes that are bundled in sentry-core, then `implementation` is fine

sentry-logback/build.gradle.kts
```diff
@@ -0,0 +1,93 @@
...
+    // Envelopes require JSON. Until a parse is done without GSON, we'll depend on it explicitly here
+    implementation(Config.Libs.gson)
```
marandaneto, line 24: ok but sentry-core already has it, do you need to access gson classes explicitly here?

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
marandaneto: yes I agree, having the granularity data would be nice

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -0,0 +1,164 @@
...
+      Sentry.init(
+          options -> {
+            options.setDsn(dsn);
+            Optional.ofNullable(maxBreadcrumbs).ifPresent(options::setMaxBreadcrumbs);
```
marandaneto, line 36: Java looks way nicer when you can use newer APIs :)

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -0,0 +1,164 @@
...
+          });
+    }
+    super.start();
+  }
+
+  @Override
+  protected void append(@NotNull ILoggingEvent eventObject) {
+    if (Sentry.isEnabled()) {
+      Sentry.captureEvent(createEvent(eventObject));
+    }
```
marandaneto, line 55: Sentry already checks internally if it's enabled`, I'd avoid this check

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -0,0 +1,164 @@
...
+      return SentryLevel.ERROR;
+    } else if (level.isGreaterOrEqual(Level.WARN)) {
+      return SentryLevel.WARNING;
+    } else if (level.isGreaterOrEqual(Level.INFO)) {
+      return SentryLevel.INFO;
+    } else if (level.isGreaterOrEqual(Level.ALL)) {
+      return SentryLevel.DEBUG;
+    } else {
+      return SentryLevel.LOG;
+    }
```
marandaneto, line 118: a `switch` here would not be cheaper? but fine with it.

I'd avoid `SentryLevel.LOG` at all, this is just for compatibility with sentry-react-native, so in case you need to fallback, do it with `DEBUG`

marandaneto:
> BTW super excited to see this landing!! Thanks a lot

yep, left a few comments but pretty excited as well, thanks!

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
maciejwalkowiak: Otherwise I am not able to test it without reflection.

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -0,0 +1,164 @@
...
+    if (throwableInformation != null) {
+      event.setThrowable(throwableInformation.getThrowable());
+    }
+
+    if (loggingEvent.getThreadName() != null) {
+      event.setExtra("thread_name", loggingEvent.getThreadName());
+    }
+
+    for (Map.Entry<String, String> entry : loggingEvent.getMDCPropertyMap().entrySet()) {
+      event.setTag(entry.getKey(), entry.getValue());
```
marandaneto, line 84: just wondering if we should do `setTag(x, y)` or `setExtras(x, y)` or `event.setContexts.out(x, y)`, because `getMDCPropertyMap` means `map diagnostic context`, the old version add as a tag so maybe we should keep as it is to keep compatibility

buildSrc/build.gradle.kts
maciejwalkowiak: All changes in this file were added by Intellij and I did not mean to push them. Sorry!

sentry-logback/build.gradle.kts
maciejwalkowiak: You're right, removed.

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
maciejwalkowiak: Done!

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
maciejwalkowiak: I think you mean that old version adds as `extras`? I see that in old SDK there was an option to tell Sentry which MDC tags qualify as `tag` and which as `extra`. In 2.0 it's not there anymore so we can either add it or pick `tags` or `extras` I think

marandaneto: yeah its trick: https://github.com/getsentry/sentry-java/blob/master/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java#L172-L178
I'd suggest then to use `setExtras` instead of `setTag`, wdyt?
I don't think we need this configuration to choose when its a tag or not, at least not in the 1st version.

sentry-logback/src/test/kotlin/io/sentry/logback/SentryAppenderTest.kt
```diff
@@ -0,0 +1,148 @@
...
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.slf4j.MDC
+
+class SentryAppenderTest {
+    private class Fixture {
+        val client = mock<ISentryClient>()
+        val captor = argumentCaptor<SentryEvent>()
```
marandaneto, line 23: not really against using `argumentCaptor`, just out of curiosity, is there any advantage of using it or the normal `argWhere`, `argThat`, `argForWhich` or even `check`? they look more natural to me when asserting values of a `verify` call.

https://github.com/nhaarman/mockito-kotlin/wiki/Mocking-and-verifying

maciejwalkowiak: Thanks for hint, I missed the existence of `check`

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
maciejwalkowiak: What's exactly the difference in practice between tags and extras? I see that they are presented differently and users can filter issues by tags. Extras are just metadata attached to an event but users can't do anything with it. Am I correct?

maciejwalkowiak:
> BTW super excited to see this landing!! Thanks a lot

Thanks @bruno-garcia. I'm not 100% happy with tests but since Sentry is heavy on static methods I didn't find a better way to test it (other than using newer version of Mockito that allows mocking static methods).

I'd appreciate your second look and in meantime I'll continue with other integrations.

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
marandaneto: yes, exactly, and looks like getMDCPropertyMap returns only metadata as well

maciejwalkowiak: But without the ability to search by this metadata its usage is very limited. I imagine users may want to put a `userId` to MDC and expect to search events for specific user.

Perhaps it makes sense to add a property to `SentryAppender` `mdcTags` - similar to get the same functionality as in Sentry SDK 1.x without cluttering `sentry-core`?

marandaneto: I get your point, for `user` for example, our protocol has a dedicated field, but with this integration, it'd be hard to know which key is user-related.
I'd say we can start with everything being an extra and improve from there when we get feedback.
@bruno-garcia opinions?

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
maciejwalkowiak: `LOG` removed. I just realised that `level` here is not an enum, we cannot make a switch.

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
maciejwalkowiak: Ok I set it as extras 👍

sentry-core/src/main/java/io/sentry/core/SentryEvent.java
bruno-garcia: we can't escape having this but the `@Internal` should do

sentry-logback/build.gradle.kts
bruno-garcia: Wouldn't we document that users using logback can add only this package and get the whole `Sentry.X` API?
Otherwise users would need to make sure they bump both `sentry-core` and `sentry-logback` while keeping them compatible.

I believe it would make sense to have all the transient package's API show up. So `api` like the `timber` makes sense?

sentry-logback/build.gradle.kts
```diff
@@ -0,0 +1,103 @@
...
+        xml.isEnabled = true
+        html.isEnabled = false
+    }
+}
+
+tasks {
+    jacocoTestCoverageVerification {
+        violationRules {
+            // TODO: Raise the minimum to a sensible value.
+            rule { limit { minimum = BigDecimal.valueOf(0.1) } }
```
bruno-garcia, line 60: This breaks the build if less than 10% code coverage.
Can we have this to 60% or so and remove the TODO already?

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
bruno-garcia: I see. Though there's no way to avoid initializing the SDK via this integration, right?
So if a user adds the `dsn` to the appender configuration file, it will init the SDK even if it was already initialized.

That's why the suggestion:

> If we can also check if the sdk is already initializing and then warn about it. (Like Sentry.isEnabled) would be best. Last init wins is OK so dispose previous bound hub l/Clientand get it init again ur it's often a mistake.

Or at least log a warning with the diagnostic logger

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -0,0 +1,190 @@
...
+    hubAdapter.captureEvent(createEvent(eventObject));
+  }
+
+  /**
+   * Creates {@link SentryEvent} from Logback's {@link ILoggingEvent}.
+   *
+   * @param loggingEvent the logback event
+   * @return the sentry event
+   */
+  @SuppressWarnings("JdkObsolete")
```
bruno-garcia, line 77: Why do we need this suppress? Maybe a note will help folks reading the code
```suggestion
  // Ignoring the obsolete because .... ?
  @SuppressWarnings("JdkObsolete")
```

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -0,0 +1,190 @@
...
+  final @NotNull SentryEvent createEvent(@NotNull ILoggingEvent loggingEvent) {
+    final SentryEvent event = new SentryEvent(new Date(loggingEvent.getTimeStamp()));
+    final Message message = new Message();
+    message.setFormatted(loggingEvent.getFormattedMessage());
```
bruno-garcia, line 81: Do we also have the template? That's what wil be used when grouping if we provide it:

```suggestion
    message.setMessage(loggingEvent.getTemplate());
```

The formatted message is used to display to the user since the actual params are already interpolated

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -0,0 +1,190 @@
...
+    event.setMessage(message);
+    event.setLogger(loggingEvent.getLoggerName());
+    event.setLevel(formatLevel(loggingEvent.getLevel()));
+
+    final ThrowableProxy throwableInformation = (ThrowableProxy) loggingEvent.getThrowableProxy();
+    if (throwableInformation != null) {
+      event.setThrowable(throwableInformation.getThrowable());
+    }
+
+    if (loggingEvent.getThreadName() != null) {
```
bruno-garcia, line 92: We actually have a `threads` model which takes a name. Maybe we can add it here though without stacktrace isn't that useful.

We tend to avoid using `extra`. The reason is that it has no "meaning". We'd prefer using `contexts` and setting the key to something more meaningful. 
If not using threads model and thread-name, this being a one-off value, it's OK to leave it on Extra though

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -0,0 +1,190 @@
...
+      event.setExtra("thread_name", loggingEvent.getThreadName());
+    }
+
+    for (Map.Entry<String, String> entry : loggingEvent.getMDCPropertyMap().entrySet()) {
```
bruno-garcia, line 96: This is a good candidate to go on `contexts` like:

`event.setContexts("MDC", ....);`

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  /**
+   * Transforms a {@link Level} into an {@link SentryLevel}.
+   *
+   * @param level original level as defined in log4j.
+   * @return log level used within sentry.
+   */
+  private static @NotNull SentryLevel formatLevel(@NotNull Level level) {
+    if (level.isGreaterOrEqual(Level.ERROR)) {
```
bruno-garcia, line 121: Does `Level` include `fatal` or `critical`? Looks like we'd map Sentry `fatal` to `error` here

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -0,0 +1,190 @@
...
+    SdkVersion sdkVersion = sentryOptions.getSdkVersion();
+
+    if (sdkVersion == null) {
+      sdkVersion = new SdkVersion();
+    }
+
+    sdkVersion.setName(BuildConfig.SENTRY_LOGBACK_SDK_NAME);
+    final String version = BuildConfig.VERSION_NAME;
+    sdkVersion.setVersion(version);
+    sdkVersion.addPackage("maven:sentry-logback", version);
```
bruno-garcia, line 142: If we're using the reference from `SentryOptions`, wouldn't we be adding a new entry on each event that goes out?
Looks like we need a clone here

sentry-logback/src/test/kotlin/io/sentry/logback/SentryAppenderTest.kt
```diff
@@ -0,0 +1,123 @@
...
+            appender.start()
+            loggerContext.start()
+        }
+    }
+
+    private val fixture = Fixture()
+
+    @Test
+    fun `converts message`() {
+        fixture.logger.debug("testing message conversion", 1, 2)
```
bruno-garcia, line 41: Would be nice to have some templates like (guessing the format): 

```suggestion
        fixture.logger.debug("testing message conversion %d %d", 1, 2)
```

Then we can assert the `message` (template) and the `formatted` field

sentry-logback/src/test/kotlin/io/sentry/logback/SentryAppenderTest.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    @Test
+    fun `attaches throwable`() {
+        val throwable = RuntimeException("something went wrong")
+        fixture.logger.warn("testing throwable", throwable)
+
+        verify(fixture.hubAdapter).captureEvent(check {
+            assertEquals(throwable, it.throwable)
+        })
+    }
```
bruno-garcia, line 122: Can we test the SDK info stuff? Would be nice to check it has the packages and name correctly there.

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
maciejwalkowiak: Because we're using `java.util.Date` and error prone forbids using this class. Ideally we should avoid using this class but I believe we do use it because newer Date classes are not supported on Android.

maciejwalkowiak: > Or at least log a warning with the diagnostic logger

We cannot access diagnostic logger from this level as it's a part of `SentryOptions`. Perhaps we should add this check `isSentryEnabled` and log warning inside `Sentry.init()` itself?

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
maciejwalkowiak: In `slf4j` there are only: error, warning, info, debug, trace. We're mapping from `slf4j` to Sentry - meaning we just won't ever get fatal level from Logback integration.

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
maciejwalkowiak: This method is invoked only once during initialization.

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
maciejwalkowiak: That's cool, I didn't know how grouping works.

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
maciejwalkowiak: Ok, got the point. The question is - do we drop it and rely on `attachThreads` or add `Thread` section to `contexts`. To be honest I do see value in having a thread name, but I don't see value for typical backend application to attach list of all threads. Especially when there can be 100 http threads running at once. 

My preference would be to add `Thread` to `contexts` and attach there fields: `id`, `name`.

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
bruno-garcia: re: Obsolete. That description you gave here would fit well as a comment. So it's clear why we ignored it.

re: logger. Sounds good to at least log on sentry.init!

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
bruno-garcia: what about we add a single thread to the list then? with id, name? When the attachThreads is false that is

PR #514, opened 2020-08-13
feat: session updates along with the event

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
feat: session updates along with the event

we'd like to send a session update along with events in a single envelope

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes


sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -84,6 +84,7 @@ static void applyMetadata(
         options.setEnableSessionTracking(sessionTrackingEnabled);
 
         if (options.getSampleRate() == null) {
+          // TODO: it needs to read a Float I guess
```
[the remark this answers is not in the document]
marandaneto: not because of that, `ManifestMetadataReader` is using `getDouble` and it should be `getFloat` if I am not mistaken or it will read the default value

sentry-core/src/main/java/io/sentry/core/OutboxSender.java
```diff
@@ -0,0 +1,176 @@
...
+                && !envelope.getHeader().getEventId().equals(event.getEventId())) {
+              logger.log(
+                  SentryLevel.ERROR,
+                  "Item %d of has a different event id (%s) to the envelope header (%s)",
+                  items,
+                  envelope.getHeader().getEventId(),
+                  event.getEventId());
+              continue;
+            }
+            hub.captureEvent(event, hint);
```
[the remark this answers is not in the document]
marandaneto: this is the outbox, the events in the outbox are raw eg NDK or RN and need to be enriched, so the envelope needs to be open and events need to be processed, `captureEvent` is the way to go or the `captureEnvelope` would need to wrap all the items and wrap it as well, as the outbox only caches 1 event per envelope and no other type, it made sense to keep as it is

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -88,43 +88,83 @@ public SentryClient(final @NotNull SentryOptions options, @Nullable Connection c
...
+      final SentryEnvelope envelope = buildEnvelope(event, session);
+
+      if (envelope != null) {
+        connection.send(envelope, hint);
+      }
     } catch (IOException e) {
-      options
-          .getLogger()
-          .log(SentryLevel.WARNING, "Capturing event " + event.getEventId() + " failed.", e);
+      options.getLogger().log(SentryLevel.WARNING, "Capturing event " + sentryId + " failed.", e);
```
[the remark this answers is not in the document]
marandaneto: we could use the string args and %s

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -135,8 +175,11 @@ public SentryClient(final @NotNull SentryOptions options, @Nullable Connection c
    * @param scope the Scope or null
    */
   @TestOnly
-  void updateSessionData(
+  @Nullable
+  Session updateSessionData(
       final @NotNull SentryEvent event, final @Nullable Object hint, final @Nullable Scope scope) {
+    Session clonedSession = null;
+
     if (ApplyScopeUtils.shouldApplyScopeData(hint)) {
```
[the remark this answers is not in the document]
marandaneto: `applyScope` already checks for that

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -88,43 +88,83 @@ public SentryClient(final @NotNull SentryOptions options, @Nullable Connection c
...
-      return SentryId.EMPTY_ID;
+      if (!sample()) {
+        options
+            .getLogger()
+            .log(
+                SentryLevel.DEBUG,
+                "Event %s was dropped due to sampling decision.",
+                event.getEventId());
+        sendEvent = false;
+        event = null;
```
bruno-garcia, line 109: If I'm not missing something, if we're setting this to null here, we'll continue the execution and attempt to send the event.

Shouldn't we just return SentryId.Empty here?

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -63,14 +60,17 @@ public SentryClient(final @NotNull SentryOptions options, @Nullable Connection c
...
 
+    boolean sendEvent = true;
+
     if (ApplyScopeUtils.shouldApplyScopeData(hint)) {
       // Event has already passed through here before it was cached
       // Going through again could be reading data that is no longer relevant
       // i.e proguard id, app version, threads
       event = applyScope(event, scope, hint);
 
       if (event == null) {
```
bruno-garcia, line 71: Could the event be null after calling `applyScope`? 
Shouldn't we check at the start of the method, then only on beforeSend, and event processors?

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -88,43 +88,83 @@ public SentryClient(final @NotNull SentryOptions options, @Nullable Connection c
...
+                SentryLevel.DEBUG,
+                "Event %s was dropped due to sampling decision.",
+                event.getEventId());
+        sendEvent = false;
+        event = null;
+      }
     }
 
-    event = executeBeforeSend(event, hint);
+    if (event != null) {
```
bruno-garcia, line 113: I find that the combination of the boolean `sendEvent` and the multiple null check is too confusing. I found it hard to follow to make sure this should work in all cases.

What about we just guard at the start for null (and return), then at the two locations where user callbacks can set it to null?

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -88,43 +88,83 @@ public SentryClient(final @NotNull SentryOptions options, @Nullable Connection c
...
 
+    SentryId sentryId = SentryId.EMPTY_ID;
+
     try {
-      connection.send(event, hint);
+      final SentryEnvelope envelope = buildEnvelope(event, session);
+
+      if (envelope != null) {
+        connection.send(envelope, hint);
+      }
```
bruno-garcia, line 129: When would envelope be null here?
Should we log this out?

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -160,28 +203,22 @@ void updateSessionData(
...
-                    // otherwise we just cache in the disk but do not flush to the network.
-                    sessionHint = new SessionUpdateHint();
                   }
-                  captureSession(session, sessionHint);
                 }
               } else {
                 options.getLogger().log(SentryLevel.INFO, "Session is null on scope.withSession");
               }
             });
+        clonedSession = scope.cloneSession();
```
bruno-garcia, line 216: We're cloning outside of the lock so we could be cloning a different session than the one updated above. For example if a thread closing a session was waiting on a lock to close the session and takes the lock before we call `clone` here.

Maybe we clone it inside the callback and assign the reference to a variable declared outside of the block. Ugly but this is guaranteed to clone the right state.

The method on the session (like `update`) could return the clone on each operation already to avoid having that public clone that also takes a lock

sentry-core/src/main/java/io/sentry/core/SentryClient.java
marandaneto: the event won't be sent, the flag is false and the event is now null, the session needs to be sent though, so the flow needs to continue

sentry-core/src/main/java/io/sentry/core/SentryClient.java
marandaneto: the event can be null when running the scope's processor

sentry-core/src/main/java/io/sentry/core/SentryClient.java
marandaneto: i will look to get rid of one of them, but i had a reason to not do so, just dont remember now, tomorrow i look into it

sentry-core/src/main/java/io/sentry/core/SentryClient.java
marandaneto: if the event was discarded and there is no session to be sent,  the `buildEnvelope` would return null, as there's nothing to be sent

sentry-core/src/main/java/io/sentry/core/SentryClient.java
marandaneto: good catch, I'll need to use an `AtomicReference` maybe because it's a lambda and I cannot assign the value to not even a final object outside of the lambda or to refactor a little bit this, i will have a look

sentry-core/src/main/java/io/sentry/core/SentryClient.java
bruno-garcia: Got it. thanks

sentry-core/src/main/java/io/sentry/core/Scope.java
```diff
@@ -392,11 +392,19 @@ public void addEventProcessor(@NotNull EventProcessor eventProcessor) {
...
+  Session withSession(@NotNull IWithSession sessionCallback) {
+    Session cloneSession = null;
     synchronized (sessionLock) {
       sessionCallback.accept(session);
+
+      if (session != null) {
+        cloneSession = session.clone();
+      }
     }
+    return cloneSession;
```
bruno-garcia, line 407: L: We could just return from within the synchronized block and avoid the variable

PR #517, opened 2020-08-14
Sentry Spring Boot Starter.

maciejwalkowiak:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Sentry Spring Boot Starter providing configuration via Spring Boot `@ConfigurationProperties`, automatically setting HTTP request details on `SentryEvent#request`, setting user details on `SentryEvent#user`.

In conjunction with `sentry-logback` enables users to utilise Sentry with close to zero extra effort.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Currently we have only Spring MVC integration (and indirectly Spring Security through getting user data from `Principal`). I think this PR should not contain anything more. 
We can consider other integrations as separate tasks & PRs.


maciejwalkowiak:
I figured out how to set up annotation processing 🎉

sentry-spring-boot-starter/build.gradle.kts
```diff
@@ -0,0 +1,117 @@
...
+        xml.isEnabled = true
+        html.isEnabled = false
+    }
+}
+
+tasks {
+    jacocoTestCoverageVerification {
+        violationRules {
+            // TODO: Raise the minimum to a sensible value.
+            rule { limit { minimum = BigDecimal.valueOf(0.1) } }
```
bruno-garcia, line 74: This is here since the first few commits to this repo.
Can we remove the TODO ad add 0.6 at least?
I belive the coverage is already at 60%

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+        SentryProperties properties,
+        ObjectProvider<SentryClient> sentryClient) {
+      Sentry.init(
+          options -> {
+            properties.applyTo(options);
+            optionsConfiguration.configure(options);
+            options.setSentryClientName(BuildConfig.SENTRY_SPRING_BOOT_SDK_NAME);
+            options.setSdkVersion(createSdkVersion(options));
+          });
+      sentryClient.ifAvailable(Sentry::bindClient);
```
bruno-garcia, line 67: Unsure about this. The options passed to `init` is used to initialize the client.
What is the use case to take both options and client in here?

Also, binding the client after calling init has some down sides (things we can do during initialization that require a client already, created in the process of init).

If letting the client com from outside will be a must, I suggest we introduce a `ClientFactory` in the options and hook here so that the SDK under the hood (through init) calls that instead of binding after the fact.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryProperties.java
```diff
@@ -0,0 +1,325 @@
...
+ *     dsn: https://00059966e6224d03a77ea5eca10fbe18@sentry.mycompany.com/14
+ *     release: "1.0.1"
+ *     dist: x86
+ *     environment: staging
+ *     server-name: megaServer
+ * </pre>
+ */
+@ConfigurationProperties("sentry")
+@Open
+public class SentryProperties {
```
bruno-garcia, line 26: Isn't this the same properties class used on logback? is this something we can push to sentry-core or something to avoid redeclaring it?

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryRequestHttpServletRequestProcessor.java
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+/** Attaches information about HTTP request to {@link SentryEvent}. */
+@Open
+public class SentryRequestHttpServletRequestProcessor implements EventProcessor {
+
+  @Override
+  public @NotNull SentryEvent process(final @NotNull SentryEvent event, final @Nullable Object hint) {
+    final RequestAttributes requestAttributes = RequestContextHolder.currentRequestAttributes();
```
bruno-garcia, line 24: We should generally make sure that our bugs don't affect the running app. That often means some not elegant code.

Here for example we could have this whole block on a try/catch and use the diagnostic logger to print out the error. That means we'd need to SentryOptions passed in (or the logger reference but the options allows setting up a logger after the SDK is initialized and swap the reference after that).

Alternatively we can catch exceptions of any event processor inside the SentryClient and log there. But some context would be gone I assume.

build.gradle.kts
```diff
@@ -51,7 +51,7 @@ allprojects {
             dependsOn("cleanTest")
         }
         withType<JavaCompile> {
-            options.compilerArgs.addAll(arrayOf("-Xlint:all", "-Werror", "-Xlint:-classfile"))
+            options.compilerArgs.addAll(arrayOf("-Xlint:all", "-Werror", "-Xlint:-classfile", "-Xlint:-processing"))
```
maciejwalkowiak, line 54: Adding Spring Boot `annotationProcessor` in Gradle config causes warnings that not all annotations get processed - which does not make much sense as we do not want to process all the annotations. That's why this warning gets turned off here.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryProperties.java
maciejwalkowiak: I would prefer to keep it as a refactoring outside of this PR.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryRequestHttpServletRequestProcessor.java
maciejwalkowiak: It's not possible to inject `SentryOptions` here as it would create a circular dependency. I will add catching exceptions in `SentryClient` when processing events.

sentry-spring-boot-starter/build.gradle.kts
```diff
@@ -0,0 +1,117 @@
...
+        mavenBom(org.springframework.boot.gradle.plugin.SpringBootPlugin.BOM_COORDINATES)
+    }
+}
+
+configure<JavaPluginConvention> {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
```
marandaneto, line 27: it'd be nice to import `org.jetbrains.kotlin.gradle.tasks.KotlinCompile` on the top of the file and just use `KotlinCompile`, same for other full package classes, just a nit though.

sentry-spring-boot-starter/build.gradle.kts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+dependencies {
+    api(project(":sentry-core"))
+    implementation(Config.Libs.springBootStarter)
+    implementation(Config.Libs.springWeb)
+    implementation(Config.Libs.servletApi)
+
+    annotationProcessor("org.springframework.boot:spring-boot-autoconfigure-processor")
+    annotationProcessor("org.springframework.boot:spring-boot-configuration-processor")
```
marandaneto, line 38: can we also add to the `Config` class?

marandaneto: have you tried to use kapt? https://kotlinlang.org/docs/reference/kapt.html

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryProperties.java
marandaneto: yeah it makes sense to me, otherwise, when we deprecate or add something, it'd need to be done on different classes and easy to forget.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryRequestHttpServletRequestProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+    if (requestAttributes instanceof ServletRequestAttributes) {
+      final HttpServletRequest request =
+          ((ServletRequestAttributes) requestAttributes).getRequest();
+      event.setRequest(resolveSentryRequest(request));
+    }
+    return event;
+  }
+
+  @SuppressWarnings("JdkObsolete")
```
marandaneto, line 34: a comment why we have `JdkObsolete` here

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryUserHttpServletRequestProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+          ((ServletRequestAttributes) requestAttributes).getRequest();
+      event.setUser(toUser(event.getUser(), request));
+    }
+    return event;
+  }
+
+  private static @NotNull User toUser(
+      final @Nullable User existingUser, final @NotNull HttpServletRequest request) {
+    final User user = existingUser != null ? existingUser : new User();
+    user.setIpAddress(toIpAddress(request));
```
marandaneto, line 32: I'm afraid that this would require the `send_default_pii` flag to be added on options, which is disabled by default.
https://docs.sentry.io/error-reporting/configuration/?platform=python

sentry-spring-boot-starter/build.gradle.kts
maciejwalkowiak: I haven't but I also wouldn't like to opt-out from Spring Boot recommended way of creating starters.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryUserHttpServletRequestProcessor.java
maciejwalkowiak: Seems like it is working for me

image

marandaneto: but that's the problem, the SDK should only capture (automatically) PII if the flag `send_default_pii` is enabled, as we don't even have the flag, sending this info would be a PII problem, basically, to capture and send the IP address should only be possible if the user has enabled this flag like `sentry.sendDefaultPII=true`, as we don't even have/expose this flag yet, I'd say that we should skip the IP address for now (maybe just commenting the code and adding a TODO).
or to implement the flag on this or another PR.

sentry-spring-boot-starter/build.gradle.kts
marandaneto: I see, well if it works, its fine, I know that on Android/Gradle if you have the Kotlin plugin applied, `kapt` is always preferred, because `annotationProcessor` won't work for Kotlin code anyway, as we don't have annotation processing on Android right now nor we use Kotlin code (only on tests), it's fine.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryUserHttpServletRequestProcessor.java
maciejwalkowiak: Ah ok, understood. I imagine username could be also considered PII. In case of Spring integration, this filter has to be added explicitly by the user as we are not able to autoconfigure it. Maybe it's ok then to stay as it's just an opt-in option?

marandaneto: yeah usually those things are opt-in.
the thing is, we need to make it opt-in when we are able to collect this info automatically eg IP Address, if the user depends on the customer setting like Sentry.setUser(...) then it's fine, I'm just not sure if the IP address here is set by the customer or Spring offers this for free, that's the question that should be answered.

maciejwalkowiak: We either take the ip address from the `X-FORWARDED-FOR` header set by load balancers or directly from `HttpServletRequest` object. But in order to use it, users have to explicitly create an instance of this filter and add it to Spring Security configuration (example https://github.com/getsentry/sentry-android/pull/517/files#diff-6c5b2fc2c37b9e5c8bdbdc86be9588ccR30)

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryProperties.java
maciejwalkowiak: We have two options:

1. Create `SentryProperties` in `core` module that will contain subset of properties from `SentryOptions` and a method `applyTo(SentryOptions)` which will copy values from `SentryProperties` to `SentryOptions`.

2. Extract subset of properties from `SentryOptions` into new class - `SentryProperties` and add field `properties` in `SentryOptions` - this means that everywhere in the code where we are currently referring to one of these properties we will need to additionally do `getProperties().getXXX`, for example: `options.getProperties().isDebug()`.

Option 1 - is relatively simple to add and would affect only backend-related integrations but does not solve the duplication between `SentryProperties` and `SentryOptions`.
Option 2- requires changing all the code where `SentryOptions` is called but completely removes any duplication of `SentryOptions` properties.

Which direction is preferred? @bruno-garcia @marandaneto

bruno-garcia: Option 1 works for me. It can also land as a separate refactor PR as you said.

One other proposal would be to have SentryOptions inherit from SentryProperties. We can use (or create if not there) the overload that takes the instance as opposed to the callback for `init`. 

The code that binds values to SentryProperties would then take the instance of SentryOptions instead, and there wouldn't be the need of a `applyTo` or any code duplication.

Would that work?

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryUserHttpServletRequestProcessor.java
bruno-garcia: What @marandaneto said.
Could you please add `Boolean sendDefaultPii` to `SentryOptions` and checking for that before adding any of this user data by default?

sentry-core/src/main/java/io/sentry/core/Sentry.java
```diff
@@ -123,6 +123,15 @@ public static void init(
...
   }
 
+  /**
+   * Initializes the SDK with a SentryOptions.
+   *
+   * @param options options the SentryOptions
+   */
+  public static void init(final @NotNull SentryOptions options) {
+    init(options, GLOBAL_HUB_DEFAULT_MODE);
+  }
```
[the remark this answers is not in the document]
bruno-garcia: Yeah I rather we don't add this if possible but if we need it here (i,e: the instance is created by a framework to bind configuration values to it) then we go with it.

Maybe we throw the `@InternalApi` on it? I asssume IDEs don't show it. [Literally what I did in .NET.](https://github.com/getsentry/sentry-dotnet/blob/02e622cd5b82449cfdabbe6a6fb926e71e5101a5/src/Sentry/SentrySdk.cs#L78-L79)

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -79,7 +79,17 @@ public SentryClient(final @NotNull SentryOptions options, @Nullable Connection c
     }
 
     for (EventProcessor processor : options.getEventProcessors()) {
```
bruno-garcia, line 81: I realize this code was here already, so maybe a separate PR but we should just run the processors once (have have the same code on `applyScope` below using processors from the scope. We could merge the lists (run scope processors last). I'll raise an issue.

sentry-core/src/main/java/io/sentry/core/SentryClient.java
```diff
@@ -79,7 +79,17 @@ public SentryClient(final @NotNull SentryOptions options, @Nullable Connection c
...
-      event = processor.process(event, hint);
+      try {
```
bruno-garcia, line 82: M: Since this is not duplicated (on a difference list of event processors), could this be extracted to a private method that takes `List<EventProcessor>` and `SentryEvent` and called from here and line 288?

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    id("org.springframework.boot") version "2.3.3.RELEASE"
+    id("io.spring.dependency-management") version "1.0.10.RELEASE"
+    kotlin("jvm")
+    kotlin("plugin.spring") version "1.3.72"
+}
+
+group = "com.example"
```
bruno-garcia, line 10:
```suggestion
group = "io.sentry.sample.spring-boot"
```

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
```diff
@@ -0,0 +1,115 @@
...
+        transport.ifAvailable(options::setTransport);
+      };
+    }
+
+    @Bean
+    public @NotNull SentryOptions sentryOptions(
+        final @NotNull Sentry.OptionsConfiguration<SentryOptions> optionsConfiguration,
+        final @NotNull SentryProperties properties,
+        final @NotNull ObjectProvider<GitProperties> gitProperties) {
+      final SentryOptions options = new SentryOptions();
```
bruno-garcia, line 62: Can we use the other overload here (`Sentry.init(o -> o ... `) and run all the code in this block in that callback?

Where is the result of this method used? Where does this instance go? Does it get configuration applied by the framework automatically?

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentrySecurityFilter.java
```diff
@@ -0,0 +1,43 @@
...
+    hub.configureScope(
+        scope ->
+            scope.addEventProcessor(
+                new SentryUserHttpServletRequestProcessor(
+                    request.getUserPrincipal(), toIpAddress(request))));
+    filterChain.doFilter(request, response);
+  }
+
+  private static @NotNull String toIpAddress(final @NotNull HttpServletRequest request) {
+    final String ipAddress = request.getHeader("X-FORWARDED-FOR");
```
bruno-garcia, line 40: Should this be opt-in?
Would technically expose the server to report a fake client IP if the request was sent to the server by the client with this (no-proxy) set?

Also, I believe the value here can be a list (chained-proxy servers) so might require some more handling.

settings.gradle.kts
```diff
@@ -2,11 +2,13 @@ rootProject.name = "sentry"
...
+    "sentry-android-ndk",
+    "sentry-android-core",
+    "sentry-core",
+    "sentry-logback",
+    "sentry-spring-boot-starter",
+    "sentry-samples:sentry-samples-android",
+    "sentry-samples:sentry-samples-console",
+    "sentry-samples:sentry-samples-logback",
+    "sentry-samples:sentry-samples-spring-boot",
+    "sentry-android-timber")
```
bruno-garcia, line 14:
```suggestion
    "sentry-spring-boot-starter",
    "sentry-android-timber",
    "sentry-samples:sentry-samples-android",
    "sentry-samples:sentry-samples-console",
    "sentry-samples:sentry-samples-logback",
    "sentry-samples:sentry-samples-spring-boot")
```

sentry-core/src/main/java/io/sentry/core/Sentry.java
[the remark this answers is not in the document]
marandaneto: IDE still shows it, but it's properly documented and we don't need to care about breaking changes, the annotation states exactly that, so if I needed, it'd be the way to go anyway.
it'd be nice to be able to hide a method or class from the intellitrace though, its not possible afaik

sentry-core/src/main/java/io/sentry/core/SentryClient.java
marandaneto: nice catch

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    id("org.springframework.boot") version "2.3.3.RELEASE"
+    id("io.spring.dependency-management") version "1.0.10.RELEASE"
```
marandaneto, line 5: would be possible to reuse the `Config` class like the SDK itself? they are probably the same groups/modules/versions

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    implementation("org.springframework.boot:spring-boot-starter-security")
+    implementation("org.springframework.boot:spring-boot-starter-web")
+    implementation("org.springframework.boot:spring-boot-starter")
+    implementation("org.jetbrains.kotlin:kotlin-reflect")
+    implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
```
marandaneto, line 23: why did you skip the versioning? this could be a source of a broken example.
I'd reuse the `Config` class too

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+tasks.withType<Test> {
+    useJUnitPlatform()
+}
+
+tasks.withType<KotlinCompile> {
+    kotlinOptions {
+        freeCompilerArgs = listOf("-Xjsr305=strict")
+        jvmTarget = "1.8"
```
marandaneto, line 38:
```suggestion
        jvmTarget = JavaVersion.VERSION_1_8.toString()
```

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+tasks.withType<Test> {
+    useJUnitPlatform()
+}
+
+tasks.withType<KotlinCompile> {
+    kotlinOptions {
+        freeCompilerArgs = listOf("-Xjsr305=strict")
```
marandaneto, line 37: why do we need this? just curiosity

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryProperties.java
marandaneto: I'd say that @bruno-garcia suggestion looks good in theory, let's experiment it if we have a method like `applyTo` we'd still need to take care of it in every change of protocol anyway.
so ideally no code duplication nor breaking changes for Android unless we don't find another way.

marandaneto: the downside of extracting this to code though is that you can't use the class `Optional` like you did and looks nice.
this could be a separate PR, so we have this working this way for now.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryRequestFilter.java
```diff
@@ -0,0 +1,36 @@
...
+import java.io.IOException;
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.jetbrains.annotations.NotNull;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+/** Pushes new {@link io.sentry.core.Scope} on each incoming HTTP request. */
+@Open
```
marandaneto, line 14: why this class can't be `final`? just curious

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryRequestFilter.java
```diff
@@ -0,0 +1,36 @@
...
+public class SentryRequestFilter extends OncePerRequestFilter {
+  private final @NotNull IHub hub;
+
+  public SentryRequestFilter(final @NotNull IHub hub) {
+    this.hub = hub;
+  }
```
marandaneto, line 20: would it make sense that all those classes that require a `hub` have a package-private ctor with the `hub` param for testing purposes, but a public ctor already sets the hub from `HubAdapter.getInstance()` so nobody needs to pass an instance when creating those objects?

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
maciejwalkowiak: The result of this method is injected into a bean factory method `sentryHub`. If we don't have separate bean `SentryOptions` I don't think we are able to properly test autoconfiguration (making it testable was the motivation to change it from what you're suggesting).

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
maciejwalkowiak: this `build.gradle.kts` is kept as close as possible to one generated using https://start.spring.io - I think if we don't touch it much it will be easier to understand for users what exactly needs to be set up to use Sentry.

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
maciejwalkowiak: Versions are managed by `io.spring.dependency-management` plugin - that's the way to use dependencies with Spring Boot.

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
maciejwalkowiak: It's also added by start.spring.io - i think it's meant to tell Kotlin to respect Spring Framework nullability related annotations.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryRequestFilter.java
maciejwalkowiak: Spring creates proxies with CGLIB around beans and it's not possible to proxy final class with CGLIB.

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
marandaneto: I get your point, but the thing is, the samples on this repo are to easier the development/testing/debugging, not really customer samples, they still could learn from it, but that's not the main reason, we have eg https://github.com/getsentry/examples/tree/master/java
we'd be updating the examples after publishing these new integrations and there we'd keep as close as possible.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryRequestFilter.java
maciejwalkowiak: It is very not-spring idiomatic to get dependencies using static methods. For anyone who worked with Spring or Java EE, dependency injection is a natural way to use dependencies, so I would say it should stay as it is.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryRequestFilter.java
marandaneto: makes sense

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryRequestFilter.java
marandaneto: if all those classes are only used internally and managed by us, I'd say its fine, but if one needs to `new SentryRequestFilter(HubAdapter.getInstance())`, then it'd be a bit weird IMO.
the other way would be to refactor `HubAdapter.getInstance()` and find a better approach in a different PR, but I'm happy the way how it is

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
maciejwalkowiak: Ok, got it. Would you also like to put all the dependencies from sample application to `Config` class?

marandaneto: for me, it only makes sense the ones that are also used in the SDK, the dependencies of the samples that are only used in the samples, its fine to be hardcoded.
if it happens that a dependency of a sample is hardcoded but still share a version that is in the config file, then I'd do a string template like `"dependency$Config.sharedVersion"` so we know that if we update something, it's done only in one place, this is just a nit about DRY though.

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
marandaneto: when you don't declare a version, you need either a `constraints` -> https://docs.gradle.org/current/userguide/single_versions.html#sec:declaring_without_version or `dependencyManagement` -> https://docs.spring.io/dependency-management-plugin/docs/current-SNAPSHOT/reference/html/
that's why I don't follow how the version is inferred, maybe the plugin does it internally, but I didn't find the docs for it.

also, I don't think that the `io.spring.dependency-management` also manages the versioning of the kotlin deps.

marandaneto: eg `implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")` could become `implementation(kotlin(Config.kotlinStdLib))`

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryUserHttpServletRequestProcessor.java
maciejwalkowiak: Ok, in this case it is enough to add it to `SentryProperties` but I am wondering if we shouldn't erase any personal information somewhere deeper in `sentry-core` if this flag is set to false?

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
maciejwalkowiak: I also don't know exactly how it works under the hood (one reason I prefer Maven over Gradle is it's simplicity), but there is no need to define anything extra to let this plugin resolve dependencies. By not specifying versions we are sure that the imported libraries are compatible with Spring Boot.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryUserHttpServletRequestProcessor.java
marandaneto: that's a good point, if the flag is a safeguard to not even set it, I don't see how we could do differently other than doing the check guard in all the setters/all the integrations.

the other option would be if `setIpAddress` gets a boolean to set it or not so at least we don't repeat the same `if` everywhere.

we could start doing this check guard in here and if we see the need for other places, we just refactor it?!

the other option would be, we always set it, if possible, and we could create a PiiEventProcessor that reads the flag and erase the values if set, it'd be an elegant way of solving it, but this allows the user to set it in the first place, I don't know if this is fine or not. cc @bruno-garcia

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryRequestFilter.java
maciejwalkowiak: They would not need to do `HubAdapter.getInstance()` but rather inject `IHub` that is created in auto-configuration.

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
marandaneto: wfm

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryRequestFilter.java
marandaneto: got it, we are good to go then

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryProperties.java
bruno-garcia: Agreed. So lets merge then

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryUserHttpServletRequestProcessor.java
bruno-garcia: We don't have any PII to be erased yet. We delayed adding the flag until the first one popped up.
Happy to merge this and add that later

marandaneto (approved):
LGTM

PR #522, opened 2020-08-20
Globally disable JdkObsolete check.

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

This is a proposal - if we decide to disable this check, same changes to other modules will have to be applied.

Disables ErrorProne check for [JdkObsolete](https://errorprone.info/bugpattern/JdkObsolete).

Since must stay compatible with older Android APIs we are going to use classes like `java.util.Date` that ErrorProne marks as `JdkObsolete` and logs a warning. Warnings are treated as errors, so our build fails.

We can mark methods that use obsolete classes explicitly with `@SurpressWarnings("JdkObsolete")` - but this isn't clear at first and ideally there should be also a comment explaining why this annotation is there, or we can disable this check globally - as anyway there is a low likelihood that any of the obsolete classes - other than these that we must use - will be used.

Related discussion https://github.com/getsentry/sentry-android/pull/511#discussion_r472289113

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,4 +1,5 @@
 import com.diffplug.spotless.LineEnding
+import net.ltgt.gradle.errorprone.errorprone
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.gradle.api.tasks.testing.logging.TestLogEvent
 
@@ -32,6 +33,7 @@ buildscript {
 }
 
 allprojects {
+    apply<net.ltgt.gradle.errorprone.ErrorPronePlugin>()
     repositories {
         google()
         jcenter()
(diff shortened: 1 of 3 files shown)


marandaneto:
> I have mixed feelings about this. I believe I'd rather keep declaring the obsoletes but with a comment to what they are about.
> 
> Maybe @marandaneto has thoughts?

would it be possible to disable it only for `java.util.Date`? I understand the need for it, but we lose other checks that might allow us to make mistakes, not only about compatibility with older Android versions but also choosing less performative objects, etc...

maciejwalkowiak:
As far as I am aware there is no option to disable check for single class. I understand the reasoning to keep it, the question is - shall we document each usage of `@SuppressWarnings("JdkObsolete")` with a proper comment?

marandaneto:
> As far as I am aware there is no option to disable check for single class. I understand the reasoning to keep it, the question is - shall we document each usage of `@SuppressWarnings("JdkObsolete")` with a proper comment?

that sounds a solution, I'll add it to my list and see if I can remember the reason for all of them :D
should we close this then? thanks for bringing up the discussion though.

maciejwalkowiak:
Yep 👍  Closing.

Issue #520, opened 2020-08-14
Method addObserver must be called on the main thread


lewisf:
**Platform:**
- [ ]  iOS
- [x]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**RN SDK version:** 1.6.3
**Android SDK version:** 2.2.0

**`react-native` version:** 0.62.17

**Are you using Expo?**
- [ ]  Yes
- [x]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)
- [ ]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

https://sentry.io/organizations/persona-identities-inc/issues/1839866967/?project=5378601&query=is%3Aunresolved

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  dsn: 'https://...@sentry.io/...'
});
```

---
**I have following issue:**

I went through barebones setup steps and am getting this error. Commenting out the `Sentry.init` gets rid of it. Stack trace seems to indicate it's all happening within the Sentry library but I'm not sure if I'm doing something wrong.

**Steps to reproduce:**
- New project on React Native 0.62.17
- Install Sentry
- Launch in Android Emulator using Pixel 2 API 29
- Red screen with error

**Actual result:**

![image](https://user-images.githubusercontent.com/588210/90283938-b8e77c00-de36-11ea-827b-cdc0cb5344c8.png)

**Expected result:**

No error.

lewisf:
Thanks. Is there anything else I can do to help with debugging?

zumaawad:
I have the same issue with react-native
```
    "react": "16.11.0",
    "react-native": "0.62.2",
    "@sentry/react-native": "^1.7.1"
```

marandaneto:
hey thanks for reporting this, quick question:

are you initing the Android SDK manually by yourself or are you letting the Sentry RN SDK do the job automatically?

If you are letting the Sentry RN SDK do the job, on which life cycle (which class/method) of the RN App. are ou initing the Sentry RN SDK? thanks :)

marandaneto:
@lewisf @zumaawad what version of the `androidx.lifecycle/lifecycle-process` lib are you working with btw? are you referencing directly or is it a transitive dependency? could you track this down, please? `2.2.0` doesn't have this check `enforceMainThreadIfNeeded` but maybe an older version have it.

[...]

PR #525, opened 2020-08-24
fix: add main thread checker for the app lifecycle integration

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: add main thread checker for the app lifecycle integration
this will make the App. to not crash it, but still won't be able to add the observer, see next steps.

https://github.com/getsentry/sentry-android/issues/520

unit test and running on older androidx versions.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## :crystal_ball: Next steps
maybe execute the addObserver on the main Looper

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
@@ -49,7 +49,6 @@ public final class AppLifecycleIntegration implements Integration, Closeable {
         Class.forName("androidx.lifecycle.ProcessLifecycleOwner");
         if (MainThreadChecker.isMainThread()) {
           addObserver(hub);
-          options.getLogger().log(SentryLevel.DEBUG, "AppLifecycleIntegration installed.");
         } else {
           // some versions of the androidx lifecycle-process require this to be executed on the main
           // thread.
@@ -63,6 +62,10 @@ public final class AppLifecycleIntegration implements Integration, Closeable {
                 SentryLevel.INFO,
                 "androidx.lifecycle is not available, AppLifecycleIntegration won't be installed",
                 e);
+      } catch (IllegalStateException e) {
+        options
+            .getLogger()
+            .log(SentryLevel.ERROR, "AppLifecycleIntegration could not be installed", e);
       }
     }
   }
@@ -75,6 +78,7 @@ public final class AppLifecycleIntegration implements Integration, Closeable {
             this.options.isEnableSessionTracking(),
             this.options.isEnableAppLifecycleBreadcrumbs());
     ProcessLifecycleOwner.get().getLifecycle().addObserver(watcher);
+    options.getLogger().log(SentryLevel.DEBUG, "AppLifecycleIntegration installed.");
   }
 
   @Override
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
@@ -1,6 +1,7 @@
 package io.sentry.android.core;
 
 import androidx.lifecycle.ProcessLifecycleOwner;
+import io.sentry.android.core.util.MainThreadChecker;
 import io.sentry.core.IHub;
 import io.sentry.core.Integration;
 import io.sentry.core.SentryLevel;
@@ -44,15 +45,22 @@ public final class AppLifecycleIntegration implements Integration, Closeable {
       try {
         Class.forName("androidx.lifecycle.DefaultLifecycleObserver");
         Class.forName("androidx.lifecycle.ProcessLifecycleOwner");
-        watcher =
-            new LifecycleWatcher(
-                hub,
-                this.options.getSessionTrackingIntervalMillis(),
-                this.options.isEnableSessionTracking(),
-                this.options.isEnableAppLifecycleBreadcrumbs());
-        ProcessLifecycleOwner.get().getLifecycle().addObserver(watcher);
-
-        options.getLogger().log(SentryLevel.DEBUG, "AppLifecycleIntegration installed.");
+        if (MainThreadChecker.isMainThread()) {
+          watcher =
+              new LifecycleWatcher(
+                  hub,
+                  this.options.getSessionTrackingIntervalMillis(),
+                  this.options.isEnableSessionTracking(),
+                  this.options.isEnableAppLifecycleBreadcrumbs());
+          ProcessLifecycleOwner.get().getLifecycle().addObserver(watcher);
+          options.getLogger().log(SentryLevel.DEBUG, "AppLifecycleIntegration installed.");
+        } else {
+          options
+              .getLogger()
+              .log(
+                  SentryLevel.WARNING,
+                  "AppLifecycleIntegration is not running on the main thread.");
+        }
       } catch (ClassNotFoundException e) {
         options
             .getLogger()
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
@@ -1,5 +1,7 @@
 package io.sentry.android.core;
 
+import android.os.Handler;
+import android.os.Looper;
 import androidx.lifecycle.ProcessLifecycleOwner;
 import io.sentry.android.core.util.MainThreadChecker;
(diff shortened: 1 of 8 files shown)


sentry-core/src/main/java/io/sentry/core/Sentry.java
```diff
@@ -133,11 +133,10 @@ private static synchronized void init(
...
       options
-        .getLogger()
-        .log(
-          SentryLevel.WARNING,
-          "Sentry has been already initialized. Previous configuration will be overwritten."
-        );
+          .getLogger()
+          .log(
+              SentryLevel.WARNING,
+              "Sentry has been already initialized. Previous configuration will be overwritten.");
```
marandaneto, line 139: just a spotless change

sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
```diff
@@ -44,15 +47,15 @@ public void register(final @NotNull IHub hub, final @NotNull SentryOptions optio
...
-
-        options.getLogger().log(SentryLevel.DEBUG, "AppLifecycleIntegration installed.");
+        if (MainThreadChecker.isMainThread()) {
+          addObserver(hub);
+          options.getLogger().log(SentryLevel.DEBUG, "AppLifecycleIntegration installed.");
+        } else {
+          // some versions of the androidx lifecycle-process require this to be executed on the main
+          // thread.
+          final Handler handler = new Handler(Looper.getMainLooper());
+          handler.post(() -> addObserver(hub));
```
marandaneto, line 57: that's my suggestion to just not ignore if it's not the main thread.

bruno-garcia: sounds good

sentry-android-core/src/test/java/io/sentry/android/core/util/MainThreadCheckerTest.kt
```diff
@@ -0,0 +1,28 @@
+package io.sentry.android.core.util
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import kotlin.test.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class MainThreadCheckerTest {
```
philipphofmann, line 10: `m`: Nice tests, would it make sense to add a test, that starts a thread, calls `MainThreadChecker.isMainThread()` inside it and makes sure it is false?

sentry-android-core/src/main/java/io/sentry/android/core/util/MainThreadChecker.java
```diff
@@ -0,0 +1,46 @@
...
+    return isMainThread(Thread.currentThread());
+  }
+
+  /**
+   * Checks if a given thread is the Android Main/UI thread
+   *
+   * @param sentryThread the SentryThread
+   * @return true if it is the main thread or false otherwise
+   */
+  public static boolean isMainThread(final @NotNull SentryThread sentryThread) {
```
philipphofmann, line 39: `m`: If I am not mistaken, we don't have a test for this.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -181,7 +181,7 @@ private void processNonCachedEvent(final @NotNull SentryEvent event) {
 
     if (event.getThreads() != null) {
       for (SentryThread thread : event.getThreads()) {
-        thread.setCurrent(Looper.getMainLooper().getThread().getId() == thread.getId());
+        thread.setCurrent(MainThreadChecker.isMainThread(thread));
```
philipphofmann, line 184: Nice that we also use it here 👏

sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
```diff
@@ -44,15 +47,15 @@ public void register(final @NotNull IHub hub, final @NotNull SentryOptions optio
...
-                hub,
-                this.options.getSessionTrackingIntervalMillis(),
-                this.options.isEnableSessionTracking(),
-                this.options.isEnableAppLifecycleBreadcrumbs());
-        ProcessLifecycleOwner.get().getLifecycle().addObserver(watcher);
-
-        options.getLogger().log(SentryLevel.DEBUG, "AppLifecycleIntegration installed.");
+        if (MainThreadChecker.isMainThread()) {
+          addObserver(hub);
+          options.getLogger().log(SentryLevel.DEBUG, "AppLifecycleIntegration installed.");
```
philipphofmann, line 52: `l`: don't we also want to have this log message in the `else` below?

sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
```diff
@@ -44,15 +47,15 @@ public void register(final @NotNull IHub hub, final @NotNull SentryOptions optio
...
-        watcher =
-            new LifecycleWatcher(
-                hub,
-                this.options.getSessionTrackingIntervalMillis(),
-                this.options.isEnableSessionTracking(),
-                this.options.isEnableAppLifecycleBreadcrumbs());
-        ProcessLifecycleOwner.get().getLifecycle().addObserver(watcher);
-
-        options.getLogger().log(SentryLevel.DEBUG, "AppLifecycleIntegration installed.");
+        if (MainThreadChecker.isMainThread()) {
```
philipphofmann, line 50: `l`: Just out of curiosity, why don't we just always do?
```
final Handler handler = new Handler(Looper.getMainLooper());
handler.post(() -> addObserver(hub));
```

philipphofmann (approved):
LGTM, maybe we can improve a few things though.

sentry-android-core/src/test/java/io/sentry/android/core/util/MainThreadCheckerTest.kt
marandaneto: it'd make sense, but as I tested them individually, not sure if it's any useful.

sentry-android-core/src/main/java/io/sentry/android/core/util/MainThreadChecker.java
marandaneto: yep, I will add it

sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
marandaneto: good catch, the `else` came later for just getting a review and showing an example and forgot to move things, thx

sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
marandaneto: if it's already in the main thread, we know that the code will be executed right away.
using a handler, we don't control when this code will be executed (most probably immediately), but let's say that an App. uses `handler/post` heavily, we might be delaying this execution for a bit and things might get out of control, just because of that wdyt?

billnbell:
Will this be a release to npm ?

bruno-garcia:
> Will this be a release to npm ?

Once we release the Android SDK, we bump the [react-native SDK's dependency](https://github.com/getsentry/sentry-react-native)

zumaawad:
When this PR will be released?

sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
philipphofmann: Make sense for me to keep this little extra complexity for such an edge case. Thanks for explaining.

philipphofmann (approved):
LGTM

sentry-android-core/src/test/java/io/sentry/android/core/util/MainThreadCheckerTest.kt
marandaneto: I added a test for AppLifecycleIntegration which basically does that

marandaneto:
> When this PR will be released?

I need to sync with the RN SDK, but most probably this week (early as possible) as we know this is a blocker for some users.

PR #527, opened 2020-08-28
Add PiiEventProcessor to filter out personal identifiable information

maciejwalkowiak:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds an option for users to decide if they want to send personal identifiable information (email, username, ip address) along with the events.

Discussion initially started when working on Spring Boot integration - we want to give users option to decide if detailed user information is sent or not - similar to how it's done in Python integration: https://docs.sentry.io/error-reporting/configuration/?platform=python

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

Breaking change: sending PII is turn off by default.

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/PiiEventProcessor.java
+++ b/sentry-core/src/main/java/io/sentry/core/PiiEventProcessor.java
@@ -1,0 +1,33 @@
+package io.sentry.core;
+
+import io.sentry.core.protocol.User;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Removes personal identifiable information from {@link SentryEvent} if {@link
+ * SentryOptions#isSendDefaultPii()} is set to false.
+ */
+@ApiStatus.Internal
+final class PiiEventProcessor implements EventProcessor {
+  private final @NotNull SentryOptions options;
+
+  PiiEventProcessor(final @NotNull SentryOptions options) {
+    this.options = options;
+  }
+
+  @Override
+  public SentryEvent process(SentryEvent event, @Nullable Object hint) {
+    if (!options.isSendDefaultPii()) {
+      final User user = event.getUser();
+      if (user != null) {
+        user.setUsername(null);
+        user.setIpAddress(null);
+        user.setEmail(null);
+        event.setUser(user);
+      }
+    }
+    return event;
+  }
+}
--- a/sentry-core/src/main/java/io/sentry/core/PiiEventProcessor.java
+++ b/sentry-core/src/main/java/io/sentry/core/PiiEventProcessor.java
@@ -1,7 +1,10 @@
 package io.sentry.core;
 
+import io.sentry.core.protocol.Request;
 import io.sentry.core.protocol.User;
-import org.jetbrains.annotations.ApiStatus;
+import io.sentry.core.util.CollectionUtils;
+import io.sentry.core.util.Objects;
+import java.util.Map;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -9,12 +12,11 @@ import org.jetbrains.annotations.Nullable;
  * Removes personal identifiable information from {@link SentryEvent} if {@link
  * SentryOptions#isSendDefaultPii()} is set to false.
  */
-@ApiStatus.Internal
 final class PiiEventProcessor implements EventProcessor {
   private final @NotNull SentryOptions options;
 
   PiiEventProcessor(final @NotNull SentryOptions options) {
-    this.options = options;
+    this.options = Objects.requireNonNull(options, "The options object is required");
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


sentry-core/src/main/java/io/sentry/core/PiiEventProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+  private final @NotNull SentryOptions options;
+
+  PiiEventProcessor(final @NotNull SentryOptions options) {
+    this.options = options;
+  }
+
+  @Override
+  public SentryEvent process(SentryEvent event, @Nullable Object hint) {
+    if (!options.isSendDefaultPii()) {
+      final User user = event.getUser();
```
marandaneto, line 23: do we have any PII set (like IP) in `event.request` that we should strip off as well? maybe in `event.request.headers` if it has a fixed key.

sentry-core/src/main/java/io/sentry/core/PiiEventProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+import io.sentry.core.protocol.User;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Removes personal identifiable information from {@link SentryEvent} if {@link
+ * SentryOptions#isSendDefaultPii()} is set to false.
+ */
+@ApiStatus.Internal
```
marandaneto, line 12: if the class is not public and only package-private, we don't need `@ApiStatus.Internal` I'd say.

sentry-core/src/main/java/io/sentry/core/PiiEventProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+final class PiiEventProcessor implements EventProcessor {
+  private final @NotNull SentryOptions options;
+
+  PiiEventProcessor(final @NotNull SentryOptions options) {
+    this.options = options;
```
marandaneto, line 17: we could use `Objects.requireNonNull` just for the sake of not getting NPE on `options.isSendDefaultPii()`

marandaneto: just reading it again and as this is internal, package-private, it's not really necessary, we are the only one creating it anyway.

marandaneto (commented):
looks good, left a few comments.
I'd uncheck `No breaking changes` on the PR template because users would need to swap the flag on App. init to send user data now and previously, this was sent by default.

A note for me would be to add an Android-friendly way of swapping this flag eg `ManifestMetadataReader` so people can configure that on the Manifest file like any other option.

sentry-core/src/main/java/io/sentry/core/PiiEventProcessor.java
maciejwalkowiak: Good point. I also realised that we need to make sure that `PiiEventProcessor` is the last processor that runs, otherwise other processors added later can actually add PII.

marandaneto: good point, I noticed that too, that should be the last one, but I dont see how we can do that easily, any other processor added by the client manually will be added later, we can guarantee that its the last one in the SDK, but not the last one overall, I guess.
ideas are welcome of course.

maciejwalkowiak: I extracted `PiiEventProcessor` to separate field. Another option would be to have special type of list with method `addBefore(processor, Class<T>)` but I think in this case it would be overengineering.

marandaneto: yeah liked the idea, I'm fine with it, maybe @bruno-garcia has opinions.

bruno-garcia: I'd prefer to avoid adding any pii to begin with.

maciejwalkowiak:
> Can we just not add abrocessor and simply not add pii?
> Also folks are against processors that are not added as integrations.

Lets take Spring integration as an example - do you mean to not have a flag on sentry options and the Spring Security filter that adds user data? This filter is not configured anyway, so the idea with a flag and processor was just a convenient way to make sure that no PII data is sent to sentry no matter what is configured on the application side.

sentry-core/src/main/java/io/sentry/core/SentryOptions.java
```diff
@@ -996,6 +1001,19 @@ public void setSdkVersion(final @Nullable SdkVersion sdkVersion) {
...
+
+  public void setSendDefaultPii(boolean sendDefaultPii) {
+    this.sendDefaultPii = sendDefaultPii;
+  }
+
+  @NotNull
+  public EventProcessor getPiiEventProcessor() {
+    return piiEventProcessor;
+  }
+
```
bruno-garcia, line 1016: I believe it's safer to keep this API out and simply have the options reference where needed to inspect for `sendDefaultPii` before adding any PII in the first place.

This will help with auditing the code for PII, avoid issues where we somehow don't run the processor accidentally (change of order or some other bug) and also forget to add a clean up logic for something we collect as PII without the guard

maciejwalkowiak:
Superseded by #531.

PR #528, opened 2020-08-28
Refactor using common parts of `SentryOptions`.

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Use common subset of SentryOptions in all 3rd party integrations to avoid code duplication and possibility to miss support of newly added properties in those integrations.

Duplication between Logback and Spring Boot integration.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


marandaneto (approved):
LGTM.
just out of curiosity, why you didn't go with the inheritance solution?

sentry-core/src/main/java/io/sentry/core/SentryCommonOptions.java
```diff
@@ -0,0 +1,263 @@
...
+  private Integer readTimeoutMillis;
+
+  /** @see SentryOptions#bypassSecurity */
+  private Boolean bypassSecurity;
+
+  /** @see SentryOptions#debug */
+  private Boolean debug;
+
+  /** @see SentryOptions#diagnosticLevel */
+  private SentryLevel diagnosticLevel = SentryLevel.DEBUG;
```
marandaneto, line 30: can you use `SentryOptions.DEFAULT_DIAGNOSTIC_LEVEL` here?

sentry-core/src/main/java/io/sentry/core/SentryCommonOptions.java
```diff
@@ -0,0 +1,263 @@
...
+  private Long flushTimeoutMillis;
+
+  /** @see SentryOptions#readTimeoutMillis */
+  private Integer readTimeoutMillis;
+
+  /** @see SentryOptions#bypassSecurity */
+  private Boolean bypassSecurity;
+
+  /** @see SentryOptions#debug */
+  private Boolean debug;
```
marandaneto, line 27: should we use primitive types? otherwise, we'd need to do NPE checks

sentry-core/src/main/java/io/sentry/core/SentryCommonOptions.java
```diff
@@ -0,0 +1,263 @@
...
+  public void setDebug(Boolean debug) {
+    this.debug = debug;
+  }
+
+  public SentryLevel getDiagnosticLevel() {
+    return diagnosticLevel;
+  }
+
+  public void setDiagnosticLevel(SentryLevel diagnosticLevel) {
+    this.diagnosticLevel = diagnosticLevel;
```
marandaneto, line 181: if you look at `SentryOptions.setDiagnosticLevel` you see that we don't allow null values, and here it is possible, if one sets it to null, it'll crash the app.

marandaneto (commented):
before LGTM, left a few comments.

sentry-core/src/main/java/io/sentry/core/SentryCommonOptions.java
bruno-garcia: The advantage is being able to know if the value was set explicitly by the user and being able to flip the default after invoking a user callback.

But for this field I agree probably doesn't make sense

sentry-core/src/main/java/io/sentry/core/SentryCommonOptions.java
maciejwalkowiak: In `applyTo` there is a check for null. The idea is that we apply only properties explicitly set by user.

sentry-core/src/main/java/io/sentry/core/SentryCommonOptions.java
maciejwalkowiak: As @bruno-garcia said, we need to know which values were set explicitly by user. This way we don't need to keep in sync default values on `SentryOptions` and `SentryCommonOptions`. This also includes data type for fields - they have to be nullable otherwise we loose this information.

sentry-core/src/main/java/io/sentry/core/SentryCommonOptions.java
maciejwalkowiak: It should be null by default. My bad.

maciejwalkowiak:
> LGTM.
> just out of curiosity, why you didn't go with the inheritance solution?

@marandaneto I don't see how I can apply inheritance here to achieve this.

sentry-core/src/main/java/io/sentry/core/SentryCommonOptions.java
bruno-garcia: It should be `DEBUG` by default (perhaps with a const `DEFAULT_DIAGNOSTIC_LEVEL`) though, right?

If the user sets `option.setDebug(true)` it should automatically start printing all diag msgs of all levels

maciejwalkowiak:
Superseded by #530

PR #529, opened 2020-08-28
Make SentryOptions inherit from SentryCommonOptions

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Refactoring for https://github.com/getsentry/sentry-android/pull/528

Make SentryOptions inherit from SentryCommonOptions and use reflection to copy values.

Thanks to using reflection and inheritance we limit even more number of places that have to be modified when new properties are added.

Existing unit & integration tests cover reflection usage.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

The question is if it's the way to go or @bruno-garcia you had something different in mind?


sentry-core/src/main/java/io/sentry/core/SentryCommonOptions.java
```diff
@@ -1,263 +1,450 @@
...
+  private @Nullable String dsn;
 
-  /** @see SentryOptions#flushTimeoutMillis */
-  private Long flushTimeoutMillis;
+  /**
+   * Controls how many seconds to wait before shutting down. Sentry SDKs send events from a
+   * background queue and this queue is given a certain amount to drain pending events Default is
+   * 2000 = 2s
+   */
+  private long shutdownTimeout = 2000; // 2s
```
bruno-garcia, line 27: I wouldn't rename any of the options not to bring any breaking changes here.

Also, I'd suggest we're always explicit on the unit. Since this is `long`, this timeout is in Seconds? MIlliseconds?

Joshua Bloch talks about that in this video which I highly recommend: https://www.youtube.com/watch?v=heh4OeB9A-c

maciejwalkowiak: I agree. It got renamed because there was a mismatch between setters/getters and the field name, thus it didn't comply to Java Beans convention making it impossible to use with Spring `@ConfigurationProperties`. Either we rename the field (as it's done in this PR) and do not introduce breaking changes (as the setters & getters do not change), or follow Joshua Bloch suggestion and rename setters and getters.

maciejwalkowiak:
Superseded by #530

PR #536, opened 2020-09-01
WIP: load SentryOptions from external sources.

maciejwalkowiak:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

The point of this PR at this stage is to discuss and think if this is the direction we want to go, or are there better ideas.

Adds the ability to load `SentryOptions` from external sources:

- environment variables
- system properties
- `sentry.properties` file

This is still WIP with missing:

- tests
- passing logger everywhere
- polishing the code (adding `@NotNull` etc)
- proper package structure
- using this code in integrations

This feature is being brought in for the sake of compatibility with 1.x branch. Most of the code is either copied or inspired by the same functionality from 1.x branch with certain changes:

- we do not support anymore resolving properties from the DSN url
- we do not use SLF4J for logging but rather pass an instance of `ILogger`.

What is supported:

- loading from JNDI registry
- loading from environment properties
- loading from system properties
- loading from properties file
  - properties file name can be:
    - `sentry.properties`
    - or file name can be obtained from env variable `SENTRY_PROPERTIES_FILE`
    - or file name can be obtained from system property `sentry.properties.file`
  - file can be located:
    - in the classpath
    - file system

`SentryOptionsProvider` is an entry-point to resolve `SentryOptions`. We can either use it inside no-arg `Sentry#init()` method, or resolve options and pass it to `Sentry#init(options)`.

TODO

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/config/provider/JndiConfigurationProvider.java
+++ b/sentry-core/src/main/java/io/sentry/core/config/provider/JndiConfigurationProvider.java
@@ -1,0 +1,88 @@
+package io.sentry.core.config.provider;
+
+
+import io.sentry.core.ILogger;
+import io.sentry.core.SentryLevel;
+
+import javax.naming.Context;
+import javax.naming.InitialContext;
+import javax.naming.NamingException;
+import javax.naming.NoInitialContextException;
+
+/**
+ * A configuration provider that looks up the configuration in the JNDI registry.
+ *
+ * @see JndiSupport to check whether Jndi is available in the current JVM without needing to load this class
+ */
+final class JndiConfigurationProvider implements ConfigurationProvider {
+    /**
+     * The default prefix of the JNDI names of Sentry configuration. This, concatenated with the configuration key name
+     * provided to the {@link #getProperty(String)} method, must form a valid JNDI name.
+     */
+    public static final String DEFAULT_JNDI_PREFIX = "java:comp/env/sentry/";
+
+    private final String prefix;
+    private final JndiContextProvider contextProvider;
+    private final ILogger logger;
+
+    /**
+     * Constructs a new instance using the {@link #DEFAULT_JNDI_PREFIX} and {@link JndiContextProvider} that returns
+     * a new {@link InitialContext} each time it is asked for a context. This ensures that any changes in the JNDI
+     * environment are taken into account on the next configuration property lookup.
+     */
+    public JndiConfigurationProvider(ILogger logger) {
+        this(DEFAULT_JNDI_PREFIX, new JndiContextProvider() {
+            @Override
+            public Context getContext() throws NamingException {
+                return new InitialContext();
+            }
+        }, logger);
+    }
+
+    /**
+     * Constructs a new instance that will look up the Sentry configuration keys using the provided prefix and using
+     * the context obtained from the JNDI context provider. Because the JNDI environment is dynamic, Sentry uses the
+     * JNDI context provider to obtain a fresh copy of the environment (if the provider chooses to return such).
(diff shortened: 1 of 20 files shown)


sentry-core/src/main/java/io/sentry/core/config/provider/SentryOptionsProvider.java
```diff
@@ -0,0 +1,83 @@
...
+    final String dsn = configurationProvider.getProperty("dsn");
+    if (dsn != null) {
+      options.setDsn(dsn);
+    }
+
+    final Long shutdownTimeout = getLong("shutdownTimeout");
+    if (shutdownTimeout != null) {
+      options.setShutdownTimeout(shutdownTimeout);
+    }
+
```
maciejwalkowiak, line 75: Here we will need to set all the non-null properties from external config

sentry-core/src/main/java/io/sentry/core/config/provider/JndiSupport.java
```diff
@@ -0,0 +1,26 @@
...
+    private JndiSupport() {
+    }
+
+    /**
+     * Checks whether JNDI is available.
+     * @return true if JNDI is available, false otherwise
+     */
+    public static boolean isAvailable() {
+        try {
+            Class.forName("javax.naming.InitialContext", false, JndiSupport.class.getClassLoader());
```
bruno-garcia, line 19: This isnt' very Android friendly.
We'd need a clear way to init without going through all these Java ways

sentry-core/src/main/java/io/sentry/core/config/provider/JndiConfigurationProvider.java
```diff
@@ -0,0 +1,88 @@
+package io.sentry.core.config.provider;
+
+
+import io.sentry.core.ILogger;
+import io.sentry.core.SentryLevel;
+
+import javax.naming.Context;
+import javax.naming.InitialContext;
+import javax.naming.NamingException;
+import javax.naming.NoInitialContextException;
```
marandaneto, line 10: this cant live in the sentry-core package.
Android SDK has a subset of the Java SDK, javax is not one of them

sentry-core/src/main/java/io/sentry/core/config/provider/JndiSupport.java
marandaneto: I'd say that none of the providers should be "executed" on Android, but just its own AndroidProvider.

it doesnt make sense to try to read system envs, locate the file etc if the only thing that makes sense for android is its own provider.
ideally, the options have a list of providers, and its empty by default, so depending on the integration, they add their own providers that make sense, something like that.
eg sentry-spring would have a system env, file system provider etc...
sentry-android would have only its own provider, that reads from the manifest

we'll also need a new module/package like io.sentry.config that has non android related things, I dont customers needing to add extra rules to not fail their build because of unknown classes, this happened with the older version of the SDK and lot of people complained.

sentry-core/src/main/java/io/sentry/core/config/provider/resource/ResourceLoaderConfigurationProvider.java
```diff
@@ -0,0 +1,77 @@
...
+            return null;
+        }
+
+        InputStream is = rl.getInputStream(filePath);
+
+        if (is == null) {
+            return null;
+        }
+
+        try (InputStreamReader rdr = new InputStreamReader(is, charset)) {
```
marandaneto, line 55: does it make sense to use a `BufferedReader` as well?

PR #537, opened 2020-09-01
Add Log4j2 integration.

maciejwalkowiak:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Log4j2 integration based on existing Logback integration.

Feature parity with 1.x

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
```diff
@@ -0,0 +1,220 @@
...
+  private @Nullable ITransport transport;
+  private @NotNull Level minimumBreadcrumbLevel = Level.INFO;
+  private @NotNull Level minimumEventLevel = Level.ERROR;
+
+  public SentryAppender(
+      String name,
+      Filter filter,
+      @Nullable String dsn,
+      @Nullable Level minimumBreadcrumbLevel,
+      @Nullable Level minimumEventLevel,
```
bruno-garcia, line 47: Looks like all params can be final. Or none but the last one was made so

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
```diff
@@ -0,0 +1,220 @@
...
+    message.setMessage(loggingEvent.getMessage().getFormat());
+    message.setFormatted(loggingEvent.getMessage().getFormattedMessage());
+    message.setParams(toParams(loggingEvent.getMessage().getParameters()));
+    event.setMessage(message);
+    event.setLogger(loggingEvent.getLoggerName());
+    event.setLevel(formatLevel(loggingEvent.getLevel()));
+
+    final ThrowableProxy throwableInformation = loggingEvent.getThrownProxy();
+    if (throwableInformation != null) {
+      event.setThrowable(throwableInformation.getThrowable());
```
bruno-garcia, line 126: Is there something else useful on the throwableInformation?

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    return event;
+  }
+
+  private @NotNull List<String> toParams(@Nullable Object[] arguments) {
+    if (arguments != null) {
+      return Arrays.stream(arguments)
+          .filter(Objects::nonNull)
+          .map(Object::toString)
```
bruno-garcia, line 146: Is it toString what the formatter call on the args? If not, this could end up being a problem.
Since we already give both formatted and template, it's probably ok not to add the arga at all.

Sentry would group on the template and display the formatted already

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  /**
+   * Transforms a {@link Level} into an {@link SentryLevel}.
+   *
+   * @param level original level as defined in log4j.
+   * @return log level used within sentry.
+   */
+  private static @NotNull SentryLevel formatLevel(@NotNull Level level) {
+    if (level.isMoreSpecificThan(Level.ERROR)) {
```
bruno-garcia, line 174: It doesn't have Fatal?

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  public void setMinimumEventLevel(final @Nullable Level minimumEventLevel) {
+    if (minimumEventLevel != null) {
+      this.minimumEventLevel = minimumEventLevel;
+    }
+  }
+
+  @ApiStatus.Internal
+  void setTransport(@Nullable ITransport transport) {
```
bruno-garcia, line 217: Dow e really need this or can we use the client here instead?
I assume it's for testing.
We could take a IHub in the constructor instead.
An internal overload would take it. The public overload calls into it with HubAdapter.getInstance for the prod use case

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
```diff
@@ -0,0 +1,220 @@
...
+
+/** Appender for Log4j2 in charge of sending the logged events to a Sentry server. */
+@Plugin(name = "Sentry", category = "Core", elementType = "appender", printObject = true)
+public final class SentryAppender extends AbstractAppender {
+  private @Nullable String dsn;
+  private @Nullable ITransport transport;
+  private @NotNull Level minimumBreadcrumbLevel = Level.INFO;
+  private @NotNull Level minimumEventLevel = Level.ERROR;
+
+  public SentryAppender(
```
marandaneto, line 42: we could have a ctor overload that takes only the necessary, so its not needed to pass  minimumBreadcrumbLevel and minimumEventLevel, its gonna be the default values, unless this is not instantiated manually by clients, so then it doesn't matter

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
```diff
@@ -0,0 +1,220 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/** Appender for Log4j2 in charge of sending the logged events to a Sentry server. */
+@Plugin(name = "Sentry", category = "Core", elementType = "appender", printObject = true)
+public final class SentryAppender extends AbstractAppender {
+  private @Nullable String dsn;
+  private @Nullable ITransport transport;
+  private @NotNull Level minimumBreadcrumbLevel = Level.INFO;
+  private @NotNull Level minimumEventLevel = Level.ERROR;
```
marandaneto, line 40: does it make sense to add this either to options or to the Level class as static default fields and we just read the default values from them? cus I see these 2 values repeated in all the integrations, like:

```
SentryOptions.DEFAULT_BREADCRUMB_LEVEL
SentryOptions.DEFAULT_EVENT_LEVEL
```

or something like that.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
```diff
@@ -0,0 +1,220 @@
...
+            Optional.ofNullable(transport).ifPresent(options::setTransport);
+          });
+    }
+    super.start();
+  }
+
+  @Override
+  public void append(LogEvent eventObject) {
+    if (eventObject.getLevel().isMoreSpecificThan(minimumEventLevel)) {
+      Sentry.captureEvent(createEvent(eventObject));
```
marandaneto, line 99: I'd get a Hub in the ctor using the HubAdapter instead of using the Sentry static class, like you did with `sentry-spring-boot-starter`

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+  }
+
+  /**
+   * Creates {@link SentryEvent} from Log4j2 {@link LogEvent}.
+   *
+   * @param loggingEvent the log4j2 event
+   * @return the sentry event
+   */
+  @SuppressWarnings("JdkObsolete")
```
marandaneto, line 112: add a comment about the `JdkObsolete` please, I believe its about the `Date`

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
marandaneto: no

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
marandaneto: yeah I agree with @bruno-garcia 
Sentry.init already overwrites the `NoOpTransport` to `HttpTransport` on Sdk init.
if it's for the sake of testing, I'd use the hub, I guess the other integrations also have references for transport and I'd refactor as well unless there's a need that I'm not aware of.

sentry-log4j2/src/test/kotlin/io/sentry/log4j2/SentryAppenderTest.kt
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    @BeforeTest
+    fun `clear MDC`() {
+        ThreadContext.clearAll()
+    }
+
+    @Test
+    fun `converts message`() {
+        fixture = Fixture(minimumEventLevel = Level.DEBUG)
```
marandaneto, line 69: I'd use this type of params in the `getSut` method as all the other tests, looks like we have a different type of Fixture way of doing now.

sentry-log4j2/src/test/kotlin/io/sentry/log4j2/SentryAppenderTest.kt
```diff
@@ -0,0 +1,287 @@
...
+        fixture.logger.debug("testing message conversion {}, {}", 1, 2)
+
+        await.untilAsserted {
+            verify(fixture.transport).send(com.nhaarman.mockitokotlin2.check { it: SentryEvent ->
```
marandaneto, line 73: I', in favor of importing the packages instead of using in the code
eg

`import com.nhaarman.mockitokotlin2.check`

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
marandaneto: does the returned throwable is already built with its cause?
cus ThrowableProxy has a getCauseProxy() method and we are not using it, but I believe this is because the returned throwable is already built with it, just to confirm.

marandaneto (commented):
looks good, left a few comments though

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: Good catch, Log4j2 actually does have `fatal`.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: The problem is that some properties are set in `MainEventProcessor` - like sdk version. This is why I moved to mocking transport instead of Hub - some things are just not possible to test when just `Hub` is mocked.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: it's not instantiated manually by users. Users use `log4j2.xml` to configure appenders.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: It depends if we want to pollute `sentry-core` with integration specific code. At this stage core is not aware of the whole mechanism of attaching breadcrumbs to events in logging integrations.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
marandaneto: true, i've mistaken with logback

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: What would be the benefit of it considering we are not mocking it anyway?

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: Yes, throwable already has a cause.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
marandaneto: I'd be fine with that, to not DRY, of course, it depends on a case by case, but this duplication is already in 4 integrations.

It'd be something similar, defining default values to be reused elsewhere.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
marandaneto: I see, makes sense.
whats about creating a new Hub, passing options, but mocking the transport in the options, it'd be pretty much the same, but the entry point here would be the `Hub` still as every other integration.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
marandaneto: you could assert/verify if an event and breadcrumb have been captured correctly according to the level, it'd be more important if we were using the `hub` instead of the `transport` though

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: Adding extra args result in following entries in Sentry UI
image

If we do not need it, we can get rid of passing parameters.

sentry-log4j2/src/test/kotlin/io/sentry/log4j2/SentryAppenderTest.kt
maciejwalkowiak: I am not sure how this can be done as in case of this test as this is an integration test and the appender that are actually testing is not used directly in tests. Could you make a PR to this PR so that I can get a better feeling of what you have in mind?

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: Ok understood. This would mean that we need to add conversion from SentryLevel to logging-framework-level - as these levels in appenders are are from log4j2/Logback packages.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: I have nothing against but I also don't see how adding this one level of indirection makes anything better..?

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
marandaneto: yeah the converters for sure, it's just to reuse the default levels, nothing more.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
marandaneto: we could discuss what would make more sense here, but that's how it is in all the others tests, we never make a reference directly to the transport, but rather the entry point, and mock things in the options if necessary, so we could assert and verify directly from them.
if we decide that this would make more sense (assert on the transport), I'd rather change everywhere, I just would like to avoid a mix of different ways of doing things.

maybe @bruno-garcia  has opinions.

sentry-log4j2/src/test/kotlin/io/sentry/log4j2/SentryAppenderTest.kt
marandaneto: I see, that's the problem, we are testing the appender but we don't have any references to it, its more like an integration test, where the appender gets created as you were setting the logger for real, what I meant is something like:

https://github.com/getsentry/sentry-android/blob/master/sentry-android-timber/src/test/java/io/sentry/android/timber/SentryTimberIntegrationTest.kt#L18-L32

but I feel that creating a Fixture every time looks different than all the other tests, usually, we create a `sut` per test but not a fixture.
maybe our `sut` should be the `logger` then as it is the one that makes the real calls.

@bruno-garcia introduced the fixture pattern in our tests, maybe he has something to share

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
marandaneto: another advantage would be to not need `await.untilAsserted` in all the tests, as we'd be testing sync calls.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
bruno-garcia: Nothing else on `throwableInformation` that we could add to the event?

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
bruno-garcia: It's nice to have, sure. I just ask because if the Formatting code calls some `format()` method or else, instead of `toString()` the value we return in the `args` of the Sentry event won't match the actual formatted string log message

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
bruno-garcia: IMHO I also rather to keep them out since core has nothing to do with min breadcrumb level of a logger integration

DRY doesn't mean avoid any and all "duplication" by breaking cohesion

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
bruno-garcia: +1 one on avoiding to couple with the static type here.
Dunno the test runner mode here but if we parallelize tests classes, we'd have racey behavior where one calls Sentry.close and other capture etc.

sentry-log4j2/src/test/kotlin/io/sentry/log4j2/SentryAppenderTest.kt
bruno-garcia: Usually with this pattern Fixture doesn't vary (so can be instantiated with the test class) but its properties are mutable to set the state you need by the time you call `getSut()` in the test body.

Would also prefer to stick to that approach if possible but TBQH no strong feelings here, whatever you folks agree is fine by me

sentry-log4j2/build.gradle.kts
```diff
@@ -0,0 +1,103 @@
...
+    reports {
+        xml.isEnabled = true
+        html.isEnabled = false
+    }
+}
+
+tasks {
+    jacocoTestCoverageVerification {
+        violationRules {
+            rule { limit { minimum = BigDecimal.valueOf(0.6) } }
```
bruno-garcia, line 60: This min coverage value could be lifted to `Config` so we can control it for the whole project eh?

Separate PR sure, just pointing out.
I'll raise an issue.

sentry-log4j2/build.gradle.kts
```diff
@@ -0,0 +1,103 @@
...
+    buildConfigField("String", "SENTRY_LOG4J2_SDK_NAME", "\"${Config.Sentry.SENTRY_LOG4J2_SDK_NAME}\"")
+    buildConfigField("String", "VERSION_NAME", "\"${project.version}\"")
+}
+
+val generateBuildConfig by tasks
+tasks.withType<JavaCompile>().configureEach {
+    dependsOn(generateBuildConfig)
+}
+
+//TODO: move these blocks to parent gradle file, DRY
```
bruno-garcia, line 81: Every project has this comment, maybe we just get rid of it?

I'll raise an issue to consider refactoring this (or remove the comment)

sentry-samples/sentry-samples-log4j2/src/main/java/io/sentry/samples/log4j2/Main.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.UUID;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.ThreadContext;
+
+public class Main {
+  private static final Logger LOGGER = LogManager.getLogger(Main.class);
+
+  public static void main(String[] args) {
+    LOGGER.debug("Hello Sentry!");
```
bruno-garcia, line 12:
```suggestion
    // The SDK was initialized through the appender configuration because a DSN was set there.
    // Update the DSN in log4j2.xml to see these events in your Sentry dashboard.
    LOGGER.debug("Hello Sentry!");
```

sentry-samples/sentry-samples-log4j2/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,20 @@
...
+<Configuration status="WARN" packages="io.sentry.log4j2">
+  <Appenders>
+    <Console name="Console" target="SYSTEM_OUT">
+      <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
+    </Console>
+    <!-- NOTE: Replace the test DSN below with YOUR OWN DSN to see the events from this app in your Sentry project/dashboard -->
+    <Sentry name="Sentry"
+            dsn="https://f7f320d5c3a54709be7b28e0f2ca7081@sentry.io/1808954"
+            minimumBreadcrumbLevel="DEBUG"
+            minimumEventLevel="WARN"
```
bruno-garcia, line 11:
```suggestion
            <!-- Modifies the default minimum level to add breacrumbs from INFO to:  -->
            minimumBreadcrumbLevel="DEBUG"
            <!-- Modifies the default minimum level to capture an event from ERROR to:  -->
            minimumEventLevel="WARN"
```

maciejwalkowiak: Unfortunately this is not valid XML

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: Potentially they won't match in some special cases. I don't think it makes sense to decode https://github.com/apache/logging-log4j2/blob/master/log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java to make sure.

Up to you - we can remove it.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: Ok, replaced with calling `hub`.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: Some information like SDK version is added later, in `MainEventProcessor` - this is why i need to mock transport on lower level.

I started with mocking `Hub`, but this does not give me tools to test integration properly.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
maciejwalkowiak: I think everything is already there. Perhaps we can do a walk through over Zoom to see all integrations in actions and with your knowledge about Sentry maybe you'll find something missing.

sentry-log4j2/src/test/kotlin/io/sentry/log4j2/SentryAppenderTest.kt
maciejwalkowiak: I am happy to change it but really I do not see how. PRs welcome :-)

maciejwalkowiak:
@bruno-garcia @marandaneto please take another look. I would be happy if we could finish it today.

sentry-log4j2/src/main/java/io/sentry/log4j2/SentryAppender.java
bruno-garcia: Let's keep it then, I assume worst case is returning the default Object's toString which is useless. Hopefully no massive blob of data

sentry-log4j2/src/test/kotlin/io/sentry/log4j2/SentryAppenderTest.kt
maciejwalkowiak: Ok Bruno explained it to me :-)

PR #539, opened 2020-09-03
Add Spring Integration.

maciejwalkowiak:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Many of beans previously instantiated in Spring Boot integration move to Spring integration. Spring Boot integration enables more fine grained SentryOptions configuration, simple way to register custom event processors, passing git commit id as the event release version.

Feature parity with Sentry 1.x

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


sentry-spring/src/test/kotlin/io/sentry/spring/SentrySpringIntegrationTest.kt
```diff
@@ -0,0 +1,148 @@
...
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RunWith(SpringRunner::class)
+@SpringBootTest(
+    classes = [App::class],
+    webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT,
+    properties = ["sentry.dsn=http://key@localhost/proj", "sentry.send-default-pii=true"]
+)
+class SentrySpringIntegrationTest {
```
maciejwalkowiak, line 44: This test class is almost a duplicate from Sentry Spring Boot integration. I am not sure yet if/how we can reduce the duplication keeping the same confidence level.

bruno-garcia (approved):
Mini first pass

bruno-garcia (commented):
approved prematurely

sentry-samples/sentry-samples-spring/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+
+group = "io.sentry.sample.spring"
+version = "0.0.1-SNAPSHOT"
+java.sourceCompatibility = JavaVersion.VERSION_1_8
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
```
marandaneto, line 18: can we reuse the things that already live in the `Config` class?

sentry-samples/sentry-samples-spring/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+    implementation("org.springframework.boot:spring-boot-starter-security")
+    implementation("org.springframework.boot:spring-boot-starter-web")
+    implementation("org.springframework.boot:spring-boot-starter")
+    implementation("org.jetbrains.kotlin:kotlin-reflect")
+    implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
+    implementation(project(":sentry-spring"))
+    implementation(project(":sentry-logback"))
```
marandaneto, line 25: is that normal to use always the conjunction of spring + logback? I feel that this would get a little bit more complex than necessary.
these samples are to easier the development and not really for the customer, so mixing spring and logback would make it a little bit harder to test things in insolation.

sentry-samples/sentry-samples-spring/src/main/resources/application.properties
```diff
@@ -0,0 +1,6 @@
+# NOTE: Replace the test DSN below with YOUR OWN DSN to see the events from this app in your Sentry project/dashboard
+sentry.dsn=https://f7f320d5c3a54709be7b28e0f2ca7081@sentry.io/1808954
+sentry.send-default-pii=true
+
+# Uncaught exception handler should be disabled if both Sentry Spring and Logback integrations are enabled
```
marandaneto, line 5: just out of curiosity, why? do they already register an uncaught exception handler? should we not do this automatically then? otherwise, customers would need to set extra properties manually.

sentry-spring/src/main/java/io/sentry/spring/SentryRequestFilter.java
```diff
@@ -1,25 +1,27 @@
...
-public class SentryRequestFilter extends OncePerRequestFilter {
+public class SentryRequestFilter extends OncePerRequestFilter implements Ordered {
   private final @NotNull IHub hub;
   private final @NotNull SentryOptions options;
 
   public SentryRequestFilter(final @NotNull IHub hub, final @NotNull SentryOptions options) {
-    this.hub = hub;
-    this.options = options;
+    this.hub = Objects.requireNonNull(hub, "hub is required");
+    this.options = Objects.requireNonNull(options, "options are required");
```
marandaneto, line 24: the reason to not use the original JDK method `Objects.requireNonNull` is because its not available on older Android APIs, but it'd be available in this package as its non-Android.
I'll leave to you to keep this one (io.sentry.core.util.Objects) or to use the original `Objects`, this is just a JFYI comment.

sentry-spring/src/main/java/io/sentry/spring/SentryRequestFilter.java
```diff
@@ -29,11 +31,23 @@ protected void doFilterInternal(
       final @NotNull FilterChain filterChain)
       throws ServletException, IOException {
     hub.pushScope();
+    // clears breadcrumbs that may have been added during application startup through one of the logging integrations.
+    hub.clearBreadcrumbs();
```
marandaneto, line 35: would it make sense to expose the `clear` method in the hub then? which is basically cleaning everything from the `scope`?

marandaneto: could we lose useful information when cleaning all the breadcrumbs?

sentry-samples/sentry-samples-spring/src/main/resources/application.properties
maciejwalkowiak: If there is no uncaught exception handler, exception is just logged with the logging framework. If its turn on in both then we have duplicated event.

sentry-spring/src/main/java/io/sentry/spring/SentryRequestFilter.java
maciejwalkowiak: This is there to avoid having startup logs attached to every event sent. If we do not clear breadcrumbs every event looks like: 
image

sentry-spring/src/main/java/io/sentry/spring/SentryRequestFilter.java
maciejwalkowiak: My bad, I intended to use the one from `Sentry`

sentry-samples/sentry-samples-spring/build.gradle.kts
maciejwalkowiak: Logback is the default framework that comes with Spring Boot and using these two is I believe the best way to take a full advantage of Sentry capabilities.

sentry-samples/sentry-samples-spring/src/main/resources/application.properties
marandaneto: does that mean we should then disable it automatically somehow? I see this being a constant support ticket -> why dow e get duplicated events?

the other thing would be, we rely on our uncaught exception handler to end a `session` automatically, if we disable it, we'd need to find another way, or the current session would never be ended and marked as crashed properly.
I'd suggest creating a new issue here https://github.com/getsentry/sentry-java/projects/1 so we decide where/how to proceed

sentry-spring/src/main/java/io/sentry/spring/SentryRequestFilter.java
marandaneto: IMO this is fine, not sure how noisy this can get though.
every event on android also has many startup breadcrumbs, but I believe that this is what breadcrumbs stand for, if it's like less than 5, all good, it'd be a problem if its a huge number and just trash.
maybe @bruno-garcia  has opinions
I'm just afraid of losing useful breadcrumbs added by the user manually.

sentry-samples/sentry-samples-spring/build.gradle.kts
marandaneto: I guess I already said that, these samples are to easier development, not for customers.

see an example: https://github.com/getsentry/sentry-android/blob/master/sentry-android-timber/build.gradle.kts#L74

there's no need to hardcode when they already live in the Config class.
If we decide using only hardcoded strings, I'm fine with that too, but rather do it everywhere and kill the Config class for good.

those generated samples and closer to real projects will live here https://github.com/getsentry/examples/tree/master/java

sentry-samples/sentry-samples-spring/build.gradle.kts
marandaneto: nice, makes sense then

sentry-samples/sentry-samples-spring-boot/src/main/java/io/sentry/samples/spring/boot/CustomEventProcessor.java
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * Custom {@link EventProcessor} implementation lets modifying {@link SentryEvent}s before they are sent to Sentry.
+ */
+@Component
+public class CustomEventProcessor implements EventProcessor {
+  @Override
+  public SentryEvent process(SentryEvent event, @Nullable Object hint) {
+    event.setTag("Java Version", System.getProperty("java.version"));
```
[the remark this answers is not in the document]
maciejwalkowiak: yep we could, but then i guess we would add it to sentry-core? I we go for it i need another idea for reasonable example of event processor ;-)

sentry-samples/sentry-samples-spring/build.gradle.kts
maciejwalkowiak: We are already reusing things from `Config` class to set the dependency management and plugins. There are no dependencies currently set in config class that we could directly replace here.

sentry-spring/src/main/java/io/sentry/spring/SentryRequestFilter.java
maciejwalkowiak: Ah sorry I misread. I think if we just use everywhere the same guards it's less confusing. Spring has it's own built in guards (which are actually more reasonable than the ones from JDK) as they don't throw `NullPointerException` but `IllegalArgumentException`.

sentry-samples/sentry-samples-spring/build.gradle.kts
marandaneto:
```
    implementation("org.springframework.boot:spring-boot-starter-security")
    implementation("org.springframework.boot:spring-boot-starter-web")
    implementation("org.springframework.boot:spring-boot-starter")
    implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
```

I see those there

https://github.com/getsentry/sentry-android/blob/feat/sentry-java/buildSrc/src/main/java/Config.kt

sentry-spring/src/main/java/io/sentry/spring/SentryRequestFilter.java
marandaneto: cool, let's keep our own then.

sentry-samples/sentry-samples-spring-boot/src/main/java/io/sentry/samples/spring/boot/CustomEventProcessor.java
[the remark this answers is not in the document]
marandaneto: not really, it should be a processor specifically for java backend/desktop, it should not be executed on Android at least, it won't return anything.
I mean it could live in sentry-core cus there's no compiling error, but I don't think it should be executed on runtime by default on all the platforms.

sentry-samples/sentry-samples-spring/build.gradle.kts
maciejwalkowiak: In `Config` they are listed under `TestLibs` as they are used to test integrations. Should I refer them from `TestLibs` in samples or move them to `Libs`?

marandaneto: I'd move to `Libs`, guess it makes more sense, we have `Timber` there too

maciejwalkowiak: Done.

sentry-samples/sentry-samples-spring-boot/src/main/java/io/sentry/samples/spring/boot/CustomEventProcessor.java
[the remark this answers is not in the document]
maciejwalkowiak: Sentry does not recognize Java runtime. I keep it for now in a sample, we can consider moving it to `sentry-core` once it's agreed what runtime values are recognized.

sentry-samples/sentry-samples-spring/src/main/resources/application.properties
maciejwalkowiak: Flag is gone. This is solved now by event deduplication.

sentry-samples/sentry-samples-spring-boot/src/main/java/io/sentry/samples/spring/boot/CustomEventProcessor.java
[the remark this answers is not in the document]
marandaneto: @maciejwalkowiak what do you mean by "Sentry does not recognize Java runtime"? if it's about the icon, this is just an UI limitation.
but agreed, we can make this in other PR, it doesn't make sense to block this one.

sentry-core/src/main/java/io/sentry/core/Breadcrumb.java
```diff
@@ -43,6 +43,22 @@
...
   }
 
+  /**
+   * Creates HTTP breadcrumb.
+   *
+   * @param url - the request URL
+   * @param method - the request method
+   * @return the breadcrumb
+   */
+  public static @NotNull Breadcrumb http(final @NotNull String url, final @NotNull String method) {
```
marandaneto, line 53: we could add status code as well?

https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/#http

wondering if the category should be `xhr`

maciejwalkowiak: I assumed `xhr` term makes sense only in the context or browser interacting with a server.

Status is optional - this breadcrumb is added when request processing starts. On the go I think we could add more methods here.

sentry-samples/sentry-samples-spring-boot/src/main/java/io/sentry/samples/spring/boot/CustomEventProcessor.java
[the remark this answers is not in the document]
maciejwalkowiak: Ah my bad. I thought unknown values are not passed but I just had NPE thrown in the processor. All good now :-) Anyway, I would prefer to close this one and move it to the next PR.

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
```diff
@@ -16,14 +16,14 @@ repositories {
...
 
 dependencies {
-    implementation("org.springframework.boot:spring-boot-starter-security")
-    implementation("org.springframework.boot:spring-boot-starter-web")
-    implementation("org.springframework.boot:spring-boot-starter")
+    implementation(Config.Libs.springBootStarterSecurity)
+    implementation(Config.Libs.springBootStarterWeb)
+    implementation(Config.Libs.springBootStarter)
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
```
marandaneto, line 23:
```suggestion
    implementation(kotlin(Config.kotlinStdLib, KotlinCompilerVersion.VERSION))
```

sentry-spring/build.gradle.kts
```diff
@@ -0,0 +1,118 @@
...
+    reports {
+        xml.isEnabled = true
+        html.isEnabled = false
+    }
+}
+
+tasks {
+    jacocoTestCoverageVerification {
+        violationRules {
+            rule { limit { minimum = BigDecimal.valueOf(0.6) } }
```
marandaneto, line 75: can we reuse the value from the Config class?

sentry-spring/src/main/java/io/sentry/spring/SentryInitBeanPostProcessor.java
```diff
@@ -0,0 +1,21 @@
+package io.sentry.spring;
+
+import com.jakewharton.nopen.annotation.Open;
+import io.sentry.core.Sentry;
+import io.sentry.core.SentryOptions;
+import org.jetbrains.annotations.NotNull;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.config.BeanPostProcessor;
+
+/** Initializes Sentry after all beans are registered. */
```
marandaneto, line 10: out of curiosity, if there's a bug on customer's beans and Sentry is the last to be inited, would that mean that we won't be able to capture the crash?

marandaneto (commented):
left a few more comments but I'd say they are not blockers, let's iterate over them if necessary, and happy to LGTM.
looks good, well done!

sentry-spring/src/main/java/io/sentry/spring/SentryInitBeanPostProcessor.java
maciejwalkowiak: This is why logging framework integration should be used as well. Logging framework is initialized first so it will catch any bug that happens during startup.

sentry-spring/src/main/java/io/sentry/spring/SentryRequestFilter.java
maciejwalkowiak: I still believe these breadcrumbs create noise, but since both @marandaneto and @bruno-garcia are more leaning towards keeping them, they are now not removed.

sentry-spring/src/main/java/io/sentry/spring/SentryInitBeanPostProcessor.java
marandaneto: ok sounds good, still curious how this would work, if a logging framework forwards the call to sentry, sentry won't be init./enabled yet, right, unless the logging framework has a sort of caching, not sure.
we can discuss this offline though, I'd like to understand a little bit more :) thanks.

marandaneto (approved):
looks very good, well done @maciejwalkowiak 
LGTM.

sentry-spring/src/test/kotlin/io/sentry/spring/SentrySpringIntegrationTest.kt
marandaneto: should we maybe add a comment to this class and the "duplicated" one about each other? so when touching either this one or the other one, we know that they need to keep kinda synced

maciejwalkowiak:
Thanks @marandaneto!

sentry-spring/src/main/java/io/sentry/spring/SentryInitBeanPostProcessor.java
maciejwalkowiak: Logging frameworks calls init first, sends startup events to sentry, then Spring initializes Sentry and configuration provided there is used from now on. The drawback of this solution is that in order to get startup events sent to sentry DSN has to be provided twice - once to Logback configuration and then to Spring configuration.

marandaneto: I see, but it's a good trade-off I'd say, we just document it properly, and it's fine IMO.
in the future we could look into it to provide a single DSN, might be possible with a little more work, just guessing here though.

maciejwalkowiak: Yes, it will be possible if we follow up on https://github.com/getsentry/sentry-android/pull/536. To get it working SDK has to have a way to obtain configuration from external source - independent from framework-specific configuration. 
I think we will release what we have and then find out next steps from users feedback.

marandaneto: that sounds great, thanks for explaining!

maciejwalkowiak:
@bruno-garcia take another look please.

sentry-samples/sentry-samples-spring-boot/src/main/java/io/sentry/samples/spring/boot/CustomEventProcessor.java
```diff
@@ -0,0 +1,35 @@
...
+    this.javaVendor = javaVendor;
+  }
+
+  public CustomEventProcessor() {
+    this(System.getProperty("java.version"), System.getProperty("java.vendor"));
+  }
+
+  @Override
+  public SentryEvent process(SentryEvent event, @Nullable Object hint) {
+    final SentryRuntime runtime = new SentryRuntime();
```
bruno-garcia, line 29: Would be nice to do this to all events (put it in the main event processor)
Likely not going to work for Android though

sentry-spring-boot-starter/build.gradle.kts
```diff
@@ -50,9 +51,9 @@ dependencies {
     testImplementation(kotlin(Config.kotlinStdLib))
     testImplementation(Config.TestLibs.kotlinTestJunit)
     testImplementation(Config.TestLibs.mockitoKotlin)
-    testImplementation(Config.TestLibs.springBootStarterTest)
-    testImplementation(Config.TestLibs.springBootStarterWeb)
-    testImplementation(Config.TestLibs.springBootStarterSecurity)
+    testImplementation(Config.Libs.springBootStarterTest)
```
bruno-garcia, line 54: I think this one was supposed to stay as `TestLibs`

sentry-spring/build.gradle.kts
```diff
@@ -0,0 +1,118 @@
...
+    errorprone(Config.CompileOnly.errorprone)
+    errorproneJavac(Config.CompileOnly.errorProneJavac8)
+    compileOnly(Config.CompileOnly.jetbrainsAnnotations)
+
+    // tests
+    testImplementation(kotlin(Config.kotlinStdLib))
+    testImplementation(Config.TestLibs.kotlinTestJunit)
+    testImplementation(Config.TestLibs.mockitoKotlin)
+    testImplementation(Config.Libs.springBootStarterTest)
+    testImplementation(Config.Libs.springBootStarterWeb)
```
bruno-garcia, line 50: here too, test libs only?

bruno-garcia (approved):

sentry-spring/build.gradle.kts
marandaneto: if it's used in the samples, it's fine to be in the Libs IMO.

sentry-samples/sentry-samples-spring-boot/src/main/java/io/sentry/samples/spring/boot/CustomEventProcessor.java
marandaneto: yeah it won't, we could have the processor on sentry-core but either it gets added to the processor list only if its non-Android or we do a conditional flag.

PR #540, opened 2020-09-04
Fix resolving isAttachStacktrace flag.

maciejwalkowiak:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

`isAttachStacktrace` was passed down in `SentryOptions` constructor before this field was initialized with setters.

This PR changes it in a way that instead of raw value, instances of `SentryOptions` is passed down which also enables potential reconfiguring of this flag during runtime.

There was a bug we found out while working on Spring integration.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
+++ b/sentry-core/src/main/java/io/sentry/core/MainEventProcessor.java
@@ -23,7 +23,7 @@ public final class MainEventProcessor implements EventProcessor {
 
     sentryExceptionFactory = new SentryExceptionFactory(sentryStackTraceFactory);
     sentryThreadFactory =
-        new SentryThreadFactory(sentryStackTraceFactory, this.options.isAttachStacktrace());
+        new SentryThreadFactory(sentryStackTraceFactory, this.options);
   }
 
   MainEventProcessor(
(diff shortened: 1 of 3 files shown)


marandaneto:
Fix: https://github.com/getsentry/sentry-java/issues/897

marandaneto (approved):
LGTM

maciejwalkowiak:
Thanks for quick review @marandaneto

bruno-garcia:
Would be nice to get tests to avoid regression

maciejwalkowiak:
@bruno-garcia I was thinking about it - this would have to be in an integration test - I am not sure if we have one already (I didn't find it).

I could create a test like that and mock `ITransport` (like in Spring integration).

bruno-garcia:
Would be very nice to get some integrations test. Doesn't need to be too complex but at least test some bootstrapping use cases.

bruno-garcia:
Maybe even mocking the http client only? Só we test the transport too which has lots of code. Would prob require a bit of refactor to inject some fake http client there

maciejwalkowiak:
Lets add it to the list

marandaneto:
> also enables potential reconfiguring of this flag during runtime.

this actually goes against the "immutable" options, everything that should be mutable on runtime should live in the Scope, that's what I always have been said.

maciejwalkowiak:
Ah that's fair point - but then if we want to have immutable options, lets make them really immutable and expose builder to users instead of object with setters. What do you think about it?

bruno-garcia:
It's a mutable object, we need it to be. Changing it's state after the SDK initialized is not a documented/supported scenario. That sad it's just something technically feasible "if you know what you are doing", and we **could** (ab)use that if needed to work around stuff. I did use that to swap the `ILogger` at runtime even after the SDK was initialized. Again, hack to work around some issue.

PR #541, opened 2020-09-04
Extract minimum required coverage by Jacoco plugin.

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Move minimum required coverage to `Config` class and use same value (`0.6`) in all modules.

Reduce duplication and possibility of having different quality gates in different modules.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/buildSrc/src/main/java/Config.kt
+++ b/buildSrc/src/main/java/Config.kt
@@ -1,3 +1,5 @@
+import java.math.BigDecimal
+
 object Config {
     val kotlinVersion = "1.3.72"
     val kotlinStdLib = "stdlib-jdk8"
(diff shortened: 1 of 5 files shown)


marandaneto (approved):
LGTM

buildSrc/src/main/java/Config.kt
```diff
@@ -72,6 +74,10 @@ object Config {
     }
 
     object QualityPlugins {
+        object Jacoco {
+            val version = "0.8.5"
+            val minimumCoverage = BigDecimal.valueOf(0.6)
+        }
         val jacocoVersion = "0.8.5"
```
bruno-garcia, line 81: This should be deleted?

bruno-garcia:
Thanks for this

buildSrc/src/main/java/Config.kt
maciejwalkowiak: Good catch. I'll remove it in Spring PR.

PR #542, opened 2020-09-07
ref: move init flag

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
ref: move init flag

we don't want to delete envelopes that have a init session in it, so we move to the next envelope.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## :crystal_ball: Next steps
discuss the TODO


sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
```diff
@@ -91,4 +118,192 @@ protected void rotateCacheIfNeeded(final @NotNull File[] files) {
...
+
+        final Session session = readSession(envelopeItem);
+
+        if (!isValidSession(session)) {
+          continue;
+        }
+
+        // if its already true, nothing to do
+        final Boolean init = session.getInit();
+        if (init != null && init) {
```
bruno-garcia, line 168: L: Maybe we log here? Since we only get valid sessions (status `OK`) at this point, if `init=true` here we have a bug, right? Meaning we didn't close the first one or we're adding `init` to more than one session update

sentry-core/src/main/java/io/sentry/core/cache/EnvelopeCache.java
```diff
@@ -132,7 +132,10 @@ public void store(final @NotNull SentryEnvelope envelope, final @Nullable Object
       }
       updateCurrentSession(currentSessionFile, envelope);
     }
-    // TODO: problem we need to update the current session file
+
+    // TODO: probably we need to update the current session file for session updates to because of
```
bruno-garcia, line 136: Agreed. This is a follow up PR?

bruno-garcia (approved):

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
marandaneto: makes sense

sentry-core/src/main/java/io/sentry/core/cache/CacheStrategy.java
```diff
@@ -91,4 +118,192 @@ protected void rotateCacheIfNeeded(final @NotNull File[] files) {
...
+                  notDeletedFile.getAbsolutePath());
+        }
+
+        saveNewEnvelope(newEnvelope, notDeletedFile, notDeletedFileTimestamp);
+        break;
+      } else {
+        // TODO: if theres no more files with that session id, should we still delete it?
+        // that means the next session update will be init=false but the one
+        // with init=true will be deleted.
+        // either we don't delete it or we generate a new envelope only with that session.
```
marandaneto, line 212: @bruno-garcia I added this comment just to discuss it here, but this is an acceptable thing as its a corner case, right?

bruno-garcia: yeah lets creeate something on asana to track but it's fine to just drop it.

I was going to comment to look into the open session but that's racey and anyway too much complexity to cover this. Happy with just removing it in this case

marandaneto: actually this situation is described in the DACI, right, do we really need to create a ticket out of it? I mean, its something that we agreed on

PR #545, opened 2020-09-08
Bump: test libs and plugins

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
Bump: test libs and plugins

keeping test libs and plugins up to date

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## :crystal_ball: Next steps
edit: Upgrade Kotlin to 1.4.0 and robolectric to 4.4, they have breaking changes so it'd take more time to fix, as they are only for testing, I won't do it now. -> done

replace `@TestOnly` with more specific `@VisibleForTesting`

### DIFF
--- a/sentry-android-core/build.gradle.kts
+++ b/sentry-android-core/build.gradle.kts
@@ -64,6 +64,11 @@ android {
         // We run a full lint analysis as build part in CI, so skip vital checks for assemble tasks.
         isCheckReleaseBuilds = false
     }
+
+    // needed because of Kotlin 1.4.x
+    configurations.all {
+        resolutionStrategy.force(Config.CompileOnly.jetbrainsAnnotations)
+    }
 }
 
 dependencies {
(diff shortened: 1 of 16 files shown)


marandaneto:
fails due to https://github.community/t/android-projects-are-not-working/131143

sentry-android-core/build.gradle.kts
```diff
@@ -64,6 +64,11 @@ android {
         // We run a full lint analysis as build part in CI, so skip vital checks for assemble tasks.
         isCheckReleaseBuilds = false
     }
+
+    // needed because of Kotlin 1.4.x
```
bruno-garcia, line 68: What is this for? Besides I assume getting to compile

bruno-garcia (approved):

sentry-android-core/build.gradle.kts
marandaneto: we use `org.jetbrains:annotations:20.1.0` and they pinned `13.0.0` but the API has only addition to the lib, so it's not an issue to overwrite it.

PR #548, opened 2020-09-08
Fix/envelope transport tests

maciejwalkowiak:


sentry-test-support/src/main/kotlin/io/sentry/test/assertions.kt
```diff
@@ -0,0 +1,21 @@
...
+import io.sentry.core.GsonSerializer
+import io.sentry.core.NoOpLogger
+import io.sentry.core.SentryEnvelope
+import io.sentry.core.SentryEvent
+import io.sentry.core.SentryOptions
+
+/**
+ * Verifies is [SentryEnvelope] contains first event matching a predicate.
+ */
+inline fun checkEvent(noinline predicate: (SentryEvent) -> Unit): SentryEnvelope {
```
marandaneto, line 13: just out of curiosity, could this not live in the sentry-core/test package?

maciejwalkowiak: Can it be then included in other modules? It wouldn't be straightforward with Maven. Perhaps there is a reasonable way to do it with Gradle?

marandaneto: mm good catch, I thought `testImplementation` would be enough, but I think I've seen ways of doing that, just don't remember by heart now.

marandaneto:
you can ignore `LGTM` btw, I just deactivated it, their Android setup is broken and I've raised an issue for them

maciejwalkowiak:
@marandaneto could you please approve or tell what needs to be changed? Creating separate module has been suggested by @bruno-garcia.

Having `feat/sentry-java` broken is currently a blocker for me.

marandaneto (approved):
LGTM

PR #549, opened 2020-09-09
Add option to set user information via `SentryUserProvider`

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

Add option to set user information via `SentryUserProvider` hook & refactor setting user information on events.

Users will have a simple way to define a bean that returns user information. 

Same mechanism is used by Sentry Spring integration to set user information from HttpRequest if `send-default-pii` is set to `true`.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


sentry-spring/src/main/java/io/sentry/spring/SentryUserProviderEventProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+      Optional.ofNullable(user.getIpAddress()).ifPresent(existingUser::setIpAddress);
+      Optional.ofNullable(user.getUsername()).ifPresent(existingUser::setUsername);
+      if (user.getOthers() != null && !user.getOthers().isEmpty()) {
+        if (existingUser.getOthers() == null) {
+          existingUser.setOthers(new ConcurrentHashMap<>());
+        }
+        for (Map.Entry<String, String> entry : user.getOthers().entrySet()) {
+          existingUser.getOthers().put(entry.getKey(), entry.getValue());
+        }
+      }
```
marandaneto, line 38: does it make sense to assign the `getOthers` to a variable instead of calling the getters multiple times? I remember @bruno-garcia advising me to do this multiple times.

bruno-garcia: You got to convince me that this is just the Java way, that these getters are _expected to be cheap_. I'm happy to back off.

My reasoning comes from C# that properties (that read like field) even though are methods are not expected to have side-effect or to be costly to be called.
Methods on the other hand the opposite. Because of that when I see a call to a method I immediately think of assigning the result to use the same reference.

It's also relevant for instances accessed by multiple thread, so you get a consistent view of some state by inspecting the same instance (say it's reading-only operations) while another thread could swap out the reference the getter returns in between the reads.

Since I'm the odd ball here, you can ignore me and go ahead with the `getX()` as it's the "Java way".

marandaneto: you forgot to update me on that though haha, thanks for explaining!

marandaneto (approved):
LGTM, looks way easier to set the user via provider, good ref.

bruno-garcia (approved):

PR #551, opened 2020-09-10
feat: support android 11

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/buildSrc/src/main/java/Config.kt
+++ b/buildSrc/src/main/java/Config.kt
@@ -1,5 +1,5 @@
 object Config {
-    val kotlinVersion = "1.4.0"
+    val kotlinVersion = "1.4.10"
     val kotlinStdLib = "stdlib-jdk8"
 
     object BuildPlugins {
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/java/Config.kt
```diff
@@ -10,9 +10,8 @@ object Config {
     }
 
     object Android {
-        private val sdkVersion = 29
+        private val sdkVersion = 30
 
-        val buildToolsVersion = "29.0.3"
```
marandaneto, line 15: > If you're using Android plugin for Gradle 3.0.0 or higher, your project automatically uses a default version of the build tools that the plugin specifies

https://developer.android.com/studio/releases/build-tools

bruno-garcia (approved):
And CI just worked? Nice

